PR #2, opened 2018-08-23
Add Travis configuration

lgonczar:
### Change description ###

* Add Travis build configuration
* Add Travis badge to the README
* Exclude (non-existing) integration tests from jacoco analysis (otherwise the build would fail)

```
[ ] Yes
[x] No
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,26 @@
+language: java
+
+branches:
+  only:
+    - master
+
+before_cache:
+- rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
+- rm -fr $HOME/.gradle/caches/*/plugin-resolution/
+
+cache:
+  directories:
+  - $HOME/.gradle/caches/
+  - $HOME/.gradle/wrapper/
+
+before_install:
+  - curl https://raw.githubusercontent.com/hmcts/reform-api-docs/master/bin/publish-swagger-docs.sh > publish-swagger-docs.sh
+
+script:
+  - ./gradlew test
+  - ./gradlew integration
+
+after_success:
+  - ./gradlew jacocoTestReport
+  - bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+  directories:
+  - $HOME/.gradle/caches/
+  - $HOME/.gradle/wrapper/
+
+before_install:
+  - curl https://raw.githubusercontent.com/hmcts/reform-api-docs/master/bin/publish-swagger-docs.sh > publish-swagger-docs.sh
+
+script:
+  - ./gradlew test
+  - ./gradlew integration
```
nitinprabhu: can be added later ?

.travis.yml
```diff
@@ -0,0 +1,26 @@
+language: java
+
+branches:
+  only:
+    - master
```
timja: why only run on master?

doncem: running only on master based branches only. including master. so one can push as many commits as he likes and travis will be executed only when pr is raised

README.md
```diff
@@ -1 +1,3 @@
-# bulk-scan-orchestrator
\ No newline at end of file
+# bulk-scan-orchestrator
```
doncem: Perhaps it is a good occasion to at least name it `Bulk Scan Orchestrator` 😉

build.gradle
```diff
@@ -80,7 +80,8 @@ pmd {
 }
 
 jacocoTestReport {
-  executionData(test, integration)
+  // TODO: add 'integration' when integration tests are added
+  executionData(test)
```
doncem: the integration task is executed in travis build so it is no harm in leaving execution data as it was

.travis.yml
lgonczar: The task exists already, so I think it's fine that it's included. We don't risk forgetting to add it later.

README.md
lgonczar: I will do the overall readme part in one of the coming PRs.

build.gradle
lgonczar: That task is included, but even if it wasn't, without that bit may cause `jacocoTestReport` issues locally until integration tests are created.

PR #4, opened 2018-08-23
Add sonarqube task to project configuration

lgonczar:
### Change description ###

Add sonarqube task to project configuration (required by the pipeline)

```
[ ] Yes
[x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,6 +7,7 @@ plugins {
   id 'org.springframework.boot' version '2.0.3.RELEASE'
   id 'org.owasp.dependencycheck' version '3.2.1'
   id 'com.github.ben-manes.versions' version '0.20.0'
+  id 'org.sonarqube' version '2.6.2'
 }
 
 group = 'uk.gov.hmcts.reform'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -89,6 +90,18 @@ jacocoTestReport {
...
 
+project.tasks['sonarqube'].dependsOn test, integration
+
+sonarqube {
+  properties {
+    property "sonar.projectName", "Reform :: Bulk Scan Orchestrator"
+    property "sonar.jacoco.reportPath", "${project.buildDir}/jacoco/test.exec"
+
+    // TODO: uncomment the following line when integration tests are created
+    // property "sonar.jacoco.itReportPath", "${project.buildDir}/jacoco/integration.exec"
```
rkondratowicz: it may be safer to add an empty integration package (or with dummy test) ;)

lgonczar: I've started considering that, too. Will do.

doncem:
Perhaps worth also adding sonarcloud step similarly like we have in processor repo? I think token is same across organisation so https://github.com/hmcts/bulk-scan-processor/blob/master/.travis.yml#L3-L7 and other lines in travis config could be just copied.

lgonczar:
I'll add it for consistency with bulk-scan-processor

lgonczar:
Not adding it in the end - we're waiting to be added to hmcts organisation in sonarcloud

build.gradle
```diff
@@ -89,6 +90,16 @@ jacocoTestReport {
   }
 }
 
+project.tasks['sonarqube'].dependsOn test, integration
+
+sonarqube {
+  properties {
+    property "sonar.projectName", "Reform :: Bulk Scan Orchestrator"
```
rkondratowicz: i'm not sure, but `settings.gradle` file may also be needed (check other repos)

PR #12, opened 2018-09-05
Add bulk scan processor client

rkondratowicz:
https://tools.hmcts.net/jira/browse/RPE-652

### Change description ###
Add a client to communicate with the bulk scan processor (to retrieve an envelopes).
It will be used in the job reading from the queue.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/bulkscanprocessorclient/client/BulkScanProcessorClient.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/bulkscanprocessorclient/client/BulkScanProcessorClient.java
@@ -1,4 +1,4 @@
-package uk.gov.hmcts.reform.bulkscan.orchestrator.services;
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.bulkscanprocessorclient.client;
 
 import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpHeaders;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/bulkscanprocessorclient/model/Envelope.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.bulkscanprocessorclient.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class Envelope {
+
+    public final String id;
+    public final String zipFileName;
```
rkondratowicz: we'll add more field once we know what's needed by this service.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/bulkscanprocessorclient/client/BulkScanProcessorClient.java
```diff
@@ -0,0 +1,44 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.bulkscanprocessorclient.client;
```
doncem: That namespace... 😶

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/bulkscanprocessorclient/client/BulkScanProcessorClient.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public BulkScanProcessorClient(
+        String baseUrl,
+        Supplier<String> serviceAuthTokenSupplier
+    ) {
+        this.baseUrl = baseUrl;
+        this.serviceAuthTokenSupplier = serviceAuthTokenSupplier;
+    }
+
+    public Envelope getEnvelopeById(String id) {
```
doncem: Can we start introducing hystrix? any client call like this is a perfect example of the benefit of hystrix circuit breaker

PR #14, opened 2018-09-05
Fix checkstyle dtd schema url

rkondratowicz:
I was getting errors running checkstyle locally.

Found fix here:
https://discuss.gradle.org/t/checkstyle-plugin-suncertpathbuilderexception/27394

New URL taken from latest google ruleset:
https://github.com/checkstyle/checkstyle/blob/checkstyle-8.10.1/src/main/resources/google_checks.xml


timja:
your builds failing for a checkstyle violation

PR #20, opened 2018-09-11
[PREVIEW]Enable docker in the pipeline

nitinprabhu:

https://tools.hmcts.net/jira/browse/RPE-699

### Change description ###

- Enable docker support in the pipeline

- For preview image will be published at hmcts.azurecr.io/hmcts/bulk-scan-orchestrator:pr-<no>

- On master image will be published at hmcts.azurecr.io/hmcts/bulk-scan-orchestrator:latest

```
[ ] Yes
[x ] No
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,11 @@ FROM hmcts/cnp-java-base:openjdk-jre-8-alpine-1.3
 
 ENV APP bulk-scan-orchestrator.jar
 ENV APPLICATION_TOTAL_MEMORY 1024M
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM hmcts/cnp-java-base:openjdk-jre-8-alpine-1.3
```
timja: 1.4 is out, where'd u get 1.3 from?

nitinprabhu: got it from rhubarb recipe service :) will update

PR #21, opened 2018-09-12
[PREVIEW] Read messages from the queue

rkondratowicz:


src/main/resources/application.yaml
```diff
@@ -12,6 +12,10 @@ spring:
 
 bulk-scan-processor-url: ${BULK_SCAN_PROCESSOR_URL:}
 
+queue:
+  conn-string: ${QUEUE_CONNECTION_STRING:}
```
timja: how about: connection-string
you're doing the longer version in the env variable anyway, may as well be consistent

nitinprabhu: Is terraform config part of separate PR ?

rkondratowicz: done

rkondratowicz: it's actually already on master, thanks to @timja 
https://github.com/hmcts/bulk-scan-orchestrator/blob/master/infrastructure/main.tf#L39

src/main/resources/application.yaml
```diff
@@ -12,6 +12,10 @@ spring:
 
 bulk-scan-processor-url: ${BULK_SCAN_PROCESSOR_URL:}
 
+queue:
+  connection-string: ${QUEUE_CONNECTION_STRING:}
+  read-interval: ${QUEUE_READ_INTERVAL:30000} # in ms
```
luigibk: Won't we need to rename all the connection string variables if we need to add other queues?

rkondratowicz: **IF** we do, then yes :)

PR #23, opened 2018-09-12
Handle test messages

rkondratowicz:
To be used in functional / smoke tests.

Use message label to determine whether an envelope should be read from the core service.


nitinprabhu:
Review comment not addressed but PR merged ?

rkondratowicz:
When will functional tests be run inside docker?

PR #38, opened 2018-09-20
Enable more endpoints

timja:
### Change description ###
In spring boot 2 only health and info are enabled by default,
There's lots more useful endpoints that make debugging a lot easier.

Things like changing log levels at runtime, checking configuration properties or environment variables
See a list here:
https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html

```
[ ] Yes
[x] No
```


rkondratowicz:
Environment variables have secrets, so maybe not `env`

timja:
passwords are redacted in env, its very useful to see what config is actually applied,

PR #42, opened 2018-09-24
Handle null queue message

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/EnvelopeParser.java
```diff
@@ -19,7 +19,7 @@
     public static Envelope parse(byte[] bytes) {
         try {
             return objectMapper.readValue(bytes, Envelope.class);
-        } catch (IOException exc) {
+        } catch (IOException | NullPointerException exc) {
```
timja: not sure we should be doing this, catching null pointer is an anti pattern.
either add a null check first or fix the code that is causing null to be sent?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/EnvelopeParser.java
```diff
@@ -19,7 +19,7 @@
     public static Envelope parse(byte[] bytes) {
         try {
             return objectMapper.readValue(bytes, Envelope.class);
```
nitinprabhu: Are we handling scenario where bytes are null ? We can put throw custom exception message rather than handling null ?

rkondratowicz: different from `InvalidMessageException`?

PR #45, opened 2018-09-26
RPE-716: Read a case from CCD

KevinJCross:
https://tools.hmcts.net/jira/secure/RapidBoard.jspa?rapidView=455

### Change description ###
Added the basic flow from a message to getting a case from CCD.
- NOTE: There is another ticket to create a Integration test to fully test this from a whole perspective.
Im continuing writing some Integration tests to test the exceptional cases with [RPE-735](https://tools.hmcts.net/jira/secure/RapidBoard.jspa?rapidView=455&view=detail&selectedIssue=RPE-735)


src/test/resources/example1.json
```diff
@@ -0,0 +1,9 @@
+{
+  "id": "0192",
+  "case_ref": "1537879748168579",
+  "jurisdiction": "SSCS",
+  "classification": "NEW_APPLICATION",
```
rkondratowicz: isn't `zipFileName` required?

bin/send_message.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+set -eu
+file_name=${1:-"example1.json"}
+environment=${2:-"sandbox"}
+file="../src/test/resources/${file_name}"
+echo "using file: ${file_name} environment:${environment}"
+shared_access_signature="SharedAccessSignature sr=sb%3A%2F%2Fbulk-scan-servicebus-sandbox.servicebus.windows.net%2F&sig=Cz5XsYG8tGYUkY0nwrtO%2Fkxc2YOPTsoDyrJ%2BwctPWEU%3D&se=1538398077&skn=SendSharedAccessKey"
```
rkondratowicz: should this be here?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/Application.java
```diff
@@ -2,10 +2,12 @@
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.scheduling.annotation.EnableScheduling;
 
 @SpringBootApplication
 @EnableScheduling
+@EnableConfigurationProperties
```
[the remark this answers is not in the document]
nitinprabhu: Not needed I believe as you have it in https://github.com/hmcts/bulk-scan-orchestrator/pull/45/files#diff-f363559b1b219ff4a0045d10ed88b26bR13

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CcdAuthService.java
```diff
@@ -0,0 +1,32 @@
...
+    public CcdAuthService(AuthTokenGenerator s2sTokenGenerator,
+                          IdamClient idamClient,
+                          JurisdictionToUserMapping users) {
+        this.s2sTokenGenerator = s2sTokenGenerator;
+        this.idamClient = idamClient;
+        this.users = users;
+    }
+
+    CcdAuthInfo authenticateForJurisdiction(String jurisdiction) {
+        String sscsToken = s2sTokenGenerator.generate();
```
nitinprabhu: variable name needs changing sscsToken ?

src/test/java/GetSASToken.java
```diff
@@ -0,0 +1,39 @@
+import javax.crypto.Mac;
+import javax.crypto.spec.SecretKeySpec;
+import java.net.URLEncoder;
+import java.util.Base64;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public class GetSASToken {
```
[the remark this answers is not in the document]
nitinprabhu: I think we can remove this class. It is already in readme ?

src/test/java/GetSasToken.java
```diff
@@ -0,0 +1,39 @@
+import java.net.URLEncoder;
+import java.util.Base64;
+import javax.crypto.Mac;
+import javax.crypto.spec.SecretKeySpec;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public class GetSasToken {
```
nitinprabhu: Don't  need this class duplicate ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CcdAuthService.java
```diff
@@ -0,0 +1,32 @@
...
+        Credential user = users.getUser(jurisdiction);
+        String userToken = idamClient.authenticateUser(user.getUsername(), user.getPassword());
+        UserDetails userDetails = idamClient.getUserDetails(userToken);
```
nitinprabhu: Do we get a null token if we are not able to authenticate ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CcdCaseRetriever.java
```diff
@@ -0,0 +1,25 @@
...
+@Service
+public class CcdCaseRetriever {
+    public static final String CASE_TYPE_ID = "Bulk_Scanned";
+    private final CoreCaseDataApi coreCaseDataApi;
+
+    CcdCaseRetriever(CoreCaseDataApi coreCaseDataApi) {
+        this.coreCaseDataApi = coreCaseDataApi;
+    }
+
+    CaseDetails retrieve(CcdAuthInfo authInfo, String caseRef) {
```
nitinprabhu: retrieve --> retrieveCCDCase or  retrieveCaseData?

bin/send_message.sh
timja: no it needs to be calculated / passed in

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/Application.java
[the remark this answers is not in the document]
timja: try remove then 👍

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CcdCaseRetriever.java
```diff
@@ -0,0 +1,25 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services;
+
+import org.springframework.stereotype.Service;
+import uk.gov.hmcts.reform.ccd.client.CoreCaseDataApi;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+
+@Service
+public class CcdCaseRetriever {
```
timja: Doesn't seem to give any value currently?
Because it has retriever in it its only ever going to have one method in it
and the method is quite simple anyway?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CompletableHelper.java
```diff
@@ -0,0 +1,21 @@
...
+        /*
+         * NOTE: this is done here instead of offloading to the forkJoin pool "CompletableFuture.runAsync()"
+         * because we probably should think about a threading model before doing this.
+         * Maybe consider using Netflix's RxJava too (much simpler than CompletableFuture).
+         */
+        CompletableFuture<Void> completableFuture = new CompletableFuture<>();
+        try {
+            runnable.run();
+            completableFuture.complete(null);
+        } catch (Throwable t) {
```
timja: it the design pattern here to catch throwable?
normally you should only catch exception, and leave errors alone

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/TestData.java
```diff
@@ -0,0 +1,46 @@
...
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.idam.Credential;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+import uk.gov.hmcts.reform.idam.client.models.UserDetails;
+
+import static java.util.Arrays.asList;
+import static java.util.Collections.emptyList;
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.CcdCaseRetriever.CASE_TYPE_ID;
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.model.Classification.NEW_APPLICATION;
+
+public interface TestData {
```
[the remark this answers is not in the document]
rkondratowicz: Please try to keep the changes in PRs that add new functionality to minimum.
It's now really hard to see the relevant changes.
Classes from previous PR related to loading user credentials got renamed / moved to other namespace, so did some exceptions (if I understand correctly)

src/main/resources/application.yaml
```diff
@@ -19,6 +19,10 @@ queue:
   connection-string: ${QUEUE_CONNECTION_STRING:}
   read-interval: ${QUEUE_READ_INTERVAL:30000} # in ms
 
+core_case_data:
```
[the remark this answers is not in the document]
timja: no this is correct, it should map to the docker url
and then
in infrastructure/main.tf the var needs to be added for the deployed envs,
`CORE_CASE_DATA_API_URL=...`

src/test/java/GetSASToken.java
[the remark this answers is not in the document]
timja: I would move this to shared infra.. in the bin directory..

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/TestData.java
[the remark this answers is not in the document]
timja: not really sure on this change two reasons,
1. the original name was fine imo
2. what benefit is the interface giving over a class?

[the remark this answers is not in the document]
KevinJCross: sure ... but its the changes that are needed to complete this PR.
Im not sure what changes that are smaller would actually be beneficial.
I actually made a new exception then deleted a duplicated type of exception that was not used any more ... but they were doing the same thing.

src/test/resources/example1.json
KevinJCross: not my change ... it was actually merged into master today.

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/config/JurisdictionToUserMappingTest.java
```diff
@@ -0,0 +1,51 @@
...
+        Credential creds = mapping.getUser("SSCS");
+        assertThat(creds.getPassword()).isEqualTo("Password12");
+        assertThat(creds.getUsername()).isEqualTo("bulkscanorchestrator+systemupdate@gmail.com");
+    }
+
+    @Test
+    public void should_return_empty_optional_if_not_found() {
+        Throwable throwable = Assertions.catchThrowable(() -> mapping.getUser("NONE"));
+        assertThat(throwable).isInstanceOf(UserNotConfiguredException.class);
+        assertThat(throwable).hasMessage("No configured user for jurisdiction:NONE found");
```
timja: This would be much nicer on junit 5
Could upgrade in a separate PR?
https://github.com/hmcts/idam-java-client/blob/master/build.gradle#L165
https://github.com/hmcts/idam-java-client/blob/master/build.gradle#L150
https://github.com/hmcts/idam-java-client/blob/master/src/test/java/uk/gov/hmcts/reform/idam/client/IdamClientTest.java#L46

bin/send_message.sh
KevinJCross: yeah ... its a first draft. we can enhance this later ?

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CcdCaseRetrieverTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+@RunWith(MockitoJUnitRunner.class)
+public class CcdCaseRetrieverTest {
+    @Mock
+    private CoreCaseDataApi dataApi;
+
+    private CcdCaseRetriever retriever;
+
+    @Test
+    public void should_retrieve_case_successfully() {
```
timja: this test does nothing imo
its just mocking everything and telling it what to return?
REQ

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessorTest.java
```diff
@@ -67,9 +77,8 @@ public void should_not_do_anything_in_notify() {
     @Test
     public void should_return_exceptionally_completed_future_if_unknown_jurisdiction() throws Exception {
         // given
-        given(someMessage.getBody()).willReturn(SampleData.envelopeJson().getBytes());
-        given(userService.getBearerTokenForJurisdiction(any()))
-            .willThrow(new NoUserConfiguredException("foo"));
+        given(someMessage.getBody()).willReturn(TestData.envelopeJson().getBytes());
+        given(authenticator.authenticateForJurisdiction(any())).willThrow(new RuntimeException());
```
timja: Shouldn't it return a `NoUserConfiguredException` ?

src/test/java/GetSASToken.java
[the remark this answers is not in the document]
KevinJCross: Well if we want to use the send_message script then we will be creating this each time. 
I think we need it somewhere ... not sure about here though

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CcdAuthInfo.java
```diff
@@ -0,0 +1,20 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services;
+
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.idam.Credential;
+import uk.gov.hmcts.reform.idam.client.models.UserDetails;
+
+public class CcdAuthInfo {
```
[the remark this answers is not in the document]
timja: I don't really like this approach, its just a generic dumping ground for properties
also it has an `sscsToken` in here the code has to be jurisdiction agnostic and should know nothing about which service.

I would suggest something like:
```
User: {
  userToken: string,
  userDetails: userDetails
}

```

and pass the jurisdiction in separately
You shouldn't need the credential anymore once you have the user object

REQ

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
```diff
@@ -3,45 +3,47 @@
...
-            completableFuture.completeExceptionally(t);
-        }
-        return completableFuture;
+        return completeRunnable(() -> process(message));
+    }
+
+    private void process(IMessage message) {
+        Envelope envelope = parse(message.getBody());
+        CcdAuthInfo authInfo = authenticator.authenticateForJurisdiction(envelope.jurisdiction);
+        CaseDetails theCase = caseRetriever.retrieve(authInfo, envelope.caseRef);
```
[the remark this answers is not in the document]
nitinprabhu: I think that would be nice CCDRetrieverService can deal with Authentication

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
```diff
@@ -3,45 +3,47 @@
...
-        try {
-            Envelope envelope = EnvelopeParser.parse(message.getBody());
-            userService.getBearerTokenForJurisdiction(envelope.jurisdiction);
-            // TODO: use data from envelope to call CCD
-            completableFuture.complete(null);
-        } catch (Throwable t) {
-            completableFuture.completeExceptionally(t);
-        }
-        return completableFuture;
+        return completeRunnable(() -> process(message));
```
nitinprabhu: This change could have been done separately imho :)

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CcdAuthService.java
KevinJCross: true ... will do.

src/test/java/GetSasToken.java
KevinJCross: its a bit odd since I thought I removed it but its still around ... I think something is a bit screwy at the moment Ill investigate.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CcdAuthService.java
KevinJCross: yup, it will through a null pointer and be caught later on.
I have not covered the exception cases yet due to the hurry to unblock this card.
As mentioned Ive got the other card to complete this.

Also the error handling has to be thought out correctly so we ack messages that will always be errors since its pointless letting it appear in the queue again.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CcdCaseRetriever.java
KevinJCross: nah  I think that is overboard .. its retrieve on the CcdCaseRetriever so the context explains the naming.

timja:
Also consider splitting renames and deps update to another PR to simplify refactoring from the logic changes

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CcdCaseRetriever.java
```diff
@@ -0,0 +1,25 @@
...
+    public static final String CASE_TYPE_ID = "Bulk_Scanned";
```
nitinprabhu: You won't be able to write unless you have definition with above case type id imported in AAT

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CcdAuthInfo.java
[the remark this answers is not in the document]
KevinJCross: Im not sure about that ... it cohesion in the code lends to the fact that Jurisdiction is actually part of the User (highly coupled) this means they really belong in the same object.
User<->Jurisdiction in the orchestrator are in practise the same thing.

We could refactor it out such that we have a jurisdiction object that authenticates and then talks to ccd to get a case with a reference.
All this information at the moment has to get passed around together so to me its sensible to encapsulate them in an object. 
User Credential could be removed though.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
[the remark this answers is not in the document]
KevinJCross: yeah ... I think so to Its just if we need this authentication stuff needed later in the processing pipe. Then its better to keep them split up.

src/main/resources/application.yaml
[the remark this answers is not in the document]
KevinJCross: so do I need to do this change ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CcdCaseRetriever.java
KevinJCross: The responsibility for this object is to adapt the information retrieved from the authenticator to the call made to the CaseAPI. 
It will in effect l protect the other classes from the slight complexities with the case api and changes it might introduce. We could inject the authenticator here and it will encapsulate the whole effort to retrieve the case for the jurisdiction using the jurisdiction user.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CompletableHelper.java
KevinJCross: I'm not sold on that.

I personally think it would be a mistake to just catch Exceptions here ... think about the blocking call .get() this might never return if we only catch exceptions and an error occured. 
The other point is that the state of the failed execution would not be caught and presented to thread checking the completable.

src/main/resources/application.yaml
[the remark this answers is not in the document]
KevinJCross: hmmm... I think there will be a problem with the naming here.
CORE_CASE_DATA_API_URL environment var will come out as a property
core.case.data.api.url which will never be able to match core_case_data.api.url ?

src/test/resources/example1.json
KevinJCross: sure ... ill change this.

bin/send_message.sh
rkondratowicz: I am mostly concerned with the security aspect

src/main/resources/application.yaml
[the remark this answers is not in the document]
timja: yes

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/config/JurisdictionToUserMappingTest.java
rkondratowicz: Test method name doesn't seem to match what's being tested

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/config/JurisdictionToUserMappingTest.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    private Credential failFound() {
+        Assert.fail("Found creds when there should be none");
+        return null;
+    }
+
+    private Credential failNotFound() {
```
rkondratowicz: looks like those 2 methods are not used

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CompletableHelper.java
timja: anyone else have thoughts on this?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CcdAuthInfo.java
```diff
@@ -0,0 +1,20 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services;
+
+import uk.gov.hmcts.reform.idam.client.models.UserDetails;
+
+public class CcdAuthInfo {
+    final UserDetails userDetails;
+    final String serviceToken;
```
timja: we shouldn't be passing this around, we should 'generate' one each time, 
Its cached and will either return a valid one or will generate a new one if required
Otherwise the token could expire in between calls

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CcdAuthInfo.java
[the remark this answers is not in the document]
timja: maybe a pair of User and the jurisdiction?
or just
`UserAndJurisdiction` 
or pass in user and jurisdiction?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CcdAuthService.java
timja: err, no I would expect a FeignException to be thrown
which will have a status code on there,
just needs to be logged when that's caught eventually

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CcdCaseRetriever.java
timja: I would think about dropping retriever, its not really doing any work currently, and could be used for other CCD operations?

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/TestData.java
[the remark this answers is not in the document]
rkondratowicz: @KevinJCross note that renaming a class and modifying its content results in github showing in a pr original class being removed and a completely new class added.
This happened (I think, it's hard to compare now...) to `SampleData`, `UserService`, `NoUserConfiguredException`

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/config/JurisdictionToUserMapping.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public Credential getUser(String jurisdiction) {
+
+        Credential credential = users.get(jurisdiction.toLowerCase());
+        if (credential != null) {
+            return credential;
+        } else {
+            throw new UserNotConfiguredException(jurisdiction);
+        }
```
doncem: just a thought on using something other than usual `get if/else` approach:

```java
response.computeIfAbsent(
    jurisdiction.toLowerCase(),
    key -> {
        throw new UserNotConfiguredException(key);
    }
);
```

This also will not add it to the `users` map

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/config/UserNotConfiguredException.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.config;
+
+public class UserNotConfiguredException extends RuntimeException {
+    public UserNotConfiguredException(String jurisdiction) {
+        super("No configured user for jurisdiction:" + jurisdiction + " found");
```
doncem: perhaps don't need the `:` in there

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CcdAuthInfo.java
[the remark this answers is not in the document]
KevinJCross: sure I rename .

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/TestData.java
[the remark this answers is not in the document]
KevinJCross: @rkondratowicz  I've re-applied these changes to master to remove the "noise"
Ill open a new PR for this.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
KevinJCross: done

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CompletableHelper.java
doncem: i personally still don't support whole block: assign future, try/catch with separate suppose "long" operation and separately completing future. it's just.. why?
java8 introduced `CompletableFuture` api for a reason and very much reliable. the main recommendation is to use your own executioner so to not fork from common pool - all should be good. even the same thing was done with service bus library's latest version jump introduced in [bulk scan processor](https://github.com/hmcts/bulk-scan-processor/pull/189)

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CompletableHelper.java
```diff
@@ -0,0 +1,21 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services;
```
doncem: object name says `CompletableHelper` and only has some abstract runnable handling. does suggest to perhaps separate out to, let's say, `helper` package and not `services`, does it not?

KevinJCross:
Moving this to a new PR to remove "extra" changes.
https://github.com/hmcts/bulk-scan-orchestrator/pull/47

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/CompletableHelper.java
doncem: on top of that, future will do exactly the same and catch `Throwable` 😉

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/Application.java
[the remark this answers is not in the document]
KevinJCross: done

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/config/JurisdictionToUserMapping.java
```diff
@@ -0,0 +1,42 @@
...
+public class JurisdictionToUserMapping {
+
+    private Map<String, Credential> users;
+
+    public void setUsers(Map<String, Map<String, String>> users) {
+        this.users = users
+            .entrySet()
+            .stream()
+            .map(this::toEntry)
+            .collect(HashMap::new, (map, pair) -> map.put(pair.first().toLowerCase(), pair.second()), HashMap::putAll);
```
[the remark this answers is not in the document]
KevinJCross: yup Ive fixed this.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/config/JurisdictionToUserMapping.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    private Pair<String, Credential> toEntry(Map.Entry<String, Map<String, String>> entry) {
```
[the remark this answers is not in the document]
KevinJCross: sure ... I was attempting to create a map entry in the first pass.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/config/JurisdictionToUserMapping.java
```diff
@@ -0,0 +1,42 @@
...
+        String key = entry.getKey().toUpperCase();
+        Credential cred = new Credential(entry.getValue().get("username"), entry.getValue().get("password"));
```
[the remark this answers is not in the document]
KevinJCross: sure ... we should add more verification here. 
I was going to add an Enum of Jurisdiction but that was frowned on.
I guess we could have not password but no user should be an error.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/config/JurisdictionToUserMapping.java
KevinJCross: sure ... I was feeling harried about getting this card completed. I prefer your solution 😄

PR #48, opened 2018-09-28
Move CCD classes to packages

doncem:
### Change description ###

Preparing for future code changes. Separation of ccd related classes from bulk scan orchestrator main service package to inner ccd one. Will be majorly used for case record creations, event publishers, callback handlers, etc

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/RootControllerTest.java
```diff
@@ -1,24 +0,0 @@
...
-
-@RunWith(SpringRunner.class)
-@WebMvcTest
-public class RootControllerTest {
-
-    @Autowired
-    private transient MockMvc mockMvc;
-
-    @Test
-    public void call_to_root_endpoint_should_result_with_204_response() throws Exception {
```
nitinprabhu: any reason we removed this test ?

doncem: because it is an integration test and it exists in integration test suite matching char by char the whole file :D.

p.s. i did include as a separate commit here https://github.com/hmcts/bulk-scan-orchestrator/pull/48/commits/ff0276ec72cd0f743a9ded7d8e7a33b79826e379.

noticed as unit tests took 3mins to run getting stuck on this test. and integration tests have only that one too

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdAuthenticator.java
```diff
@@ -1,15 +1,15 @@
-package uk.gov.hmcts.reform.bulkscan.orchestrator.services;
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd;
 
 import uk.gov.hmcts.reform.idam.client.models.UserDetails;
 
 import java.util.function.Supplier;
 
-public class Authenticator {
+public class CcdAuthenticator {
```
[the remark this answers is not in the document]
doncem: the way i see it is used in ccd client only as a container of tokens/ids. once imported in the class and used in whichever place a reviewer is aware this is ccd related. in case we re-use this for something else - both authenticator and authenticator factory can be moved out. oh, and authenticator factory was already prefixed with ccd. so just picked up one change to align both classes

PR #49, opened 2018-09-28
Adding Kotlin dependencies.

KevinJCross:
To help out with easing the integration test writing I'm adding the Kotlin support.

### Change description ###
Adding kotlin dependencies to enable better test writing and conscise code.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,6 +8,7 @@ plugins {
   id 'org.owasp.dependencycheck' version '3.3.1'
   id 'com.github.ben-manes.versions' version '0.20.0'
   id 'org.sonarqube' version '2.6.2'
+  id 'org.jetbrains.kotlin.jvm' version '1.2.71'
 }
 
 group = 'uk.gov.hmcts.reform'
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -140,6 +152,7 @@ def versions = [
 ]
 
 dependencies {
+  compile group: 'org.jetbrains.kotlin', name:'kotlin-stdlib-jdk8'
```
doncem: is this dependency required to be in application jar? perhaps just needed for development and can be marked as `compileOnly`

KevinJCross: yup .. its for the kotlin standard libraries, which includes a lot of useful extensions to java standard libraries.

timja: kotlin docs say use compile:
https://kotlinlang.org/docs/reference/using-gradle.html

KevinJCross:
wow ... the "Codacy/PR Quality Review Pending" is taking a while. I wonder if its building and index or cache (21 mins and counting)?

PR #50, opened 2018-09-28
Introduce factory for case record creator

doncem:
### Change description ###

Proposition of abstraction of case record creation used in submitting evidence to CCD. API accepts `CaseDataContent` so there is a separate interface holding just that information. For internal orchestrator knowledge - a separate `CaseRecord` interface is in place holding information about available `Record` and common `Construct` to be **only** used in privately by inheritance.

Factory privately constructs collection of available `Record`s by `Construct` reference only and each time builds up new `CaseDataContent` aware `Record` instance to be used in event publisher.

The exact fields to construct are subject to change once case records are actually created and successfully pushed to CCD. This PR establishes separation as a proposal

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecord.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecord.java
@@ -1,0 +1,19 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model;
+
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.model.Envelope;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+
+public interface CaseRecord extends CaseDataCreator {
+
+    @FunctionalInterface
+    interface Construct {
+        CaseDataCreator apply(Envelope envelope, CaseDetails caseDetails);
+    }
+
+    enum Record {
+        EXCEPTION_RECORD,
+        SUPPLEMENTARY_EVIDENCE,
+    }
+
+    Record getCaseRecordIdentifier();
+}
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecordFactory.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecordFactory.java
@@ -1,11 +1,12 @@
 package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model;
 
+import com.google.common.collect.ImmutableMap;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model.caserecord.ExceptionRecord;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model.caserecord.SupplementaryEvidenceRecord;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.model.Envelope;
 import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
 
-import java.util.HashMap;
+import java.util.EnumMap;
 import java.util.Map;
 
 import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model.CaseRecord.Construct;
@@ -13,12 +14,10 @@ import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model.CaseR
 
 public final class CaseRecordFactory {
 
-    private static final Map<Record, Construct> AVAILABLE_MODELS = new HashMap<>();
-
-    static {
-        AVAILABLE_MODELS.put(Record.EXCEPTION_RECORD, ExceptionRecord::new);
-        AVAILABLE_MODELS.put(Record.SUPPLEMENTARY_EVIDENCE, SupplementaryEvidenceRecord::new);
-    }
+    private static final Map<Record, Construct> AVAILABLE_MODELS = new EnumMap<>(ImmutableMap.of(
+        Record.EXCEPTION_RECORD, ExceptionRecord::new,
+        Record.SUPPLEMENTARY_EVIDENCE, SupplementaryEvidenceRecord::new
+    ));
 
     public static CaseDataCreator getCaseDataCreator(Envelope envelope, CaseDetails caseDetails) {
         Record record;
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecordFactory.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecordFactory.java
@@ -19,12 +19,12 @@ public final class CaseRecordFactory {
         Record.SUPPLEMENTARY_EVIDENCE, SupplementaryEvidenceRecord::new
     ));
 
-    public static CaseDataCreator getCaseDataCreator(Envelope envelope, CaseDetails caseDetails) {
+    public static CaseDataCreator getCaseDataCreator(Envelope envelope, CaseDetails existingCase) {
         Record record;
 
         switch (envelope.classification) {
             case SUPPLEMENTARY_EVIDENCE:
-                record = Record.SUPPLEMENTARY_EVIDENCE;
+                record = existingCase == null ? Record.EXCEPTION_RECORD : Record.SUPPLEMENTARY_EVIDENCE;
 
                 break;
             case EXCEPTION:
@@ -35,7 +35,7 @@ public final class CaseRecordFactory {
                 break;
         }
 
-        return AVAILABLE_MODELS.get(record).apply(envelope, caseDetails);
+        return AVAILABLE_MODELS.get(record).apply(envelope, existingCase);
     }
 
     private CaseRecordFactory() {
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecordFactory.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecordFactory.java
@@ -1,0 +1,45 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model;
+
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model.caserecord.ExceptionRecord;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model.caserecord.SupplementaryEvidenceRecord;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.model.Envelope;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model.CaseRecord.Construct;
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model.CaseRecord.Record;
+
+public final class CaseRecordFactory {
+
+    private static final Map<Record, Construct> AVAILABLE_MODELS = new HashMap<>();
+
+    static {
+        AVAILABLE_MODELS.put(Record.EXCEPTION_RECORD, ExceptionRecord::new);
+        AVAILABLE_MODELS.put(Record.SUPPLEMENTARY_EVIDENCE, SupplementaryEvidenceRecord::new);
+    }
+
+    public static CaseDataCreator getCaseDataCreator(Envelope envelope, CaseDetails caseDetails) {
+        Record record;
+
+        switch (envelope.classification) {
+            case SUPPLEMENTARY_EVIDENCE:
+                record = Record.SUPPLEMENTARY_EVIDENCE;
+
+                break;
+            case EXCEPTION:
+            case NEW_APPLICATION:
+            default:
+                record = Record.EXCEPTION_RECORD;
+
+                break;
+        }
+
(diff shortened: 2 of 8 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecordFactory.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model.CaseRecord.Construct;
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model.CaseRecord.Record;
+
+public final class CaseRecordFactory {
+
+    private static final Map<Record, Construct> AVAILABLE_MODELS = new HashMap<>();
```
KevinJCross: There is a map type for enum to object mapping `EnumMap`

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecordFactory.java
```diff
@@ -0,0 +1,44 @@
...
+
+    static {
+        AVAILABLE_MODELS.put(Record.EXCEPTION_RECORD, ExceptionRecord::new);
```
KevinJCross: Rather do this with say an inmutable map ?
```java
ImmutableMap.of(
    EXCEPTION_RECORD, ExceptionRecord::new
    SUPPLEMENTARY_EVIDENCE,   SupplementaryEvidenceRecord::new 
)

Or even better just bake it into the enum ?
enum{
EXCEPTION_RECORD(ExceptionRecord::new); 
  private final constuct;
  Record( Construct constructor){
     construct = constructor
  } 
  public Something construct(){
     return construct()
  }
}
Unfortunately you lose the type information ... so it might be simpler to use a sealed Kotlin class for this ( its actually simpler in the long run).

File: src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecordFactory.java [hunk 3]
```diff
@@ -0,0 +1,44 @@
...
+        AVAILABLE_MODELS.put(Record.SUPPLEMENTARY_EVIDENCE, SupplementaryEvidenceRecord::new);
+    }
+
+    public static CaseDataCreator getCaseDataCreator(Envelope envelope) {
+        Construct caseRecordConstruct;
+
+        switch (envelope.classification) {
```
KevinJCross: why not just get it from the map directly ?
AVAILABLE_MODELS.get(envelope.classification).appy(evelope) ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecordFactory.java
doncem: Cool. I'll go with `ImmutableEnumMap` then. In case there will be requirement of >5 records - we can alter the code

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecordFactory.java
doncem: Classification differ. And anyway, this is entry point to decide which record to create. not only envelope classification decides that

rkondratowicz:
Would be best for someone actually working on stuff related on this to review.
Or wait for the design to emerge **after** we know exactly what we need to do imho.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecordFactory.java
```diff
@@ -0,0 +1,47 @@
...
+
+public final class CaseRecordFactory {
+
+    private static final Map<Record, Construct> AVAILABLE_MODELS;
+
+    static {
+        AVAILABLE_MODELS = new EnumMap<>(ImmutableMap.of(
+            Record.EXCEPTION_RECORD, ExceptionRecord::new,
+            Record.SUPPLEMENTARY_EVIDENCE, SupplementaryEvidenceRecord::new
+        ));
```
rkondratowicz: this is now a (nicely formatted) one liner and could be moved to field declaration I think
ie:
```
private static final Map<Record, Construct> AVAILABLE_MODELS = new ...
```

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecord.java
```diff
@@ -0,0 +1,19 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model;
+
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.model.Envelope;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+
+public interface CaseRecord extends CaseDataCreator {
+
+    @FunctionalInterface
+    interface Construct {
+        CaseDataCreator apply(Envelope envelope, CaseDetails caseDetails);
```
KevinJCross: Why not just a bi-Function ?
Its recommended to use the standard java  functional interfaces

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecord.java
```diff
@@ -0,0 +1,19 @@
...
+    }
+
+    enum Record {
```
KevinJCross: This should probably be by itself otherwise rename it .....
CaseRecord.Record does not read well .... but 
CaseRecord.Type does.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecord.java
```diff
@@ -0,0 +1,19 @@
...
+        EXCEPTION_RECORD,
+        SUPPLEMENTARY_EVIDENCE,
+    }
+
+    Record getCaseRecordIdentifier();
```
KevinJCross: This could be getType ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/caserecord/ExceptionRecord.java
```diff
@@ -0,0 +1,32 @@
...
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+import uk.gov.hmcts.reform.ccd.client.model.Classification;
+
+public class ExceptionRecord implements CaseRecord {
+
+    private final Envelope envelope;
+    private final CaseDetails caseDetails;
+
+    public ExceptionRecord(Envelope envelope, CaseDetails caseDetails) {
+        this.envelope = envelope;
```
KevinJCross: Null checks ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/caserecord/ExceptionRecord.java
```diff
@@ -0,0 +1,32 @@
...
+        this.caseDetails = caseDetails;
+    }
+
+    @Override
+    public CaseDataContent createDataContent(
+        Classification classification,
+        String eventToken,
+        boolean ignoreWarning
+    ) {
+        return null;
```
KevinJCross: Hmm should there be an exception rather than null ... or a Null object ... 
I.e. a CaseDataContent that has nothing ... 
Or an optional ? 

Or have an interface that does not have get CaseRecord and compose the interfaces ....
EnvelopeContainer
CaseDetailsContainer
CaseDataContainer 
and have a visitor to to retrieve the info or use a instance of ?

❌

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/caserecord/SupplementaryEvidenceRecord.java
```diff
@@ -0,0 +1,32 @@
...
+        this.caseDetails = caseDetails;
+    }
+
+    @Override
+    public CaseDataContent createDataContent(
+        Classification classification,
+        String eventToken,
+        boolean ignoreWarning
+    ) {
+        return null;
```
KevinJCross: does this not have CaseData ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/caserecord/ExceptionRecord.java
doncem: Envelope already has been accessed multiple times before reaching this point. Should I just add null check for it at the beginning anyway?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/caserecord/ExceptionRecord.java
doncem: Should have thrown exception in the first place. Will add

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/caserecord/ExceptionRecord.java
```diff
@@ -0,0 +1,32 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model.caserecord;
+
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model.CaseRecord;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.model.Envelope;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDataContent;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+import uk.gov.hmcts.reform.ccd.client.model.Classification;
+
+public class ExceptionRecord implements CaseRecord {
```
nitinprabhu: well exception record is also a case record

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecordFactory.java
```diff
@@ -0,0 +1,44 @@
...
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.model.Envelope;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+
+import java.util.EnumMap;
+import java.util.Map;
+
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model.CaseRecord.Construct;
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model.CaseRecord.Record;
+
+public final class CaseRecordFactory {
```
nitinprabhu: You will either have a exception record or update the case why do we need a factory ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/caserecord/SupplementaryEvidenceRecord.java
```diff
@@ -0,0 +1,32 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model.caserecord;
+
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.model.CaseRecord;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.model.Envelope;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDataContent;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+import uk.gov.hmcts.reform.ccd.client.model.Classification;
+
+public class SupplementaryEvidenceRecord implements CaseRecord {
```
nitinprabhu: There is nothing called as SupplementaryEvidenceRecord either it is an exception or case

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecord.java
nitinprabhu: I think this is classification

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/model/CaseRecordFactory.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private static final Map<Record, Construct> AVAILABLE_MODELS = new EnumMap<>(ImmutableMap.of(
+        Record.EXCEPTION_RECORD, ExceptionRecord::new,
+        Record.SUPPLEMENTARY_EVIDENCE, SupplementaryEvidenceRecord::new
+    ));
+
+    public static CaseDataCreator getCaseDataCreator(Envelope envelope, CaseDetails existingCase) {
+        Record record;
+
+        switch (envelope.classification) {
```
nitinprabhu: what if you have a new classification ? violates open/close principle ?

doncem:
Giving the feedback I am again in the loss. Beginning with simple concept of describing what is what. Different sources form sentences and gives false assumptions. I will refrain from doing this task and accept Rafal's suggestion about pattern introduction for later. Any1 else can jump in and do this extra bit of work especially if have better knowledge about domain definitions

PR #52, opened 2018-09-28
*experimental*: moving to a delegated property to simplify the authenticator

KevinJCross:
https://tools.hmcts.net/jira/browse/RPE-671

### Change description ###
This is an attempt to see how a kotlin delegated property could simplify the Authenticator.
It might be a little to deep in kotlin for now but I think its worth a review.
Thoughts ?
Can always leave this out (till later maybe) if it seems too obscure.

```
[ ] Yes
[ x] No
```


luigibk:
Can we use Kotlin? I thought we could not...

timja:
fine to start using it I think, but this example isn't the best one to go for.
Code looks more complex than before

KevinJCross:
Sure this was just an experiment ... so Ill drop it anyway.

The simpler looking version is not to use the property delegates but just to put.
```kotlin
class CcdAuthenticator(
        private val serviceTokenSupplier: Supplier<String>,
        internal val userDetails: UserDetails,
        private val userTokenSupplier: Supplier<String>
) {

    val userToken: String
        get() = this.userTokenSupplier.get()

    val serviceToken: String
        get() = this.serviceTokenSupplier.get()

    companion object {
        fun from(
                serviceTokenSupplier: Supplier<String>,
                userDetails: UserDetails,
                userTokenSupplier: Supplier<String>): CcdAuthenticator {
            return CcdAuthenticator(serviceTokenSupplier, userDetails, userTokenSupplier)
        }
    }
}
```

PR #54, opened 2018-09-28
Added first part of the integration tests for message processing.

KevinJCross:
### Change description ###
Added the first part of the integration test to enable "integration" style testing.
Id prefer to start up a RabbitMq instance locally instead of replacing the IMessageReceiver bean.
That might come a bit later I think.
This does highlight something a bit odd with the logging ... its missing the Exception message and type.

```
[ ] Yes
[x] No
```


src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CaseRetrievalTest.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.context.annotation.Primary
+import org.springframework.test.context.junit.jupiter.SpringExtension
+import uk.gov.hmcts.reform.bulkscan.orchestrator.Application
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.IMessageReceiverProvider
+import java.util.concurrent.CountDownLatch
+
+val latch = CountDownLatch(1)
+
+@Import(Application::class)
+open class TestConfig {
```
timja: I think if you apply the kotlin-spring compiler plugin this doesn't need to be open
https://kotlinlang.org/docs/tutorials/spring-boot-restful.html
https://kotlinlang.org/docs/reference/compiler-plugins.html#kotlin-spring-compiler-plugin

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/IMessageReceiverProvider.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus;
+
+import com.microsoft.azure.servicebus.IMessageReceiver;
+
+public interface IMessageReceiverProvider {
```
timja: Remove the `I` please, thats a .NET style thing

rkondratowicz: is this interface just for tests? can we mock `ReceiverProvider`?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CaseRetrievalTest.kt
KevinJCross: sure ... we can improve :D

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/IMessageReceiverProvider.java
KevinJCross: yup it is ... this is so we can replace the connection to the SB

KevinJCross: @timja Well the thing is that it is providing an IMessageRetriever ... it not an interface really. Its the naming of the type of retriever ... lol
Il rename it anyway

build.gradle
```diff
@@ -176,15 +177,17 @@ dependencies {
 
   compile group: 'com.microsoft.azure', name: 'azure-servicebus', version: '1.2.7'
 
-  testCompile group: 'org.springframework.boot', name: 'spring-boot-starter-test', version: versions.springBoot
+  annotationProcessor group: 'org.springframework.boot', name: 'spring-boot-configuration-processor'
```
rkondratowicz: it's already added below

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/MessageProcessorTest.java
```diff
@@ -25,7 +25,8 @@
 
 @RunWith(MockitoJUnitRunner.class)
 public class MessageProcessorTest {
-    @Mock private ReceiverProvider receiverProvider;
+    @Mock
+    private ReceiverFactory receiverProvider;
```
rkondratowicz: keep formatting consistent

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/MessageProcessor.java
```diff
@@ -7,26 +7,26 @@
...
-import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.ReceiverProvider;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.MessageReceiverFactory;
 
 @Service
 public class MessageProcessor {
 
     private static final Logger logger = LoggerFactory.getLogger(MessageProcessor.class);
 
-    private final ReceiverProvider receiverProvider;
+    private final MessageReceiverFactory receiverProvider;
```
rkondratowicz: field name no longer matches type name. but... was this class name necessary? (and method name rename)

KevinJCross: Its now injecting via an interface ... so it can be replaced with a bean in the integration test.
Ideally we would not need to do this but since we are using shared infrastructure for testing we are forced to mock this out otherwise it would be too fragile.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CaseRetrievalTest.kt
KevinJCross: done

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/IMessageReceiverProvider.java
KevinJCross: made it a messageReceiverFactory :D

PR #56, opened 2018-10-01
[PREVIEW] Use new document model

rkondratowicz:
Handle new field that is sent by the core component.


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/model/Envelope.java
```diff
@@ -13,21 +13,21 @@
     public final String jurisdiction;
     public final String zipFileName;
     public final Classification classification;
-    public final List<String> docUrls;
+    public final List<Document> documents;
```
luigibk: This list is not really final as in "immutable". It's just final as in "it cannot be replaced with another list".

PR #58, opened 2018-10-02
Use bulkscan user

luigibk:
https://tools.hmcts.net/jira/browse/RPE-708

### Change description ###
Use bulkscan user

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -23,6 +23,7 @@ locals {
     // configures a user for a jurisdiction
     // add secrets to all bulk-scan vaults in the form idam-users-<jurisdiction>-username idam-users-<jurisdiction>-password
     SSCS = "idam-users-sscs"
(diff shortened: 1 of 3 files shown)


infrastructure/main.tf
```diff
@@ -23,6 +23,7 @@ locals {
     // configures a user for a jurisdiction
     // add secrets to all bulk-scan vaults in the form idam-users-<jurisdiction>-username idam-users-<jurisdiction>-password
     SSCS = "idam-users-sscs"
```
nitinprabhu: Do we need this ?

luigibk: Excel spreadsheet updated. I've been told by @timja we do need that.

timja: well, no point removing it.
the creds are there for SSCS, and we're about to spike out integrating with them

PR #59, opened 2018-10-03
bugfix- correctly throw error if there is no configuration.

KevinJCross:
### Change description ###
Throw correct error if no user configuration is supplied


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/JurisdictionToUserMapping.java
```diff
@@ -11,7 +12,7 @@
 @ConfigurationProperties(prefix = "idam")
 public class JurisdictionToUserMapping {
 
-    private Map<String, Credential> users;
+    private Map<String, Credential> users = ImmutableMap.of();
```
nitinprabhu: You can do `Collections.emptyMap()` looks more readable imo

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/config/JurisdictionToUserMappingTest.java
```diff
@@ -41,4 +41,12 @@ public void should_throw_exception_if_not_found() {
             .isInstanceOf(NoUserConfiguredException.class)
             .hasMessage("No user configured for jurisdiction: none");
     }
+
+    @Test
+    public void should_throw_exception_if_none_configured() {
```
nitinprabhu: should we say `should_throw_no_user_configured_exception_when_jurisdiction_is_not_found` ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/JurisdictionToUserMapping.java
rkondratowicz: Good point, we reasign this field in the setter as well, 'immutable' part doesn't make much sense here

KevinJCross: Actually I can't ...  it will through a not supported exception. 
I tried this first and the test failed.

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/config/JurisdictionToUserMappingTest.java
KevinJCross: Not really this is saying that if there is no configuration it should throw the exception.

KevinJCross: done

PR #60, opened 2018-10-03
feature/add_example_message_file_for_IT_test

KevinJCross:
### Change description ###
Adding a message in a .json file (easier to read and manually modify)
Also adding tests so that they are always parseable. 
Added a String interface to make it easier for testing.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/EnvelopeParser.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/EnvelopeParser.java
@@ -26,6 +26,10 @@ public class EnvelopeParser {
         }
     }
 
+    public static Envelope parse(String json) {
+        return parse(json.getBytes());
+    }
+
     private EnvelopeParser() {
         // utility class
     }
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/EnvelopeParser.java
```diff
@@ -26,6 +26,10 @@ public static Envelope parse(byte[] bytes) {
         }
     }
 
+    public static Envelope parse(String json) {
```
KevinJCross: an easier interface for testing.

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/SampleData.java
```diff
@@ -38,23 +43,39 @@
...
+                        .put("type", "doc_type")
+                        .put("scanned_at", toIso8601(Instant.EPOCH))
+                        .put("url", "https://example.gov.uk/123")
+                    )
                 )
-            )
-            .toString();
+                .toString();
+        } catch (Exception e) {
+            throw new RuntimeException("Could not make envelopeJson", e);
```
KevinJCross: added this so that we don't have to keep puting Exception on the test methods.

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/SampleData.java
```diff
@@ -38,23 +43,39 @@
...
+        }
+    }
+
+    public static String fromFile(String file) {
```
KevinJCross: just read it from a file on the class path.

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/SampleData.java
rkondratowicz: we always do anyway

KevinJCross: Its just more boilerplate and cumbersome. Why if we dont need it ?

PR #61, opened 2018-10-04
Enable use of integrationtest profile to simplify tests.

KevinJCross:
### Change description ###
Using the integrationtest profile removes a lot of the configuration boilerplate on the tests.
Maybe we should consider having a **"local"** profile to fill this out for docker or aks local setups.


src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/IntegrationTestConfig.kt
```diff
@@ -6,15 +6,13 @@ import com.github.tomakehurst.wiremock.core.WireMockConfiguration
 import com.microsoft.azure.servicebus.IMessageReceiver
 import org.mockito.Mockito
 import org.springframework.beans.factory.annotation.Value
-import org.springframework.context.annotation.Bean
-import org.springframework.context.annotation.Configuration
-import org.springframework.context.annotation.Import
-import org.springframework.context.annotation.Primary
+import org.springframework.context.annotation.*
```
KevinJCross: hmmm * ... will have to change my settings.

PR #62, opened 2018-10-08
Update API calls to ccd with try/catch block

doncem:

[Create exception record for supplementary evidence type](https://tools.hmcts.net/jira/browse/RPE-656)

### Change description ###

Helper PR to deal with the situation when case is not found in the land of CCD

- moved logging straight after the actual call to retrieve
- save api hit in case `caseRef` is empty
- wrap API calls with try/catch

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CaseRetriever.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CaseRetriever.java
@@ -6,10 +6,12 @@ import feign.FeignException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.cloud.client.loadbalancer.LoadBalanced;
-import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Service;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.exceptions.CcdClientException;
 import uk.gov.hmcts.reform.ccd.client.CoreCaseDataApi;
 import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+
+import static org.springframework.http.HttpStatus.NOT_FOUND;
 
 @Service
 public class CaseRetriever {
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CaseRetriever.java
```diff
@@ -8,28 +15,101 @@
...
+    private void logResponse(String jurisdiction, String caseRef, HttpStatus status, String originalMessage) {
+        if (HttpStatus.NOT_FOUND.equals(status)) {
+            log.info(
+                "Not found case ref:{} for {} jurisdiction {}. Error: {}",
+                caseRef,
+                jurisdiction,
+                originalMessage
+            );
+        } else if (status.is4xxClientError()) {
+            log.warn(
```
timja: I would think that any error back from CCD other than not found is an error, i.e. either we've done something wrong or a service team has.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CaseRetriever.java
```diff
@@ -17,19 +26,43 @@
...
+            return caseDetails;
+        } catch (FeignException exception) {
+            if (exception.status() == NOT_FOUND.value()) {
+                log.info("Case not found. Ref:{}, jurisdiction:{}", caseRef, jurisdiction);
+
+                return null;
+            } else {
+                throw new CcdClientException(exception.getMessage(), exception);
+            }
+        } catch (Throwable exception) {
```
timja: Why're we catching throwable? and do we need to catch anything other than FeignException here?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CaseRetriever.java
```diff
@@ -17,19 +26,43 @@
...
+            );
+
+            logCaseDetails(caseDetails);
+
+            return caseDetails;
+        } catch (FeignException exception) {
+            if (exception.status() == NOT_FOUND.value()) {
+                log.info("Case not found. Ref:{}, jurisdiction:{}", caseRef, jurisdiction);
+
+                return null;
```
timja: Would optional make more sense than null here?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/exceptions/CcdClientException.java
```diff
@@ -0,0 +1,8 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.exceptions;
+
+public class CcdClientException extends RuntimeException {
+
+    public CcdClientException(String message, Throwable cause) {
```
timja: Does this give us any benefit over FeignException?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CaseRetriever.java
timja: just this here?

doncem: A good choice to prevent `null` values but not leaning towards it. Unnecessary wrapper - both branches in same method call will have to be wrapped. Then envelope processor has a pre-check before hitting case retriever. So overall there are 3 branches as of this moment. It is easier to handle `null`s at this point.

For future, in case there will be no need to have information about `CaseDetail` in order to make event calls, I can see the benefit of `Optional` being used

PR #63, opened 2018-10-10
Support adding supplementary evidence based on queue messages

lgonczar:

### Change description ###

Support adding supplementary evidence based on queue messages. Not included here:
- functional and integration tests (coming...)
- extracting common interface for event creators (too early)

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/SupplementaryEvidenceCreator.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/SupplementaryEvidenceCreator.java
@@ -38,7 +38,7 @@ public class SupplementaryEvidenceCreator {
         StartEventResponse startEventResponse =
             startEvent(info, envelope.jurisdiction, envelope.caseRef);
 
-        log.info("Started {} event for case {}", startEventResponse.getEventId(), envelope.caseRef);
+        log.debug("Started {} event for case {}", startEventResponse.getEventId(), envelope.caseRef);
 
         CaseDataContent caseDataContent = prepareCaseDataContent(
             startEventResponse.getToken(),
@@ -47,7 +47,7 @@ public class SupplementaryEvidenceCreator {
 
         submitEvent(info, envelope.jurisdiction, envelope.caseRef, caseDataContent);
 
-        log.info("Submitted {} event for case {}", startEventResponse.getEventId(), envelope.caseRef);
+        log.debug("Submitted {} event for case {}", startEventResponse.getEventId(), envelope.caseRef);
     }
 
     private CaseDataContent prepareCaseDataContent(
@@ -59,7 +59,6 @@ public class SupplementaryEvidenceCreator {
             .event(Event.builder()
                 .id(EVENT_TYPE_ID)
                 .summary("Attach scanned documents")
-                .description("Attach scanned documents")
                 .build())
             .data(supplementaryEvidence)
             .build();
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/SupplementaryEvidenceCreator.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/SupplementaryEvidenceCreator.java
@@ -1,0 +1,101 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Service;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd.SupplementaryEvidence;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd.mappers.SupplementaryEvidenceMapper;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.model.Envelope;
+import uk.gov.hmcts.reform.ccd.client.CoreCaseDataApi;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDataContent;
+import uk.gov.hmcts.reform.ccd.client.model.Event;
+import uk.gov.hmcts.reform.ccd.client.model.StartEventResponse;
+
+@Service
+public class SupplementaryEvidenceCreator {
+
+    private static final Logger log = LoggerFactory.getLogger(SupplementaryEvidenceCreator.class);
+
+    private static final String CASE_TYPE_ID = "Bulk_Scanned";
+    private static final String EVENT_TYPE_ID = "attachScannedDocs";
+
+    private final CcdAuthenticatorFactory authenticatorFactory;
+    private final CoreCaseDataApi coreCaseDataApi;
+
+    SupplementaryEvidenceCreator(
+        CcdAuthenticatorFactory authenticatorFactory,
+        CoreCaseDataApi coreCaseDataApi
+    ) {
+        this.authenticatorFactory = authenticatorFactory;
+        this.coreCaseDataApi = coreCaseDataApi;
+    }
+
+    public void createSupplementaryEvidence(Envelope envelope) {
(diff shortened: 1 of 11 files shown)


lgonczar:
Will split into two

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/SupplementaryEvidenceCreator.java
```diff
@@ -0,0 +1,101 @@
...
+
+    public void createSupplementaryEvidence(Envelope envelope) {
+        log.info("Creating supplementary evidence for case {}", envelope.caseRef);
+
+        CcdAuthenticator info = authenticatorFactory.createForJurisdiction(envelope.jurisdiction);
+
+        StartEventResponse startEventResponse =
+            startEvent(info, envelope.jurisdiction, envelope.caseRef);
+
+        log.info("Started {} event for case {}", startEventResponse.getEventId(), envelope.caseRef);
```
timja: looks debug like to me?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/SupplementaryEvidenceCreator.java
```diff
@@ -0,0 +1,101 @@
...
+
+        CaseDataContent caseDataContent = prepareCaseDataContent(
+            startEventResponse.getToken(),
+            SupplementaryEvidenceMapper.fromEnvelope(envelope)
+        );
+
+        submitEvent(info, envelope.jurisdiction, envelope.caseRef, caseDataContent);
+
+        log.info("Submitted {} event for case {}", startEventResponse.getEventId(), envelope.caseRef);
```
timja: same here?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/SupplementaryEvidenceCreator.java
```diff
@@ -0,0 +1,101 @@
...
+    private CaseDataContent prepareCaseDataContent(
+        String eventToken,
+        SupplementaryEvidence supplementaryEvidence
+    ) {
+        return CaseDataContent.builder()
+            .eventToken(eventToken)
+            .event(Event.builder()
+                .id(EVENT_TYPE_ID)
+                .summary("Attach scanned documents")
+                .description("Attach scanned documents")
```
timja: no point putting the same summary and description in just drop the description imo

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/SupplementaryEvidenceCreatorTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @Before
+    public void setUp() {
+        given(authenticatorFactory.createForJurisdiction(any())).willReturn(AUTH_DETAILS);
+
+        creator = new SupplementaryEvidenceCreator(authenticatorFactory, coreCaseDataApi);
+    }
+
+    @Test
+    public void createSupplementaryEvidence_starts_and_submits_event() {
```
timja: what benefit does this give us?
The verifications seem to be around that CCD calls happened, to we care how they happy?
What we're interested in is did the update persist in CCD, not the api calls needed to get it there?

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/SupplementaryEvidenceCreatorTest.java
```diff
@@ -0,0 +1,131 @@
...
+            .given(coreCaseDataApi)
+            .startEventForCaseWorker(any(), any(), any(), any(), any(), any(), any());
+
+        Throwable actualException = catchThrowable(
+            () -> creator.createSupplementaryEvidence(SampleData.envelope(1))
+        );
+
+        assertThat(actualException).isSameAs(expectedException);
+
+        verify(coreCaseDataApi, never())
```
timja: do we need this call?,
 do we need this test?

that call is very unlikely to throw a runtimeexception, 99.99999% of the time it would be a feign exception

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/SupplementaryEvidenceCreator.java
lgonczar: 👍 wasn't sure that was allowed

timja: summary and description are both optional fields

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/SupplementaryEvidenceCreatorTest.java
lgonczar: True, although as an external client is provided and we assume it's reliable, checking if it's used correctly (i.e. called with the right parameters) adds, in my opinion, some value. It doesn't check if the configuration is correct, of course, and this test doesn't replace integration tests.

timja: Sure I guess in our case its worth checking we are using the right jurisdiction and user token

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/SupplementaryEvidenceCreator.java
```diff
@@ -0,0 +1,101 @@
...
+    SupplementaryEvidenceCreator(
+        CcdAuthenticatorFactory authenticatorFactory,
+        CoreCaseDataApi coreCaseDataApi
+    ) {
+        this.authenticatorFactory = authenticatorFactory;
+        this.coreCaseDataApi = coreCaseDataApi;
+    }
+
+    public void createSupplementaryEvidence(Envelope envelope) {
+        log.info("Creating supplementary evidence for case {}", envelope.caseRef);
```
doncem: since supplementary evidence processing is potentially going to master, this whole block should be skipped/erred in case `envelope.caseRef` is empty

lgonczar: Will handle null

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/SupplementaryEvidenceCreatorTest.java
lgonczar: It might be too much. Will remove.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/SupplementaryEvidenceCreator.java
lgonczar: Actually, I've just checked and as of now envelopes without case_ref won't even be parsed, as the field is required.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/SupplementaryEvidenceCreator.java
```diff
@@ -0,0 +1,100 @@
...
+
+        CcdAuthenticator info = authenticatorFactory.createForJurisdiction(envelope.jurisdiction);
```
rkondratowicz: why `info`?

PR #64, opened 2018-10-10
Create CCD models for holding supplementary evidence

lgonczar:

### Change description ###

Create CCD models for providing supplementary evidence. Extracted from https://github.com/hmcts/bulk-scan-orchestrator/pull/63, which is a bit too big for review.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/CcdDocument.java
```diff
@@ -0,0 +1,13 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class CcdDocument {
```
timja: this should probably live here:
https://github.com/hmcts/core-case-data-store-client/tree/master/src/main/java/uk/gov/hmcts/reform/ccd/client/model

PR #65, opened 2018-10-12
Fix exception handling + extra logging (needed to debug servicebus fa…

luigibk:
…iled connections).

https://tools.hmcts.net/jira/browse/RPE-736

### Change description ###
Extra debug logging mostly but also a fix for exception handling.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -34,3 +34,12 @@ idam:
     id: 'bsp'
     secret: '123456'
     redirect_uri: 'http://localhost/receiver'
+
+logging:
+  level:
+    uk.gov.hmcts.reform.bulkscan: WARN
+    org.springframework: WARN
+    com.microsoft.azure.servicebus: DEBUG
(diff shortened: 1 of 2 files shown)


src/main/resources/application.yaml
```diff
@@ -34,3 +34,12 @@ idam:
...
     redirect_uri: 'http://localhost/receiver'
+
+logging:
+  level:
+    uk.gov.hmcts.reform.bulkscan: WARN
+    org.springframework: WARN
+    com.microsoft.azure.servicebus: DEBUG
+    org.apache.qpid: DEBUG
+    org.apache.catalina: WARN
+    springfox.documentation: WARN
```
rkondratowicz: perhaps this could be configurable using env vars (I assume we don't want those on prod?)

luigibk: No, I was planning to remove it altogether, but you are right it could be made configurable.

PR #73, opened 2018-10-17
Log additional message processing information

lgonczar:
### Change description ###

Log additional message processing information

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/MessageProcessor.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/MessageProcessor.java
@@ -24,15 +24,20 @@ public class MessageProcessor {
 
     @Scheduled(fixedDelayString = "${queue.read-interval}")
     public void run() {
+        logger.info("Started processing queue messages.");
+
         IMessage msg = null;
+        int processedMessagesCount = 0;
+
         try {
             IMessageReceiver msgReceiver = messageReceiverFactory.create();
             while ((msg = msgReceiver.receive()) != null) {
-
                 envelopeProcessor.onMessageAsync(msg).get();
-
                 msgReceiver.complete(msg.getLockToken());
+                processedMessagesCount++;
             }
+
+            logger.info("Message processing complete. Processed {} messages", processedMessagesCount);
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
             logger.error("interrupted", e);


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/MessageProcessor.java
```diff
@@ -24,15 +24,20 @@
...
-
                 envelopeProcessor.onMessageAsync(msg).get();
-
                 msgReceiver.complete(msg.getLockToken());
+                processedMessagesCount++;
             }
+
+            logger.info("Message processing complete. Processed {} messages", processedMessagesCount);
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
```
KevinJCross: this also needs to be counted.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/MessageProcessor.java
```diff
@@ -24,15 +24,20 @@
...
+        int processedMessagesCount = 0;
+
         try {
             IMessageReceiver msgReceiver = messageReceiverFactory.create();
             while ((msg = msgReceiver.receive()) != null) {
-
                 envelopeProcessor.onMessageAsync(msg).get();
-
                 msgReceiver.complete(msg.getLockToken());
+                processedMessagesCount++;
```
KevinJCross: I think we should be using a metric here and report this back ... especially how many errors.

https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-metrics.html

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/MessageProcessor.java
lgonczar: What do you mean?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/MessageProcessor.java
lgonczar: I agree. It would be worth creating a story/task to add those.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/MessageProcessor.java
doncem: can be easily counted via appinsights taking into account the line below: `logger.error("interrupted", e)`. this will be registered with exceptions and can be filtered smth like `message == "interrupted"`

PR #83, opened 2018-10-23
RPE-670#2 introduce a processor/service for the callbacks

KevinJCross:
https://tools.hmcts.net/jira/browse/RPE-670

### Change description ###
Addition of a processor for processing the events

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackProcessor.kt
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackProcessor.kt
@@ -1,0 +1,20 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd
+
+import mu.KotlinLogging
+import org.springframework.stereotype.Service
+import uk.gov.hmcts.reform.ccd.client.model.CallbackRequest
+
+private val logger = KotlinLogging.logger {}
+
+@Service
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackProcessor.kt
```diff
@@ -0,0 +1,20 @@
...
+private val logger = KotlinLogging.logger {}
+
+@Service
+class CallbackProcessor {
+    /**
+     * returns
+     *  - a list of errors
+     *  - *or* an empty list  if it successfully attaches the supplementary evidence.
+     */
+    fun processEventType(type: String, callback: CallbackRequest): List<String> {
```
timja: probably over kill when we will only ever have 1 event type in orchestrator

KevinJCross: hmmm ... sure.
Are you suggesting to do event/type validation in the controller and specialise the "service" to a attachEventProcessor ?

timja: specialise the service to do what you need today, and not something too generic, as we aren't expecting to add more callbacks.

not blocking this PR though, but imo its too generic and does nothing atm

PR #87, opened 2018-10-23
Update case type

doncem:

[Create exception record for supplementary evidence type](https://tools.hmcts.net/jira/browse/RPE-656)

### Change description ###

`ExceptionRecord` is a different case type for ccd resource. Modifying all `EventPublisher` related classes to manually state which type is it during the autowire construct time

Work to be done: move case retriever to the publisher container so relevant calls can be made at the time of need

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EventPublisher.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EventPublisher.java
@@ -4,10 +4,5 @@ import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.model.Envel
 
 public interface EventPublisher {
 
-    // temporary adding custom case types until ccd api is updated
-    // at the moment BULKSCAN jurisdiction is able to use following types anyway
-    String CASE_TYPE_BULK_SCANNED = "Bulk_Scanned";
-    String CASE_TYPE_EXCEPTION_RECORD = "ExceptionRecord";
-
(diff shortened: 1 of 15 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EventPublisher.java
```diff
@@ -4,5 +4,8 @@
 
 public interface EventPublisher {
 
+    String BULK_SCANNED = "Bulk_Scanned";
+    String EXCEPTION_RECORD = "ExceptionRecord";
```
rkondratowicz: Are those names of events?
Would be nice to either rename or put in a class that indicates what those strings are IMHO.

rkondratowicz: 🛑 
Scrolling down in PR, turns out those are case type IDs apparently.
Indicate somehow what those are for sure.

doncem: I'll rename as I need to put a comment they are temporary anyway. Need ccd client to be updated where we can query those case types. @timja explained/reminded it's temporary hack

rkondratowicz: just moving them to `CaseTypeIds` seems cleanest to me

PR #89, opened 2018-10-23
[PREVIEW] Stop functional test from consuming message from the queue

lgonczar:
### Change description ###

Stop a functional test from consuming message from the queue

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/ReadFromTheQueueTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/ReadFromTheQueueTest.java
@@ -12,8 +12,9 @@ import org.junit.Before;
 import org.junit.Test;
 
 import java.util.UUID;
+import java.util.concurrent.TimeUnit;
 
-import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
 
 public class ReadFromTheQueueTest {
 
@@ -39,7 +40,7 @@ public class ReadFromTheQueueTest {
     }
 
     @Test
-    public void should_read_message_from_the_queue() throws Exception {
+    public void should_consume_message_from_the_queue() throws Exception {
         // given
         Message message = new Message();
         message.setMessageId(UUID.randomUUID().toString());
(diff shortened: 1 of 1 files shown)


src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/ReadFromTheQueueTest.java
```diff
@@ -39,19 +41,21 @@ public void setUp() throws Exception {
...
         message.setBody(SampleData.envelopeJson());
 
         // when
         writeClient.send(message);
-        Thread.sleep(readInterval + 5_000L); // wait for msg to be processed by the service.
 
         // then
-        assertThat(readClient.receive()).isNull();
 
-        // TODO: once implemented, check whether case in CCD has been created, envelope status updated etc...
```
lgonczar: This is handled by a different test

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/ReadFromTheQueueTest.java
```diff
@@ -39,19 +41,21 @@ public void setUp() throws Exception {
...
+        // wait for msg to be processed (and removed from the queue) by the service
+        await()
+            .atMost(readInterval + 5_000L, TimeUnit.MILLISECONDS)
+            .pollDelay(1, TimeUnit.SECONDS)
+            .until(() -> readClient.peek() == null);
```
timja: Does this assume that the queue will be empty?
Will it be broken by another test putting a message on the queue?

lgonczar: Yes, that's what's going to happen, as long as the two tests are using the same queue. Quite unlikely, as the orchestrator should process all messages in one go. What this PR is doing is:
* it replaces `.receive()` with `.peek()`, so that the test doesn't affect the queue
* reduces the waiting

Whether we need that test at all is another question.

timja: right so doesn't test anything on PRs as anything could be consuming the queue currently...
May as well get it in as its an improvement.

PR #90, opened 2018-10-23
RPE-670#5 Call the start attach event ccd api.

KevinJCross:
https://tools.hmcts.net/jira/browse/RPE-670

### Change description ###
Start the call to the Start attach event.
The TBD eventId still needs decision so needs some discussion :D

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackProcessor.kt
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackProcessor.kt
@@ -1,7 +1,9 @@
 package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd
 
+import feign.FeignException
 import mu.KotlinLogging
 import org.springframework.stereotype.Service
+import uk.gov.hmcts.reform.ccd.client.CoreCaseDataApi
 import uk.gov.hmcts.reform.ccd.client.model.CallbackRequest
 import uk.gov.hmcts.reform.ccd.client.model.CallbackTypes.ABOUT_TO_SUBMIT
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackProcessor.kt
```diff
@@ -48,6 +53,33 @@ class CallbackProcessor {
...
+                caseDetails.data["attachToCaseReference"] as String?,
+                "TBD"
+            )
+            return listOf()
+        } catch (e: FeignException) {
+            logger.error(e) {}
+            return listOf("Internal Error: response ${e.status()} submitting event")
+        } catch (e: Exception) {
+            logger.error(e) {}
+            return listOf("Internal Error: ${e::class.simpleName}:${e.message}")
```
timja: probably should just be Internal server error, if you're feeling brave finding out the request id for app insights and returning that here too?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackProcessor.kt
```diff
@@ -48,6 +53,33 @@ class CallbackProcessor {
...
+        try {
+            val authenticator = authenticatorFactory.createForJurisdiction(caseDetails.jurisdiction)
+            ccdApi.startEventForCaseWorker(
+                authenticator.userToken,
+                authenticator.serviceToken,
+                authenticator.userDetails.id,
+                caseDetails.jurisdiction,
+                caseDetails.caseTypeId,
+                caseDetails.data["attachToCaseReference"] as String?,
+                "TBD"
```
timja: so the uh elephant here, is there no appropriate event already?

PR #94, opened 2018-10-24
Non static init

luigibk:
https://tools.hmcts.net/jira/browse/RPE-781

### Change description ###
Fix Bulk Scan Orchestrator's NoClassDefFoundError test problem

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/SampleData.java
+++ b/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/SampleData.java
@@ -100,6 +100,10 @@ public class SampleData {
 
     }
 
+    public static byte[] exampleJsonAsBytes() {
+        return fileContentAsBytes(EXAMPLE_JSON_FILE);
+    }
+
     public static byte[] fileContentAsBytes(String file) {
         try {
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/SampleData.java
```diff
@@ -104,12 +100,10 @@
 
     }
 
-    public static String fromFile(String file) {
+    public static byte[] fileContentAsBytes(String file) {
```
timja: i wonder if this should just be `exampleJsonAsBytes` sample data shouldn't be as generic as this imo, this is more of a util

luigibk: I agree that it is more of a util, but it's certainly more generic than `exampleJsonAsBytes`

timja: is it though? you only needed to make one change in tests

luigibk: I am in favour of more generic methods generally. Have a look now

PR #96, opened 2018-10-24
Update case retriever dependency

doncem:
### Change description ###

Move `CaseRetriever` to another component of dependency chain instead of `EnvelopeProcessor` leaving the processing steps dead simple:

- parse
- choose publisher
- publish

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/AbstractEventPublisher.java
```diff
@@ -58,7 +58,7 @@
  */
 abstract class AbstractEventPublisher implements EventPublisher {
 
-    private final String caseTypeId;
+    private final CaseTypeId caseTypeId;
```
timja: but casetype is a string? there is only one fixed case type that we should be aware of which is exceptionrecord
all other case types will be dynamic as far as I can tell we should not be using an enum here

doncem:
Overridden by #104

PR #104, opened 2018-10-25
RPE-657 Add case type to processing layer

timja:

### Change description ###
Case type for supplementary evidence is different for each jurisdiction,
we shouldn't be hardcoding it

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
```diff
@@ -53,12 +56,12 @@ private void process(IMessage message) {
         Envelope envelope = parse(message.getBody());
         CaseDetails theCase = Strings.isNullOrEmpty(envelope.caseRef)
             ? null
-            : caseRetriever.retrieve(envelope.jurisdiction, CaseTypeId.BULK_SCANNED, envelope.caseRef);
-
+            : caseRetriever.retrieve(envelope.jurisdiction, TEMP_CASE_TYPE, envelope.caseRef);
```
KevinJCross: why not just use
 ```java
caseTypeId  = optional.ofNullable(envelope.caseRef)
.map( ref ->caseRetriever.retrieve(envelope.jurisdiction, TEMP_CASE_TYPE, ref )
.map( aCase -> getCaseTypeId(aCase))
```

timja: I've not changing that part of the code,
I raised this on Don's PR and he said he preferred it this way and it was simpler.

As its a matter of preference and its been discussed before I'm happy to leave it this way.

PR #113, opened 2018-10-26
RPE-670#5 case null check

KevinJCross:
**Before creating a pull request make sure that:**
RPE-670

### Change description ###
adding null check for the case data
Also combining the validations into one class to extract at a later date.


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackProcessor.java
```diff
@@ -0,0 +1,9 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd;
+
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+
+import java.util.List;
+
+public interface CallbackProcessor {
```
timja: why do you need this interface?

KevinJCross: Well not really ... its one of those springy things about object contracts.
Could remove it if ya want

timja: would prefer, seems unneeded to me

KevinJCross: It seems I got then to  branches mixed in this PR ... just give me a sec

PR #114, opened 2018-10-26
RPE-670#6 refactor cleanup

KevinJCross:
RPE-670

### Change description ###
Minor refactor of test and pulling out the interface with static methods


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -0,0 +1,41 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+
+import static io.vavr.control.Validation.invalid;
+import static io.vavr.control.Validation.valid;
+import static java.lang.String.format;
+import static uk.gov.hmcts.reform.ccd.client.model.CallbackTypes.ABOUT_TO_SUBMIT;
+
+interface CallbackValidations {
```
rkondratowicz: What does PMD think about this?
We usually put static helper methods in a final class.

KevinJCross: no failures on PMD.
I find interfaces better for using to write plain functions.
Its better for these type of helpers since you don't have to cover the constructor in code coverage and everything is automatically  public static final.

PR #135, opened 2018-11-05
Feature: delegate publisher

doncem:
### Change description ###

The goal is to **not** make API request to retrieve case when it is definitely not needed to.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/CreateExceptionRecord.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/CreateExceptionRecord.java
@@ -9,6 +9,8 @@ import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.model.Envel
 @Component
 class CreateExceptionRecord extends AbstractEventPublisher {
 
+    static final String CASE_TYPE = "ExceptionRecord";
+
     private final ModelMapper<? extends CaseData> mapper;
 
     CreateExceptionRecord(ExceptionRecordMapper mapper) {
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/CreateExceptionRecord.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/CreateExceptionRecord.java
@@ -17,6 +17,11 @@ class CreateExceptionRecord extends AbstractEventPublisher {
         this.mapper = mapper;
     }
 
+    @Override
+    public void publish(Envelope envelope) {
+        publish(envelope, CASE_TYPE);
+    }
+
     /**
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/CreateExceptionRecord.java
```diff
@@ -9,12 +9,19 @@
 @Component
 class CreateExceptionRecord extends AbstractEventPublisher {
 
+    static final String CASE_TYPE = "ExceptionRecord";
```
nitinprabhu: can this be private ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EventPublisherContainer.java
```diff
@@ -27,15 +30,24 @@
...
+                return caseDetails == null ? exceptionRecordCreator : new DelegatePublisher(
+                    attachDocsPublisher,
+                    caseDetails
+                );
             case EXCEPTION:
                 return exceptionRecordCreator;
             case NEW_APPLICATION:
             default:
-                return null;
+                return new DelegatePublisher(null, null);
```
nitinprabhu: Can we not throw an exception here ? Ideally we are validating the classification in processor anyways so we don't expect this control to be executed.

doncem: We are. This way is backwards compatible too. We do nothing if we received something unrecognisable. Exception throw will be something new entirely and will bring message back to deadletterqueue. If that's expected - happy to invent one

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/CreateExceptionRecord.java
doncem: It is used in DelegatePublisher and then in functional tests from there. Most of classes and it's elements are package private because it's not needed to be exposed. Everything orchestrated in orchestrator :D

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EventPublisherContainer.java
nitinprabhu: We should at least log in app insights and configure alerts so that bsp team knows whenever this happens in production. Please create a ticket in backlog for alerts config and as of now just log it in app insights. Makes sense ?

doncem: It is logged. Check out DelegatePublisher ;). Perhaps needs better logging as you said - in separate ticket. A matter of review

PR #147, opened 2018-11-07
RPE-670#11 Case reference input cleanup and error message improvements

KevinJCross:
RPE-670

### Change description ###
- Added non number character removal from user input
- Improved the errors to be more human readable
- added comprehensive unit test

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
@@ -34,7 +34,7 @@ final class CallbackValidations {
     }
 
     @NotNull
-    static <T> Validation<String, T> internalError(String error, T arg1) {
+    private static <T> Validation<String, T> internalError(String error, T arg1) {
         log.error("{}:{}", error, arg1);
         String formatString = "Internal Error: " + error;
         return invalid(format(formatString, arg1));
@@ -51,13 +51,17 @@ final class CallbackValidations {
 
     @NotNull
     static Validation<String, String> hasCaseReference(CaseDetails theCase) {
-        Object caseReference = null;
+        Object caseObj = null;
+        String caseReference;
         return theCase != null
             && theCase.getData() != null
-            && (caseReference = theCase.getData().get(ATTACH_TO_CASE_REFERENCE)) != null
-            && (caseReference instanceof String)
-            ? valid((String) caseReference)
-            : internalError("no case reference found: %s", String.valueOf(caseReference));
+            && (caseObj = theCase.getData().get(ATTACH_TO_CASE_REFERENCE)) != null
+            && (caseObj instanceof String)
+            && (!(caseReference = ((String) caseObj).replaceAll("[^0-9]", "")).isEmpty())
+            ? valid(caseReference)
+            : invalid((caseObj != null)
+            ? format("Invalid case reference: '%s'", String.valueOf(caseObj))
+            : "No case reference supplied");
     }
 
     @NotNull
--- a/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidationsTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidationsTest.java
@@ -1,0 +1,39 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd;
+
+import io.vavr.control.Validation;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+
+import static org.assertj.core.api.SoftAssertions.assertSoftly;
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.TestCaseBuilder.caseWithReference;
+
+class CallbackValidationsTest {
+
+    private static Object[][] caseReferenceTestParams() {
+        return new Object[][]{
+            {caseWithReference("£1234234393"), true, "1234234393"}
+        };
+    }
+
+    @ParameterizedTest(name = "valid:{1} value:{2}")
+    @MethodSource("caseReferenceTestParams")
+    @DisplayName("Should accept and remove non 0-9 chars in the case reference")
+    void caseReferenceTest(CaseDetails input, boolean valid, String realValue) {
+        Validation<String, String> validation = CallbackValidations.hasCaseReference(input);
+        if (valid) {
+            assertSoftly(softly -> {
+                softly.assertThat(validation.isValid()).isTrue();
+                softly.assertThat(validation.get()).isEqualTo(realValue);
+            });
+        } else {
+            assertSoftly(softly -> {
+                softly.assertThat(validation.isValid()).isFalse();
+                softly.assertThat(validation.getError()).isEqualTo("someValue");
+            });
+        }
+    }
+
+
+}
--- a/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidationsTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidationsTest.java
@@ -8,12 +8,21 @@ import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
 
 import static org.assertj.core.api.SoftAssertions.assertSoftly;
 import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.TestCaseBuilder.caseWithReference;
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.TestCaseBuilder.createCaseWith;
 
 class CallbackValidationsTest {
 
     private static Object[][] caseReferenceTestParams() {
         return new Object[][]{
-            {caseWithReference("£1234234393"), true, "1234234393"}
+            {caseWithReference("£1234234393"), true, "1234234393"},
+            {caseWithReference("1234-234-393"), true, "1234234393"},
+            {caseWithReference("1234 234 393"), true, "1234234393"},
+            {caseWithReference("  AH1234 234 393"), true, "1234234393"},
+            {caseWithReference("#"), false, "Invalid case reference: '#'"},
+            {caseWithReference(""), false, "Invalid case reference: ''"},
+            {null, false, "No case reference supplied"},
+            {createCaseWith(b -> b.data(null)), false, "No case reference supplied"},
+            {caseWithReference(null), false, "No case reference supplied"},
         };
     }
 
@@ -30,7 +39,7 @@ class CallbackValidationsTest {
         } else {
             assertSoftly(softly -> {
                 softly.assertThat(validation.isValid()).isFalse();
-                softly.assertThat(validation.getError()).isEqualTo("someValue");
+                softly.assertThat(validation.getError()).isEqualTo(realValue);
             });
         }
     }
--- a/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/TestCaseBuilder.java
+++ b/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/TestCaseBuilder.java
@@ -1,0 +1,18 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd;
+
+import com.google.common.collect.ImmutableMap;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+
+import java.util.function.Function;
+
+public class TestCaseBuilder {
+    private TestCaseBuilder(){
+    }
+    private static CaseDetails createCaseWithRef(Function<CaseDetails.CaseDetailsBuilder, CaseDetails.CaseDetailsBuilder> builder){
+        return builder.apply(CaseDetails.builder()).build();
+    }
+
+    static CaseDetails caseWithReference(String caseReference) {
+        return createCaseWithRef(b -> b.data(ImmutableMap.of("attachToCaseReference", caseReference)));
+    }
+}
--- a/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/TestCaseBuilder.java
+++ b/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/TestCaseBuilder.java
@@ -1,18 +1,22 @@
 package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd;
 
-import com.google.common.collect.ImmutableMap;
 import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.function.Function;
 
-public class TestCaseBuilder {
+class TestCaseBuilder {
     private TestCaseBuilder(){
     }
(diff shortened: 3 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -51,13 +51,17 @@ private CallbackValidations() {
...
+        String caseReference;
         return theCase != null
             && theCase.getData() != null
-            && (caseReference = theCase.getData().get(ATTACH_TO_CASE_REFERENCE)) != null
-            && (caseReference instanceof String)
-            ? valid((String) caseReference)
-            : internalError("no case reference found: %s", String.valueOf(caseReference));
+            && (caseObj = theCase.getData().get(ATTACH_TO_CASE_REFERENCE)) != null
+            && (caseObj instanceof String)
+            && (!(caseReference = ((String) caseObj).replaceAll("[^0-9]", "")).isEmpty())
```
lgonczar: Is it ok that we accept a mix of digits and other characters, while it should always be just digits?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -51,13 +51,17 @@ private CallbackValidations() {
 
     @NotNull
     static Validation<String, String> hasCaseReference(CaseDetails theCase) {
-        Object caseReference = null;
+        Object caseObj = null;
```
lgonczar: I'd refactor this method to something that will perhaps be less optimal in terms of lines of code or the number of statements, but easier to read.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
KevinJCross: well from the testing we can enter what is from the web page ..
I imagine that the user will just copy and paste from the web page the reference number
i.e.
`#102932-129831-198171  `
So this makes it up to us to make it into a parseable long

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
nitinprabhu: I agree with @lgonczar this method is difficult to read

KevinJCross: yeah its just a mass of if statments ... tbh Im not sure its going to any easier to follow if I break it down.
i.e.
```java
        Object caseObj = null;
        String caseReference;
        if( theCase != null) {
            if (theCase.getData() != null) {
                caseObj = theCase.getData().get(ATTACH_TO_CASE_REFERENCE);
                if (caseObj != null) {
                    if ( caseObj instanceof String ) {
                        caseReference = ((String) caseObj).replaceAll("[^0-9]", "");
                        if (!caseReference.isEmpty()) {
                            return valid(caseReference);
                        }
                    }
                }
            }
        }
        if ( caseObj != null ){
            return invalid(format("Invalid case reference: '%s'", String.valueOf(caseObj)));
        }else{
            return invalid( "No case reference supplied");
        }
```

KevinJCross: I personally think the above is actually easier to follow ??

KevinJCross: So the Optional version of this is:
```java

    @NotNull
    static Validation<String, String> hasCaseReference(CaseDetails theCase) {
        return Optional.ofNullable(theCase)
            .map(CaseDetails::getData)
            .map(data -> data.get(ATTACH_TO_CASE_REFERENCE))
            .flatMap(CallbackValidations::getCaseReference)
            .orElseGet(() -> invalid("No case reference supplied"));
    }

    private static Optional<Validation<String, String>> getCaseReference(Object reference) {
        Optional<Validation<String, String>> refOpt = validRef(reference).map(Validation::valid);
        return refOpt.isPresent() ? refOpt : Optional.of(invalid(format("Invalid case reference: '%s'",reference)));
    }
```

KevinJCross: Not much better mostly because optional does not support orElseMap hence the `refOpt.isPresent() ? `
Ill quickly try *RxSingle* which is better for this.

KevinJCross: actually we only have v1 Rx included so I used vavr Option instead

```java
 static Validation<String, String> hasCaseReference(CaseDetails theCase) {
        return Option.of(theCase)
            .flatMap(c -> Option.of(c.getData()))
            .flatMap(data -> Option.of(data.get(ATTACH_TO_CASE_REFERENCE)))
            .flatMap(item -> Option.of(item)
                .filter(ref -> ref instanceof String)
                .map(ref -> ((String) ref))
                .map(ref -> ref.replaceAll("[^0-9]", ""))
                .filter(StringUtils::isNotEmpty)
                .map(Validation::<String, String>valid)
                .orElse(() -> Option.of(invalid(format("Invalid case reference: '%s'", item))))
            )
            .getOrElse( ()->invalid("No case reference supplied"));
    }
```
I think I prefer this version maybe extract the flatmap though.

KevinJCross: The reason is that its input data ... every layer can be null so we have to check.
Otherwise the code is not *solid* and will break.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -51,13 +53,14 @@ private CallbackValidations() {
...
-        return theCase != null
-            && theCase.getData() != null
-            && (caseReference = theCase.getData().get(ATTACH_TO_CASE_REFERENCE)) != null
-            && (caseReference instanceof String)
-            ? valid((String) caseReference)
-            : internalError("no case reference found: %s", String.valueOf(caseReference));
+        return caseRefValidator.validate(theCase);
+    }
+
+    private static Optional<String> validRef(Object reference) {
```
doncem: Given the `caseRefValidator` presence, is this method being used?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -51,13 +53,14 @@ private CallbackValidations() {
...
+        return Optional.of(reference)
+            .filter(ref -> ref instanceof String)
+            .map(ref -> ((String) ref).replaceAll("[^0-9]", ""))
+            .filter(ref -> !ref.isEmpty());
```
doncem: Although this implementation is acceptable and safe normally in java process I see something as follows:
```java
some method definition(String reference) {
    Assert.notEmpty(reference, "message");

    String ref = com.google.common.base.Strings.emptyOrNull(reference.replaceAll("[^0-9]", ""));

    return Optional.ofNull(ref);
}
```

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CaseReferenceValidator.java
```diff
@@ -0,0 +1,31 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd;
+
+import io.vavr.control.Option;
```
doncem: Hmm. Just now paying attention to `vavr` usage over here in this repository. As I scroll in this file I see common usage of `java.util.Optional`. Is there a benefit of using extra library just to have some sort of feature going on? I do get the `cats` for `scala` just questioning `flatmap` with friends usage over here

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CaseReferenceValidator.java
```diff
@@ -0,0 +1,31 @@
...
+    private Option<Validation<String, String>> validateAndConvert(Object item) {
+        return Option.of(item)
+            .filter(ref -> ref instanceof String)
+            .map(ref -> ((String) ref))
+            .map(ref -> ref.replaceAll("[^0-9]", ""))
+            .filter(StringUtils::isNotEmpty)
+            .map(Validation::<String, String>valid)
+            .orElse(() -> Option.of(invalid(format("Invalid case reference: '%s'", item))));
+    }
+}
```
doncem: If everything may be an option, can save reference stack trace by
```java
private Option<T> getOption(T t) {
    return Option.of(t);
}
```

Thought it might introduce double mapping. I'd rather keep flatmaps. Once again, more of a functional approach/flavour

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CaseReferenceValidator.java
KevinJCross: if we are using the Optional from java 9 then all would be good the problem is that Optional in java 8 does not have  Optional<T> orElse ( Optional<T> ) 
this means that the solution mentioned in the comments above can be used.
I guess I can back port this if its required.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
KevinJCross: sonar picked this up to 
```
src/.../orchestrator/services/ccd/CallbackValidations.java
Remove this unused private "validRef" method.
yesterday
L59
```

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
KevinJCross: Moved this back to using Optional since vavr might be to big a step and with java 9+ would look a little simpler.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
KevinJCross: and there is not type checking either :/ ... since this is a union json element type it is quite possible to get a wrong object turning up in here at some point

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CaseReferenceValidator.java
KevinJCross: Ive moved it back to Optional since Vavr Option might not go down well here.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CaseReferenceValidator.java
KevinJCross: Ive moved it back to optional since optional auto flatmap null to empty ( for better or worse)

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CaseReferenceValidator.java
```diff
@@ -0,0 +1,38 @@
...
+            .filter(aRef -> aRef instanceof String)
+            .map(aRef -> ((String) aRef).replaceAll("[^0-9]", ""))
+            .map(Strings::emptyToNull)
+            .map(Validation::valid);
+        //This is done because java 8 does not have "Optional<T> orElse( Optional<T>)"
+        if (valid.isPresent()) {
+            return valid;
+        } else {
+            return Optional.of(invalid(format("Invalid case reference: '%s'", caseRef)));
+        }
```
doncem: do we need to return optional at all here? looks like we could just `.orElse` and get rid of another flatmap one method above. plus second `.orElseGet` over there is not needed either as this method is always returning non empty optional value

KevinJCross: nope ... thats the problem with java 8 optional.
This is pretty  much how I have to do it.
give it a go if you like but this is the exact reason orElse() was added to optional in java 9

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CaseReferenceValidator.java
```diff
@@ -0,0 +1,38 @@
...
+
+import static io.vavr.control.Validation.invalid;
+import static java.lang.String.format;
+
+class CaseReferenceValidator {
+    private static final String ATTACH_TO_CASE_REFERENCE = "attachToCaseReference";
+
+    @Nonnull
+    Validation<String, String> validate(CaseDetails theCase) {
+        return Optional.ofNullable(theCase)
```
lgonczar: As we're only making specific checks here, we could as well do sth like this:
```
String caseRef = getCaseRef(theCase); // alternatively, this could be optional, and then you could map, etc.

if (caseRef == null) {
    return invalid("No case reference supplied");
} else {
    return validateCaseRef(caseRef)
}
```

My main point is that we can extract some of the complexity into a separate method, which also names that bit.

Not very strong about it, as it's not crucial what approach we end up with. It's a bottom level thing.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CaseReferenceValidator.java
doncem: I get what you mean. In short, question is "do we need `Optional<?>` as a returned value here?"

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CaseReferenceValidator.java
```diff
@@ -0,0 +1,38 @@
...
+            .map(CaseDetails::getData)
+            .map(data -> data.get(ATTACH_TO_CASE_REFERENCE))
+            .flatMap(this::validateCaseRef)
+            .orElseGet(() -> invalid("No case reference supplied"));
+    }
+
+    private Optional<Validation<String, String>> validateCaseRef(Object caseRef) {
+        Optional<Validation<String, String>> valid = Optional.of(caseRef)
```
lgonczar: We only check if caseRef is a String having a digit in it. I think if we used conditions, it would be the clearest. Again, a suggestion only.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CaseReferenceValidator.java
nitinprabhu: looks much cleaner and easier to read what @lgonczar has suggested

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CaseReferenceValidator.java
KevinJCross: The problem is that every part can be null or the wrong type so writing solid code means you need to check them.
Hence the functional chain.

PR #150, opened 2018-11-07
RPE-670#12 scanned document fetching and validation

KevinJCross:
RPE-670

### Change description ###
- check that the exception record has a list of documents
- validate that there is at least one document
- retrieve record for Processor


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackProcessor.java
```diff
@@ -69,10 +72,11 @@ public CallbackProcessor(CcdApi ccdApi) {
 
     private void attachCase(String exceptionRecordJurisdiction,
                             String caseRef,
-                            CaseDetails exceptionRecord) {
+                            CaseDetails exceptionRecord,
+                            List<Map<String, Object>> exceptionDocuments) {
         CaseDetails theCase = ccdApi.getCase(caseRef, exceptionRecordJurisdiction);
         StartEventResponse event = ccdApi.startAttachScannedDocs(theCase);
-        Map<String, Object> data = insertNewScannedDocument(exceptionRecord.getData(), theCase.getData());
+        Map<String, Object> data = insertNewScannedDocument(exceptionDocuments, theCase.getData());
```
nitinprabhu: exception documents should be list of ScannedDocument(with all attributes of document)  objects. Why we are using a map here ?

KevinJCross: this is how the data is presented to us json as a map.
The spikes did not reveal this and contract has never been evident or actually documented.
Even though I did ask for them.

Hence I've used what was given via trial and error ( actually trying this).
The benefit of this approach is that we are using the shape of the data rather than forcing it to a structure, which also allows for data to change and the information will always move correctly.

KevinJCross: Sure so who defines the ScannedDocument and the ScannedRecord structures and where are they?
I have not been given this bit of information before  and how can we add a test to check when these change ?

nitinprabhu: Structure is defined in ccd excel spreadsheet ComplexType tab.

```

ScannedDocument | # Record Meta data
-- | --
ScannedDocument | documentType
ScannedDocument | documentLink
ScannedDocument | documentControlNumber
ScannedDocument | filename
ScannedDocument | docScanDate
```

nitinprabhu: If you update the spreadsheet test should break as the structure is changed and hence it is worth defining a ScannedDocument POJO with fields corresponding in https://github.com/hmcts/bulk-scan-shared-infrastructure/blob/master/docker/ccd-definition-import/data/CCD_Definition_Exception_Record.template.xlsx

PR #151, opened 2018-11-08
RPE-670#13 detect and error on duplicate document number

KevinJCross:
RPE-670

### Change description ###
* detect a duplicate document number attachment
* error back to user with correct information.
* avoid changing exception record state or submitting anything.


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackProcessor.java
```diff
@@ -96,13 +112,24 @@ private void attachCase(String exceptionRecordJurisdiction,
...
-    @SuppressWarnings({"unchecked", "squid:S1135"})
-    //TODO WIP
     private Map<String, Object> insertNewScannedDocument(List<Map<String, Object>> exceptionDocuments,
-                                                         Map<String, Object> caseData) {
-        //TODO check that document Id is unique and not duplicate in caseData
-        List<Object> caseList = (List<Object>) caseData.get(SCANNED_DOCUMENTS);
-        caseList.addAll(exceptionDocuments);
-        return ImmutableMap.of(SCANNED_DOCUMENTS, caseList);
+                                                         List<Map<String, Object>> existingDocuments) {
+        //TODO assert that there is one document in the exception record.
```
timja: is 1 document correct?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackProcessor.java
```diff
@@ -96,13 +112,24 @@ private void attachCase(String exceptionRecordJurisdiction,
...
+        existingDocuments.add(exceptionDocuments.get(0));
+        return ImmutableMap.of(SCANNED_DOCUMENTS, existingDocuments);
+    }
+
+    private void checkForDuplicateAttachment(List<Map<String, Object>> exceptionDocuments,
+                                             String caseRef,
+                                             List<Map<String, Object>> existingDocuments) {
+        String exceptionRecordId = (String) exceptionDocuments.get(0).get("documentNumber");
```
timja: always only 1?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackProcessor.java
nitinprabhu: it's not correct exception record can have multiple documents

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackProcessor.java
```diff
@@ -75,16 +78,29 @@ private void attachCase(String exceptionRecordJurisdiction,
...
-            insertNewScannedDocument(exceptionDocuments, theCase.getData()),
+            insertNewScannedDocument(exceptionDocuments, existingDocuments),
             event.getEventId(),
             event.getToken()
         );
     }
 
+    @SuppressWarnings("unchecked")
+    private List<Map<String, Object>> getDocuments(CaseDetails theCase) {
+        return (List<Map<String, Object>>) theCase.getData().get(SCANNED_DOCUMENTS);
```
nitinprabhu: scanned documents should be list of scanneddoc objects and not a map.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackProcessor.java
KevinJCross: Ok sure ... Il change that. Its was an assumption. 
That will make the duplicate document check a bit more complex

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackProcessor.java
KevinJCross: yup as mentioned above...
I've only seen one so and there are no real requirements written down.
Thats why we have a PR and the //TODO check assumption :smile:

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackProcessor.java
KevinJCross: So the correct behaviour on the detection of a duplicate document 1 or more via document number in the case of multiple documents should be ?
1 - error and do not progress the attachment
2 - Leave out the duplicates ( what about newer or changed document) ?
3 - put a warning about duplicate documents and do 2 ?

KevinJCross: Note CCD will fail if there is a duplicate DCN and do so quietly to the user.

timja: error I guess? is it likely to happen? i.e. is there a legitimate way that it's possible to happen?

KevinJCross: if you attach the same exception record twice

KevinJCross: The way i came across it is attaching a duplicate exception record ( same message sent twice) and trying to attach it to the same case.

KevinJCross: Ill add this to a new PR#14 to deal with multiple received documents. Which I'll complete once this is merged.

KevinJCross: Ill update the //TODO comment

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackProcessor.java
KevinJCross: updated the TODO and will deal with more than one doc in the next PR

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackProcessor.java
KevinJCross: At the moment they are maps and i'm not sure we can/should re-deserialize this, write to json string and deserialize into an object, due to the changeable nature of this piece of json.

KevinJCross: As a cleanup activity I want to wrap it into a Document container object for ease of use but this does not currently affect the functionality.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackProcessor.java
KevinJCross: yes that is exactly why we need to do the duplicate check.
Hence this PR

PR #154, opened 2018-11-12
RPE-657 Supplementary Evidence bug fix

aliveni:

### Change description ###
Updated Attach Scanned Documents Event handler to  append 
the scanned documents to the existing evidence documents in the ccd case.
Added functional test to check if the evidence documents in the ccd case 
are not removed by the Supplementary Evidence event handler.

```
[ ] Yes
[ ] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/AttachDocsToSupplementaryEvidence.java
```diff
@@ -20,6 +32,31 @@ CaseData mapEnvelopeToCaseDataObject(Envelope envelope) {
...
     }
 
+    @Override
+    public CaseDataContent buildCaseDataContent(StartEventResponse eventResponse, Envelope envelope) {
+
+        List<ScannedDocument> ccdScannedDocuments = getScannedDocumentsForSupplementaryEvidence(
+            eventResponse.getCaseDetails()
+        );
+
+        List<ScannedDocument> scannedDocuments = getScannedDocuments(envelope);
```
timja: What do people think, should this be a set to prevent any duplicates when we add all the documents down a line?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/helper/ScannedDocumentsHelper.java
```diff
@@ -0,0 +1,69 @@
...
+        objectMapper.enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS);
+    }
+
+    private ScannedDocumentsHelper() {
+        // utility class
+    }
+
+    @SuppressWarnings("unchecked")
+    public static List<ScannedDocument> getScannedDocumentsForExceptionRecord(CaseDetails caseDetails) {
+        List<Map<String, Object>> data = (List<Map<String, Object>>) caseDetails.getData().get("scanRecords");
```
timja: @nitinprabhu do we need different property names in exception records and supplementary evidence (`scanRecords` and `scannedDocuments`) or could we pick one name and go with that.

@aliveni don't do that in this PR though, unrelated change, just would be good to fix it

nitinprabhu: @timja I don't think so we need different names. Yes can be clubbed into one 👍

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/AttachDocsToSupplementaryEvidence.java
nitinprabhu: I think we should not filter any documents even if they are duplicates. Caseworker will need to remove it through UI once they are satisfied that document is duplicate or not required for case progression.

PR #158, opened 2018-11-14
BSP-246#1 removing all suppressed warnings

KevinJCross:
BSP-246

### Change description ###
- removing all sonar warning suppressions from CallbackProccessor 
- renamed CallbackProccessor to AttachCaseCallbackService
- removed unused validations.


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CcdCallbackController.java
```diff
@@ -1,42 +1,52 @@
...
         produces = APPLICATION_JSON_VALUE
     )
-    public ResponseEntity<CallbackResponse> handleCallback(
-        @PathVariable("type") String type,
-        @RequestBody CallbackRequest callback) {
+    public ResponseEntity<CallbackResponse> handleCallback(@RequestBody CallbackRequest callback) {
+        if (callback != null && callback.getCaseDetails() != null) {
+            return respondWith(attachCaseCallbackService.process(callback.getCaseDetails()));
+        } else {
+            return respondWith(ImmutableList.of("Internal Error: callback or case details were empty"));
```
rkondratowicz: More like client error.
Also, it would be nice to add a comment explaining why we return `200 OK` here despite errors.

KevinJCross: hmmm ...  Ill refactor

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CcdCallbackController.java
```diff
@@ -1,42 +1,52 @@
...
-        path = "/callback/{type}",
+        path = "/callback/attach_case",
         consumes = APPLICATION_JSON_VALUE,
         produces = APPLICATION_JSON_VALUE
     )
-    public ResponseEntity<CallbackResponse> handleCallback(
-        @PathVariable("type") String type,
-        @RequestBody CallbackRequest callback) {
+    public ResponseEntity<CallbackResponse> handleCallback(@RequestBody CallbackRequest callback) {
+        if (callback != null && callback.getCaseDetails() != null) {
```
doncem: This can be avoided by using built-in validation feature. Need to provide `@NotNull` and `@Valid` (maybe more) interfaces.

case details _nullability_ can be solved in service automatically too and necessary exception caught and translated in spring advice: proper and super clean way of dealing with this situation plus kind of minimising branching

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CcdCallbackController.java
doncem: not agreeing nor disagreeing. but i'm in favour of @rkondratowicz. the way response is built it would be better to respond OK but with errors. it is client error - somebody called with empty sections in body therefore making a 400. but again, the structure of response suggests an alternative

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -17,66 +17,51 @@
...
         }
     }
 
-    //FOR the to do warnings
-    @SuppressWarnings("squid:S1135")
-    private void attachCase(String exceptionRecordJurisdiction,
-                            String caseRef,
-                            CaseDetails exceptionRecord,
-                            List<Map<String, Object>> exceptionDocuments) {
+    private void doAttachCase(String exceptionRecordJurisdiction,
```
doncem: `attachCase` was good enough. but nvm ;)

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CcdCallbackController.java
KevinJCross: sure ... the hibernate validations would help here :D
but that's assuming you have control of the object definition being serialised.
Its in another repo so ... its debatable if we should add them there since it could break others.\
Maybe adding our own definition here might help.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
KevinJCross: it's really to separate the error handling and the actual code. I personally find try catch statements clumsy and annoying and would rather pass back errors in a monad but ... well ... thats a little to FP for java.
If your interested have a look at the vavr Try.of().recover() stuff. 😄

PR #159, opened 2018-11-14
BSP-246#2 return a new list if case has none.

KevinJCross:
BSP-246

### Change description ###
As last TODO cleanup is to return a new list if the retrieved case has none.


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/Documents.java
```diff
@@ -26,7 +27,10 @@ private Documents() {
 
     private static Set<String> findDuplicates(List<Map<String, Object>> exceptionDocuments,
                                               List<Map<String, Object>> existingDocuments) {
-        return Sets.intersection(getDocumentIdSet(existingDocuments), getDocumentIdSet(exceptionDocuments));
+        return Sets.intersection(
+            getDocumentIdSet(existingDocuments),
+            getDocumentIdSet(exceptionDocuments)
+        );
```
rkondratowicz: One way to ensure 'small PRs' mentioned during every retrospective is to refrain from completely unrelated changes.
Also git log/blame will make more sense...

PR #160, opened 2018-11-14
RPE-260 Fixed functional test for attach exception record to the case…

aliveni:
https://tools.hmcts.net/jira/browse/BPS-260

### Change description ###
Fixed functional test for attaching the exception record to the existing case.
Added another test to attach exception record to the existing case with evidence

```
[ ] Yes
[ x] No
```


src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -95,34 +93,61 @@ public void setup() throws InterruptedException, ServiceBusException, JSONExcept
...
+        await("Exception record is attached to the case")
+            .atMost(60, TimeUnit.SECONDS)
+            .pollDelay(2, TimeUnit.SECONDS)
+            .until(this::isExceptionRecordAttachedToTheCase);
+
+        verifyExistingCaseIsUpdatedWithExceptionRecordData();
+    }
+
+    public void callCallbackUrlToAttachExceptionToCase() {
+        Map<String, Object> caseData = exceptionRecord.getData();
```
rkondratowicz: avoid state in the test and pass parameters for clarity

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -95,34 +93,61 @@ public void setup() throws InterruptedException, ServiceBusException, JSONExcept
...
+        caseData.put("attachToCaseReference", String.valueOf(caseDetails.getId()));
```
rkondratowicz: and here.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -95,34 +93,61 @@ public void setup() throws InterruptedException, ServiceBusException, JSONExcept
...
+    }
+
+    @Test
+    public void should_atatch_exception_record_to_the_existing_case() throws JSONException {
+        //given
+        Envelope newEnvelope = updateEnvelope("envelopes/new-envelope.json");
+        caseDetails = ccdCaseCreator.createCase(newEnvelope);
+
+        // when
+        callCallbackUrlToAttachExceptionToCase();
```
rkondratowicz: IMHO it would be much better if this took a parameter instead of relying on global state.
See comments below.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -95,34 +93,61 @@ public void setup() throws InterruptedException, ServiceBusException, JSONExcept
...
             .builder()
-            .eventId("attach_case")
-            .caseDetails(exceptionRecordCase)
-            .caseDetailsBefore(existingCase)
+            .eventId("attachToExistingCase")
+            .caseDetails(exceptionRecord)
             .build();
 
-        Response callbackResponse = RestAssured
+        RestAssured
```
KevinJCross: So im still not convinced this is the right approach for a "Functional test"
The integration tests are comprehensive and do exactly this.

I still strongly believe we should be poking ccd to make the callback to orchestrator.
This the the only code path that is not tested in our automated testing and needs to be done.
and this is the bit that is the easiest to break too.

KevinJCross: @vakepati

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -95,34 +93,59 @@ public void setup() throws InterruptedException, ServiceBusException, JSONExcept
...
+        await("Exception record is attached to the case")
+            .atMost(60, TimeUnit.SECONDS)
+            .pollDelay(2, TimeUnit.SECONDS)
+            .until(() -> isExceptionRecordAttachedToTheCase(1));
+    }
+
+    @Test
+    public void should_atatch_exception_record_to_the_existing_case_with_evidence_documents() throws JSONException {
+        //given
+        Envelope newEnvelope = updateEnvelope("envelopes/new-envelope-with-evidence.json");
```
KevinJCross: why just put the new-Envelope as a immutable object (that you can copy to change) on a Test Utility class. This would save a the loading parsing.
i.e. follow the  DRY principle.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -95,34 +93,59 @@ public void setup() throws InterruptedException, ServiceBusException, JSONExcept
...
+        callCallbackUrlToAttachExceptionToCase(caseDetails, exceptionRecord);
+
+        //then
+        await("Exception record is attached to the case")
+            .atMost(60, TimeUnit.SECONDS)
+            .pollDelay(2, TimeUnit.SECONDS)
+            .until(() -> isExceptionRecordAttachedToTheCase(2));
+    }
+
+    private void callCallbackUrlToAttachExceptionToCase(CaseDetails caseDetails, CaseDetails exceptionRecord) {
```
nitinprabhu: may be better call it as invokeCallbackEndpointForLinkingDocsToCase ?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -95,34 +93,59 @@ public void setup() throws InterruptedException, ServiceBusException, JSONExcept
...
             .pollDelay(2, TimeUnit.SECONDS)
             .until(() -> isExceptionRecordCreated(randomPoBox));
-
     }
 
     @Test
-    @Disabled
     public void should_atatch_exception_record_to_the_existing_case() {
+        //given
+        String caseData = SampleData.fileContentAsString("envelopes/new-envelope.json");
```
KevinJCross: maybe put this into a method call createNewCase() that returns caseDetails

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
nitinprabhu: You still need this functional test as you are just mocking in the integration which doesn't assures that it will update case in CCD when call-back endpoint is invoked. If you have concerns around this please take it on slack

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -95,34 +93,59 @@ public void setup() throws InterruptedException, ServiceBusException, JSONExcept
...
+        Envelope newEnvelope = EnvelopeParser.parse(caseData);
+        caseDetails = ccdCaseCreator.createCase(newEnvelope);
+
+        // when
+        callCallbackUrlToAttachExceptionToCase(caseDetails, exceptionRecord);
+
+        //then
+        await("Exception record is attached to the case")
+            .atMost(60, TimeUnit.SECONDS)
```
KevinJCross: why 60 seconds ? ... Rather make this a global define so we can modify this quickly and easily if ccd is slow or fast ( rather fail tests quickly if possible).

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -95,34 +93,59 @@ public void setup() throws InterruptedException, ServiceBusException, JSONExcept
...
+            .pollDelay(2, TimeUnit.SECONDS)
```
KevinJCross: Why 2 seconds ?
Maybe this should also be a global define that we can tweek

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -95,34 +93,59 @@ public void setup() throws InterruptedException, ServiceBusException, JSONExcept
...
+            .until(() -> isExceptionRecordAttachedToTheCase(1));
```
KevinJCross: How about extracting a Ccd Asserter (that is used globally so we can tweek it for all tests easily) that you can just pass in a lambda 
or have "expectations" predefined if they are used often ?
DRY principle again :)

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -134,33 +157,36 @@ private boolean isExceptionRecordCreated(UUID randomPoBox) {
...
         );
 
-        exceptionRecordCase = caseDetailsList.isEmpty() ? null : caseDetailsList.get(0);
-        return exceptionRecordCase != null;
+        exceptionRecord = caseDetailsList.isEmpty() ? null : caseDetailsList.get(0);
+        return exceptionRecord != null;
     }
 
-    private void verifyExistingCaseIsUpdatedWithExceptionRecordCaseDetails() {
+    private Boolean isExceptionRecordAttachedToTheCase(int expectedScannedDocsSize) {
```
KevinJCross: I dont think this is specific enough for a test assertion.
We should be asserting on the document number and the dm reference at the very least.

These should be passed in via a varargs parameter.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -134,33 +157,36 @@ private boolean isExceptionRecordCreated(UUID randomPoBox) {
...
         assertThat(scannedDocuments).isNotEmpty();
-        assertThat(scannedDocuments.size()).isEqualTo(1);
-
         assertThat(updatedScannedDocuments).isNotEmpty();
-        assertThat(updatedScannedDocuments.size()).isEqualTo(1);
 
-        ScannedDocument updatedCaseEvidence = updatedScannedDocuments.get(0);
-        ScannedDocument caseEvidence = scannedDocuments.get(0);
+        return updatedScannedDocuments.size() == expectedScannedDocsSize
+            && updatedScannedDocuments.containsAll(scannedDocuments);
```
KevinJCross: so the size parameter is not actually needed here since we do a contains all .
why not just `updatedScannedDocuments.size() == scannedDocuments.size` ?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
KevinJCross: @nitinprabhu No doubt we need the interactions in this test so having this is better than not but we should also be testing the actual callback which can easily be broken ( I broke it in the demo for instance)

nitinprabhu: that would be contract testing. Callback is invoked as part of click of submit event for attach to existing case.

nitinprabhu: If you mess with the definition callback will be broken any ways

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -95,34 +93,59 @@ public void setup() throws InterruptedException, ServiceBusException, JSONExcept
...
-
     }
 
     @Test
-    @Disabled
     public void should_atatch_exception_record_to_the_existing_case() {
+        //given
+        String caseData = SampleData.fileContentAsString("envelopes/new-envelope.json");
+        Envelope newEnvelope = EnvelopeParser.parse(caseData);
+        caseDetails = ccdCaseCreator.createCase(newEnvelope);
```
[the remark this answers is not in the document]
aliveni: I have 2 test cases, one case with no evidence and the other case with evidence. So it can't be moved to setup

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
aliveni: Updated the test to pass the parameters

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
aliveni: Done

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -95,34 +93,59 @@ public void setup() throws InterruptedException, ServiceBusException, JSONExcept
...
+
+        //then
+        await("Exception record is attached to the case")
+            .atMost(60, TimeUnit.SECONDS)
+            .pollDelay(2, TimeUnit.SECONDS)
+            .until(() -> isExceptionRecordAttachedToTheCase(1));
+    }
+
+    @Test
+    public void should_atatch_exception_record_to_the_existing_case_with_evidence_documents() throws JSONException {
```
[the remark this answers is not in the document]
aliveni: done

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
aliveni: @KevinJCross I've updated the test now. Could you please take another look?
assertThat(updatedScannedDocuments).containsAll(exceptionRecordDocuments);

the above assertion verifies all the fields in the ScannedDocument fields in the ScannedDcoument-> equals() method

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
aliveni: done

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -95,34 +93,59 @@ public void setup() throws InterruptedException, ServiceBusException, JSONExcept
             .atMost(60, TimeUnit.SECONDS)
             .pollDelay(2, TimeUnit.SECONDS)
             .until(() -> isExceptionRecordCreated(randomPoBox));
-
     }
 
     @Test
-    @Disabled
     public void should_atatch_exception_record_to_the_existing_case() {
```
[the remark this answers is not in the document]
aliveni: done

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
aliveni: renamed

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
aliveni: updated the assertions now

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
aliveni: I'll do it as part of another pr. This change needs to be made in all functional tests

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
aliveni: This applies to all functional tests, will do in separate pr/story

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
aliveni: same here

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
aliveni: again, need to change in all functional tests

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -134,33 +154,57 @@ private boolean isExceptionRecordCreated(UUID randomPoBox) {
...
-            String.valueOf(existingCase.getId())
+            caseDetails.getJurisdiction(),
+            String.valueOf(caseDetails.getId())
         );
 
         List<ScannedDocument> updatedScannedDocuments = getScannedDocumentsForSupplementaryEvidence(updatedCase);
 
-        List<ScannedDocument> scannedDocuments = getScannedDocumentsForExceptionRecord(exceptionRecordCase);
+        return !updatedScannedDocuments.isEmpty()
+            && updatedScannedDocuments.size() == expectedScannedDocsSize;
```
rkondratowicz: `!updatedScannedDocuments.isEmpty()` is not needed

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
KevinJCross: sure ... open a card :)

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
KevinJCross: yup ... although you can do it here then use the card to do it globally

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
KevinJCross: hmmm not sure I agree ... contract testing would be similar to our integration test and this.
Im dont believe that contract testing would actually cover this due to the dynamic nature of the spreadsheets so likelyhood is that this particular call would never properly fall under those sets of tests.

Anyway there is still a gap of testing the actual callback from ccd in the environment before the code goes to production.
I think that there should be a pipeline for publishing the spreadsheet that kicks off an orchestrator build that checks that it can still interoperate with ccd. 
This is why I think this should be included in our functional tests to ensure that it does not break and go live broken.

Obviously that is my opinion from my experience in CD ... so it should be up to QA to decide if the current test/s are good enough for CD.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
aliveni: ok, removed it now

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -95,34 +87,62 @@ public void setup() throws InterruptedException, ServiceBusException, JSONExcept
...
+        verifyExistingCaseIsUpdatedWithExceptionRecordData(caseDetails, exceptionRecord);
     }
 
     @Test
-    @Disabled
-    public void should_atatch_exception_record_to_the_existing_case() {
+    public void should_attach_exception_record_to_the_existing_case_with_evidence_documents() throws JSONException {
+        //given
+        Envelope newEnvelope = updateEnvelope("envelopes/new-envelope-with-evidence.json");
+        CaseDetails caseDetails = ccdCaseCreator.createCase(newEnvelope);
```
KevinJCross: better but the newEnvelope is always the same ... why not build it in make it static and re-use it in other tests :D

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -95,34 +87,62 @@ public void setup() throws InterruptedException, ServiceBusException, JSONExcept
...
         // when
+        invokeCallbackEndpointForLinkingDocsToCase(caseDetails, exceptionRecord);
+
+        //then
+        await("Exception record is attached to the case")
+            .atMost(60, TimeUnit.SECONDS)
+            .pollDelay(2, TimeUnit.SECONDS)
+            .until(() -> isExceptionRecordAttachedToTheCase(caseDetails, 2));
+
+        verifyExistingCaseIsUpdatedWithExceptionRecordData(caseDetails, exceptionRecord);
```
KevinJCross: Why not create an object called Verify and a inner class ExistingCase with a static method IsUpdatedWithExceptionRecord() ?
make it easier to read an re-useable from other tests.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -94,73 +87,124 @@ public void setup() throws InterruptedException, ServiceBusException, JSONExcept
...
         List<ScannedDocument> updatedScannedDocuments = getScannedDocumentsForSupplementaryEvidence(updatedCase);
 
-        List<ScannedDocument> scannedDocuments = getScannedDocumentsForExceptionRecord(exceptionRecordCase);
+        return updatedScannedDocuments.size() == expectedScannedDocsSize;
+    }
 
-        assertThat(scannedDocuments).isNotEmpty();
-        assertThat(scannedDocuments.size()).isEqualTo(1);
+    private void verifyExistingCaseIsUpdatedWithExceptionRecordData(
+        CaseDetails caseDetails,
```
KevinJCross: Its too tempting in this method to use the old caseDetails to check for the documents.
It would be better to pass in 
```
           caseDetails.getJurisdiction(),
            String.valueOf(caseDetails.getId())
```
and force the method/maintainer to look the case up

Also Since the method is called `ExceptionRecordData` you should just pass that in too 
Id have expected a method signature similar to 
verifyCaseHasDocuments( jurisdiction, caseRef, List<ScannedDocuments> )
then build the list in the test to check that the updated test contains the documents you are expecting.

PR #161, opened 2018-11-15
Fix OWASP checks about Guava

doncem:
### Change description ###

3 places required attention:

- 20.0 version coming together with app insights sorted by "fix" applied with library version bump from 2.1.1 to 2.1.2
- 20.0 version comes in with wiremock - yet another shade 😞 
- by providing 27 version in the list of dependencies does not really mean it is applied. apparently. moving to resolution strategy in global config

```
[ ] Yes
[x] No
```


build.gradle
```diff
@@ -157,12 +157,18 @@ repositories {
...
-  reformLogging   : '3.0.2',
+  reformLogging   : '3.0.3',
   springBoot      : springBoot.class.package.implementationVersion,
   springfoxSwagger: '2.9.2',
   junit           : '5.3.1'
 ]
 
+configurations.all {
+  resolutionStrategy {
+    force "com.google.guava:guava:27.0-jre"
```
lgonczar: Is it safe?

lgonczar: Fair enough. It would be worth adding a comment explaining why we're forcing this version.

doncem: in git annotations you probably will find the commit message explaining it but fair point - can also add a comment

PR #167, opened 2018-11-21
Fix `evidenceHandled` flag

doncem:

[Evidence handle flag not updated when case is updated](https://tools.hmcts.net/jira/browse/BPS-231)

### Change description ###

`null` case is ignored. Has to be specific. Since the CCD type is `YesOrNo` this means either `Yes` OR `No` value is desired.

```
[ ] Yes
[x] No
```


nitinprabhu:
Is there no change in integration and functional tests ? If not could you please add assertions that whenever a case is updated for supplementary evidence through BS component we always update the flag on case ?

doncem:
Yes, wanted to verify that if possible, will retry again today as could not set up environment properly on my machine yesterday.
Nope, there is no additional change in testing or elsewhere. Flag is strictly `Yes` or `No`. Tried boolean `false` and it brought back the error about invalid type of `YesOrNo`.

PR #169, opened 2018-11-22
Include OCR data in documents

lgonczar:

Part of https://tools.hmcts.net/jira/browse/BPS-247

### Change description ###

Read OCR data from the queue messages.

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/SampleData.java
```diff
@@ -139,7 +140,8 @@ public static Envelope envelope(int numberOfDocuments) {
                     String.format("control_number_%s", index),
                     String.format("type_%s", index),
                     LocalDate.parse("2018-10-01").plus(index, DAYS).atStartOfDay().toInstant(ZoneOffset.UTC),
-                    String.format("https://example.gov.uk/%s", index)
+                    String.format("https://example.gov.uk/%s", index),
+                    Base64.getEncoder().encodeToString(String.format("test_%s", index).getBytes())
```
nitinprabhu: could we use data similar to which our supplier is going to send us ?

lgonczar: I created a minimal example of such data (single element) and it was already a not-so-short string. From unit tests' point of view (those that test deserialisation of messages) it doesn't even matter if it's Base64.

For other reviewers: there's a separate discussion about having the actual OCR data (String->String map) in this field, which we might end up with.

PR #173, opened 2018-12-03
Make the set of supported jurisdictions configurable

lgonczar:
### Change description ###

Make it configurable which jurisdictions should be supported by the orchestrator, so that we can support integration testing on aat before given service goes live, while keeping prod unaffected.

```
[ ] Yes
[x] No
```


lgonczar:
Originally I included enabling Divorce in aat, but in the end decided not to mix these things.

aliveni:
@lgonczar Can you please enable Probate in aat? Probate will start using aat today

PR #175, opened 2018-12-06
Read OCR data from the queue

lgonczar:

### Change description ###

Include OCR data in the Document class that represents documents in queue messages. This is just a part of the solution. In future PRs this data will be sent to CCD.

```
[ ] Yes
[x] No
```


lgonczar:
Maybe it will get to that, but I'm thinking about a scenario where, depending on whether OCR data is null or not, we either include this field in CCD (overwriting the existing version) data or not (leaving the existing OCR data as it is).

PR #180, opened 2018-12-07
Mostly working helm chart

KevinJCross:
https://tools.hmcts.net/jira/browse/RPE-876

### Change description ###
This is the initial  commit of the first revision of the helm chart for bulk-scan orchestrator
The notes and documentation are to still to be added.
There still seems to be an authentication error with the $cbs location that looks like it needs to be worked on.
Make sure you use --wait on the helm for the post-install hooks to complete.


charts/bulk-scan-orchestrator/tests/test-service.yaml
```diff
@@ -0,0 +1,20 @@
+# ---
```
timja: needs an update

charts/runlocal.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+set -eu
+chart=./bulk-scan-orchestrator/
```
timja: other charts are using a makefile for this, any reason to use a shell script instead?

KevinJCross: This was to pick up our own orchestrator and deploy it from your machine .. Tbh we will just use draft in the future it is more just a helpful script.
I don't see the purpose to have a makefile under a gradle project, if we really wanted to integrate it we should have a gradle target to deploy our chart from local.

charts/bulk-scan-orchestrator/tests/test-service.yaml
KevinJCross: the problem is that there is nothing that really would indicate a proper deployment unless the health check fails if we can connect to the service bus.
I think that is outside of this card tbh.

KevinJCross:
This branch is closed **for Now** until the card is un-blocked by Bulk-Scan.
https://tools.hmcts.net/jira/browse/BPS-529

PR #181, opened 2018-12-10
Include OCR data in CCD data

lgonczar:

### Change description ###

- Include OCR data in CCD data. This involves moving OCR data from document to envelope, as that's how the processor is going to send it (https://github.com/hmcts/bulk-scan-processor/pull/323)
- Use system-default time zone when processing DateTimes. It was inconsistent and new tests failed because of it.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/CcdKeyValue.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/CcdKeyValue.java
@@ -1,0 +1,14 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class CcdKeyValue {
+
+    public final String key;
+    public final String value;
+
+    public CcdKeyValue(@JsonProperty("key") String key, @JsonProperty("value") String value) {
+        this.key = key;
+        this.value = value;
+    }
+}
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/CcdKeyValue.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/CcdKeyValue.java
@@ -1,13 +1,11 @@
 package uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
-
 public class CcdKeyValue {
 
     public final String key;
     public final String value;
 
-    public CcdKeyValue(@JsonProperty("key") String key, @JsonProperty("value") String value) {
+    public CcdKeyValue(String key, String value) {
         this.key = key;
         this.value = value;
     }
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/mappers/ExceptionRecordMapper.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/mappers/ExceptionRecordMapper.java
@@ -7,6 +7,7 @@ import uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd.ExceptionRecord;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.model.Envelope;
 
 import java.util.List;
+import java.util.Map;
 
 import static java.util.stream.Collectors.toList;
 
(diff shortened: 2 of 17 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/CcdKeyValue.java
```diff
@@ -0,0 +1,14 @@
...
+
+public class CcdKeyValue {
+
+    public final String key;
+    public final String value;
+
+    public CcdKeyValue(@JsonProperty("key") String key, @JsonProperty("value") String value) {
+        this.key = key;
+        this.value = value;
+    }
```
doncem: Do you need this constructor? The request body will be serialised from public fields assigned to this object.
Looking at the code below seems like more convenient way to construct it is from actual ocr map entry e.g.

```java
public CcdKeyValue(Map.Entry<String, String> entry) {
    this.key = entry.getKey();
    this.value = entry.getValue();
}
```

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/mappers/ExceptionRecordMapper.java
```diff
@@ -19,7 +25,24 @@ public ExceptionRecord mapEnvelope(Envelope envelope) {
             envelope.jurisdiction,
             getLocalDateTime(envelope.deliveryDate),
             getLocalDateTime(envelope.openingDate),
-            mapDocuments(envelope.documents)
+            mapDocuments(envelope.documents),
+            mapOcrData(envelope)
         );
     }
+
+    private List<CcdCollectionElement<CcdKeyValue>> mapOcrData(Envelope envelope) {
```
doncem: Can we just pass `ocrData` here instead? function doesn't really need anything from envelope - just map ocr data to something else

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/mappers/ExceptionRecordMapper.java
```diff
@@ -19,7 +25,24 @@ public ExceptionRecord mapEnvelope(Envelope envelope) {
...
+        if (envelope.ocrData != null) {
+            return envelope
+                .ocrData
+                .entrySet()
+                .stream()
+                .map(entry ->
+                    new CcdCollectionElement<>(
+                        new CcdKeyValue(entry.getKey(), entry.getValue())
+                    )
+                ).collect(toList());
```
doncem: Now giving the different constructor of `CcdKeyValue` object, we are able to stream this with no additional indentation :)
```java
return ocrData
    .entrySet()
    .stream()
    .map(CcdKeyValue::new)
    .map(CcdCollectionElement::new)
    .collect(toList());
```

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/mappers/ExceptionRecordMapper.java
```diff
@@ -19,7 +25,24 @@ public ExceptionRecord mapEnvelope(Envelope envelope) {
...
+        } else {
+            return null;
```
doncem: Still not favouring null for a collection field, but if we definitely sure it's the right way of implementation - it's fine (I did not just say that 🙊)

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/mappers/ExceptionRecordMapper.java
lgonczar: We can, will do

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/CcdKeyValue.java
lgonczar: I think it's clearer and in the long run easier to maintain when a data class receives its fields in the constructor and doesn't have any conversion logic, especially when the constructor is as simple as this one.

doncem: `CcdKeyValue` type is essentially `Map.Entry` anyway. Just thought it makes life easier if things are alike etc. Well, suggested along with the mapping function really 😉.
The question whether you need constructor was incomplete. I think we don't need `@JsonProperty` annotations there for sure

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/mappers/ExceptionRecordMapper.java
lgonczar: From what I've been told, that's going to make a difference in CCD's UI (presence of an empty section or the lack of it)

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/CcdKeyValue.java
lgonczar: You're right about `@JsonProperty`. Well spotted, I don't know what I was thinking.

PR #182, opened 2018-12-10
Handle messages with NEW_APPLICATION classification

lgonczar:

### Change description ###

Handle messages with NEW_APPLICATION classification.
Also, do a tiny refactoring around message creation in ExceptionRecordCreatorTest.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
@@ -36,6 +36,7 @@ import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
+import static java.util.stream.Collectors.toMap;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.awaitility.Awaitility.await;
 import static uk.gov.hmcts.reform.bulkscan.orchestrator.helper.ScannedDocumentsExtractor.getScannedDocuments;
@@ -195,6 +196,12 @@ public class AttachExceptionRecordToExistingCaseTest {
             exceptionRecord.getId(),
             expectedExceptionRecordsSize
         );
+
+        Map<String, String> exceptionRecordOcrData = getOcrData(exceptionRecord);
+
+        if (exceptionRecordOcrData != null) {
+            assertThat(getOcrData(updatedCase)).isEqualTo(getOcrData(exceptionRecord));
+        }
     }
 
     private void isExceptionReferenceAttachedToTheScannedDocuments(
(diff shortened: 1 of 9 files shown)


src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -227,4 +233,9 @@ private Envelope updateEnvelope(String envelope) throws JSONException {
         }
         return EnvelopeParser.parse(updatedCaseData.toString());
     }
+
+    @SuppressWarnings("unchecked")
+    private Map<String, String> getOcrData(CaseDetails caseDetails) {
+        return (Map<String, String>)caseDetails.getData().get("scanOCRData");
```
doncem: CI: missing space after cast

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EventPublisherContainer.java
```diff
@@ -44,8 +44,8 @@ public EventPublisher getPublisher(
                     caseDetails
                 );
             case EXCEPTION:
-                return exceptionRecordCreator;
             case NEW_APPLICATION:
+                return exceptionRecordCreator;
```
doncem: wait a minute. new application case type is same/similar as exception record?

lgonczar: Yes. Felt weird to me, too. Data will be different, but the orchestrator is supposed to process those exactly the same way.

nitinprabhu: only difference is exception may be for supplementary evidence or it may be for new application. So when it is exception classification orchestrator should send everything.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -195,6 +192,22 @@ private void verifyExistingCaseIsUpdatedWithExceptionRecordData(
             exceptionRecord.getId(),
             expectedExceptionRecordsSize
         );
+
+        assertCorrectOcrDataAfterUpdate(originalCase, exceptionRecord, updatedCase);
```
nitinprabhu: Ocr data is always present as part of exception record creation for new application. Whenever a case is updated we should never update/add ocr data. So I am not sure why we have this assertion ?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/helper/CaseDataExtractor.java
```diff
@@ -28,7 +29,27 @@ private ScannedDocumentsExtractor() {
...
+                .stream()
+                .map(ccdCollectionElement -> ((Map<String, String>) ccdCollectionElement.get("value")))
+                .collect(
+                    toMap(
+                        map -> map.get("key"),
+                        map -> map.get("value")
+                    )
+                );
+        } else {
+            return null;
```
rkondratowicz: perhaps an empty map would be safer?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
lgonczar: Ok, I assumed that exception record can change OCR data. Will remove this case.

Does it mean OCR data will only be included when classification is NEW_APPLICATON?

nitinprabhu: yes that's correct it will only when classification is NEW_APPLICATON or EXCEPTION(because this case might be any)

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/helper/CaseDataExtractor.java
lgonczar: Given we're only verifying data from CCD (not sending to them), can be done.

PR #183, opened 2018-12-11
BPS-400 Handle NullPointerException when scanned documents are null in ccd case

aliveni:
https://tools.hmcts.net/jira/browse/BPS-400

### Change description ###
Added assertions to check ccd data and scanned documents are not null

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/helper/ScannedDocumentsHelper.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/helper/ScannedDocumentsHelper.java
@@ -3,12 +3,12 @@ package uk.gov.hmcts.reform.bulkscan.orchestrator.helper;
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
-import org.springframework.util.Assert;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd.ScannedDocument;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.model.Document;
 import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
 
 import java.time.ZoneId;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/helper/ScannedDocumentsHelper.java
```diff
@@ -29,9 +30,14 @@ private ScannedDocumentsHelper() {
 
     @SuppressWarnings("unchecked")
     public static List<Document> getDocuments(CaseDetails caseDetails) {
-        List<Map<String, Object>> data = (List<Map<String, Object>>) caseDetails.getData().get("scannedDocuments");
+        Map<String, Object> caseData = caseDetails.getData();
+        Assert.notNull(caseData, "Case Data must not be null");
```
rkondratowicz: where is `getDocuments` used?
Do we actually want to throw `IllegalArgumentException` or return an empty list??

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/helper/ScannedDocumentsHelper.java
```diff
@@ -29,9 +30,14 @@ private ScannedDocumentsHelper() {
...
 
-        return data.stream()
+        List<Map<String, Object>> scannedDocuments = (List<Map<String, Object>>) caseData.get("scannedDocuments");
+        // ccd returns an empty list when scanned documents are not present.
+        Assert.notNull(scannedDocuments, "Scanned Documents must not be null");
```
lgonczar: Should we fail in this case? Or maybe we should set it to null?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/helper/ScannedDocumentsHelper.java
lgonczar: I guess it's used for extracting existing documents from case data before adding more. As for the error, maybe we should just treat it as an empty case to add fields to? I don't know myself, just asking.

aliveni: It's being used when attaching evidence to an existing case - `AttachDocsToSupplementaryEvidence.buildCaseDataContent()`. 
we want to throw `IllegalArgumentException` as ccd data is saved as empty collection in database when no scanned documents or case data is not provided

aliveni: After discussing with probate, some user roles (eg: solicitor) don't have access to scannedDocuments field. In such cases, scannedDocuments are null in the ccd case. So I'm changing the code to return empty list

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/helper/ScannedDocumentsHelper.java
aliveni: I'm changing the code to return an empty list

lgonczar:
👍

PR #189, opened 2018-12-13
See if tests pass with no changes

timja:
pretty sure my other PR shouldn't be failing but let's see


timja:
Yup flakey tests...

PR #190, opened 2018-12-17
Make sure OCR data is parsed and mapped to CCD object as received

doncem:
### Change description ###

No other interactions with ocr data found - just parsing from service bus and mapping to ccd collection items. Suggesting similar approach for deserialisation and additional check of java api: `Collector.toList` uses `ArrayList` so should be fine with element collection/addition

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/feature/OcrDataOrderTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/feature/OcrDataOrderTest.java
@@ -1,0 +1,60 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.feature;
+
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.SampleData;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.helper.CcdCollectionElementComparator;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd.CcdCollectionElement;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd.CcdKeyValue;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd.ExceptionRecord;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd.mappers.ExceptionRecordMapper;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.EnvelopeParser;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.model.Envelope;
+
+import java.util.ArrayList;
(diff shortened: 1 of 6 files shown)


src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/feature/OcrDataOrderTest.java
```diff
@@ -0,0 +1,60 @@
...
+        assertThat(record.ocrData).isInstanceOf(ArrayList.class);
+        assertThat(record.ocrData.size()).isEqualTo(envelope.ocrData.size());
+
+        // then
+        int i = 0;
+        Iterator<Map.Entry<String, String>> entries = envelope.ocrData.entrySet().iterator();
+
+        while (entries.hasNext()) {
+            Map.Entry<String, String> expectedEntry = entries.next();
+            assertThat(record.ocrData.get(i).value.key).isEqualTo(expectedEntry.getKey());
```
nitinprabhu: do we need these asserts ? We already have below asserts to validate they are in same order ?

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/feature/OcrDataOrderTest.java
```diff
@@ -0,0 +1,60 @@
...
+            assertThat(record.ocrData.get(i).value.value).isEqualTo(expectedEntry.getValue());
+            i++;
+        }
+
+        // and (just for sanity/clearance)
+        assertThat(record.ocrData)
+            .usingElementComparator(new CcdCollectionElementComparator())
+            .containsExactly(
```
nitinprabhu: can we containsExactly in same order or containsExactly does it by default ?

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/helper/CcdCollectionElementComparator.java
```diff
@@ -0,0 +1,14 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.helper;
+
+import uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd.CcdCollectionElement;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd.CcdKeyValue;
+
+import java.util.Comparator;
+
+public class CcdCollectionElementComparator implements Comparator<CcdCollectionElement<CcdKeyValue>> {
```
nitinprabhu: If this used for test class can we please keep as in inner class ?

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/feature/OcrDataOrderTest.java
doncem: I would prefer just keeping a loop which assures any size of ocr data order. i think i included the below section for giggles. for people who prefer reading text rather than code

PR #202, opened 2019-01-09
Update variables.tf

nitinprabhu:


lgonczar:
Does it also need an update in mapping to users?

PR #206, opened 2019-01-10
Prevent messages from being reprocessed immediately after failure

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-446

### Change description ###

Prevent messages from being reprocessed immediately after failure. This requires message handler to manage the finalise messages itself instead of throwing/not throwing exceptions.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
@@ -4,18 +4,24 @@ import com.google.common.base.Strings;
 import com.microsoft.azure.servicebus.ExceptionPhase;
 import com.microsoft.azure.servicebus.IMessage;
 import com.microsoft.azure.servicebus.IMessageHandler;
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
```diff
@@ -70,18 +83,82 @@ private void process(IMessage message) {
...
     @Override
     public void notifyException(Throwable exception, ExceptionPhase phase) {
         log.error("Error while handling message at stage: " + phase, exception);
     }
+
+    class MessageProcessingResult {
+        public final MessageProcessingResultType resultType;
+        public final Exception exception;
+
+        public MessageProcessingResult(MessageProcessingResultType resultType) {
```
rkondratowicz: should it be legal to create a result of type `UNRECOVERABLE_FAILURE` without providing an exception?

lgonczar: no, why?

PR #210, opened 2019-01-15
Copy document subtype values to document type 

aliveni:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/BPS-449

### Change description ###
Supplementary Evidence documents shouldn't contain "subtype" values.
If subtype values exist in the service bus message, copy subtype value to the type.
TODO: In the next pr, a functional test will be added.

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[ x] No
```

PR #211, opened 2019-01-15
Convert ModelMapper abstract class to interface

aliveni:
This is required for the previous pr https://github.com/hmcts/bulk-scan-orchestrator/pull/210/
Qualitygate reported that ModelMapper should be converted to interface as I had to override mapDocument method in SupplementaryEvidenceMapper.


rkondratowicz:
any context for this?

aliveni:
This is required for the previous pr `https://github.com/hmcts/bulk-scan-orchestrator/pull/210/`
Qualitygate reported that ModelMapper should be converted to interface as I had to override `mapDocument` method in SupplementaryEvidenceMapper

PR #215, opened 2019-01-16
Stop using azure-servicebus-spring-boot-starter

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-452

### Change description ###

Stop using azure-servicebus-spring-boot-starter as part of preparation for using two queues

```
[ ] Yes
[x] No
```


lgonczar:
It is already excluded

lgonczar:
> I would vote for not beaning all this up, not registering `MessageOperations` as service either and just void-ly register handler in one config. It suppose to not be exposed for the application

It would be ideal. What I wanted to avoid, however, was the need to inject the dependencies of the handler in the config class, in order to create the handler. The handler itself needs QueueClient.

PR #216, opened 2019-01-16
Prepare 'envelopes' queue configuration for introduction of another queue

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-452

### Change description ###

Prepare 'envelopes' queue configuration for introduction of another queue.
Also, remove some unused configuration bits

```
[ ] Yes
[x] No
```

### DIFF
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -59,7 +59,7 @@ withPipeline(type, product, component) {
       def serviceBusConnectionStr = kubectl.getSecret(serviceBusSecret, namespace, "{.data.connectionString}")
       //Regex: Endpoint=sb:\/\/([-a-zA-Z0-9]+)\.servicebus\.windows\.net.*
       def serviceBusNamespace = serviceBusConnectionStr.substring(14).split("\\.")[0]
-      env.QUEUE_READ_CONN_STRING = "${serviceBusConnectionStr};EntityPath=envelopes"
+
       env.ENVELOPES_QUEUE_WRITE_CONN_STRING = "${serviceBusConnectionStr};EntityPath=envelopes"
 
       // create queue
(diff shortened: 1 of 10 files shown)


Jenkinsfile_CNP
```diff
@@ -59,8 +59,8 @@ withPipeline(type, product, component) {
       def serviceBusConnectionStr = kubectl.getSecret(serviceBusSecret, namespace, "{.data.connectionString}")
       //Regex: Endpoint=sb:\/\/([-a-zA-Z0-9]+)\.servicebus\.windows\.net.*
       def serviceBusNamespace = serviceBusConnectionStr.substring(14).split("\\.")[0]
-      env.QUEUE_READ_CONN_STRING = "${serviceBusConnectionStr};EntityPath=envelopes"
-      env.QUEUE_WRITE_CONN_STRING = "${serviceBusConnectionStr};EntityPath=envelopes"
+
+      env.ENVELOPES_QUEUE_WRITE_CONN_STRING = "${serviceBusConnectionStr};EntityPath=envelopes"
```
rkondratowicz: should it not be the 'read' conn string?

README.md
```diff
@@ -14,12 +14,13 @@ IDAM_USERS_BULKSCAN_PASSWORD = Password12
...
-    connection-string: XXXXX
-    read-interval: YYYY
+    envelopes:
+      connection-string: XXXXX
+      queue-name: YYYY
   ```
   - where: 
     - XXXX is the connection string from azure to the queue you intend to use.
-    - YYYY is optional but recommended to reduce this to 500 ms for speed of development.
+    - YYYY is the name of the queue
```
rkondratowicz: what's `envelopes` then?

Jenkinsfile_CNP
rkondratowicz: or is that for tests?

lgonczar: It should be `write` - it's for tests, not the app

README.md
lgonczar: Envelopes is the queue named `envelopes`. Alternatively, I can rename the prefixed in this PR to sth like `incoming-envelopes`. The new one will be `processed-envelopes`.

lgonczar:
> so we are replacing read connection with write for.. reading part? erm.. this pr looks like a clash of two similar tasks

I confused everyone. There are actually two changes in this PR, in separate commits. One change adds ENVELOPES_ prefix (more or less) and the other one removes unused configuration.

PR #219, opened 2019-01-17
Remove delegate publisher

rkondratowicz:


rkondratowicz:
this will require some cleanup of the existing solution first

PR #220, opened 2019-01-18
Simplify mappers

rkondratowicz:
interface is never used, mappers are never mocked in tests, can be static.


lgonczar:
Why static? We can inject and mock instances. Otherwise, we have to test everything down.

doncem:
> interface is never used

can be easily solved (#221)

> mappers are never mocked in tests

good!

> can be static

static content would be perfect as no hard dependencies in mappers. interface not working though for that. didn't want to go into deep but perhaps going back to abstract class could solve this and reference to the object as well so to keep actual publisher changes at minimum. but not sure worth keeping this way. + #221  seems lot's simpler

rkondratowicz:
@doncem 
> _interface is never used_
>can be easily solved (#221)

Imho there's nothing to solve, this interface is just not needed.

The end goal is to get rid of the base abstract class and the `DelegatePublisher`.
This is the first step.

PR #229, opened 2019-01-24
Add dependency suppression for jackson-databind

lgonczar:
### Change description ###

Add dependency suppression for jackson-databind

```
[ ] Yes
[x] No
```

### DIFF
--- a/config/owasp/suppressions.xml
+++ b/config/owasp/suppressions.xml
@@ -18,12 +18,12 @@
   </suppress>
 
   <suppress>
-  <notes><![CDATA[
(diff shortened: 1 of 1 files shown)


config/owasp/suppressions.xml
```diff
@@ -46,7 +46,17 @@
   </suppress>
 
   <suppress>
-    <notes><![CDATA[ slf4j only has a beta version released with the 'fix', looks like lots of changes in it... ]]></notes>
```
lgonczar: The diff is a big ugly - I only added a new `suppression` node

PR #234, opened 2019-01-28
Feature: custom actuator endpoint

doncem:

[Smoke test for verifying Idam credentials](https://tools.hmcts.net/jira/browse/BPS-478)

### Change description ###

let the actuator configure and manage endpoint which exposes the answer whether provided credentials for configured jurisdiction are correct or not

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpoint.java
```diff
@@ -0,0 +1,84 @@
...
+            logException(credential.getUsername(), exception);
+
+            return HttpStatus.valueOf(exception.status());
+        } catch (Exception exception) {
+            logException(credential.getUsername(), exception);
+
+            return HttpStatus.UNAUTHORIZED;
+        }
+
+        return HttpStatus.OK;
```
rkondratowicz: should go to `try` block.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpoint.java
```diff
@@ -0,0 +1,84 @@
...
+
+            log.debug("Successful authentication");
+        } catch (FeignException exception) {
+            logException(credential.getUsername(), exception);
+
+            return HttpStatus.valueOf(exception.status());
+        } catch (Exception exception) {
+            logException(credential.getUsername(), exception);
+
+            return HttpStatus.UNAUTHORIZED;
```
rkondratowicz: more like 500 imho

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpoint.java
```diff
@@ -0,0 +1,84 @@
...
+                entry -> checkCredentials(entry.getValue())
+            ));
+    }
+
+    @ReadOperation
+    public HttpStatus jurisdiction(@Selector String jurisdiction) {
+        try {
+            return checkCredentials(credential.apply(jurisdiction));
+        } catch (NoUserConfiguredException exception) {
+            return HttpStatus.UNAUTHORIZED;
```
rkondratowicz: more like 400 imho

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpoint.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import static java.util.Map.Entry;
+
+@Component
+@Endpoint(id = "jurisdictions")
+@EnableConfigurationProperties(JurisdictionToUserMapping.class)
+public class JurisdictionEndpoint {
```
rkondratowicz: name not clear. this endpoint does not return jurisdictions.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpoint.java
```diff
@@ -0,0 +1,84 @@
...
+
+    private static final Logger log = LoggerFactory.getLogger(JurisdictionEndpoint.class);
+
+    private final Map<String, Credential> jurisdictions;
+    private final Function<String, Credential> credential;
```
rkondratowicz: surely this can be clearer

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpoint.java
```diff
@@ -0,0 +1,84 @@
...
+            logException(credential.getUsername(), exception);
+
+            return HttpStatus.UNAUTHORIZED;
+        }
+
+        return HttpStatus.OK;
+    }
+
+    private void logException(String username, Exception exception) {
+        log.error("Received an error while authentication {}", username, exception);
```
rkondratowicz: Did you mean 'authenticating'?

rkondratowicz: should include jurisdiction

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpoint.java
rkondratowicz: Why not a controller?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpoint.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private HttpStatus checkCredentials(String jurisdiction, Credential credential) {
+        try {
+            idamClient.authenticateUser(credential.getUsername(), credential.getPassword());
+
+            log.debug("Successful authentication");
+
+            return HttpStatus.OK;
+        } catch (FeignException exception) {
+            logException(jurisdiction, credential.getUsername(), exception);
```
lgonczar: This catch doesn't represent an error situation for this endpoint (even though for the service as a whole it is). I think it would be best to log it as nothing more severe than warning and let the caller decide what to do with errors coming from IdAM.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpoint.java
```diff
@@ -0,0 +1,89 @@
...
+        } catch (Exception exception) {
+            logException(jurisdiction, credential.getUsername(), exception);
+
+            return HttpStatus.INTERNAL_SERVER_ERROR;
+        }
+    }
+
+    private void logException(String jurisdiction, String username, Exception exception) {
+        log.error(
+            "Received an error while authenticating {} jurisdiction with {} username",
```
lgonczar: Maybe something like "An error occurred..."? We didn't necessarily receive an error from the called service.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpoint.java
```diff
@@ -0,0 +1,89 @@
...
+
+            return HttpStatus.OK;
+        } catch (FeignException exception) {
+            logException(jurisdiction, credential.getUsername(), exception);
+
+            return HttpStatus.valueOf(exception.status());
+        } catch (Exception exception) {
+            logException(jurisdiction, credential.getUsername(), exception);
+
+            return HttpStatus.INTERNAL_SERVER_ERROR;
```
lgonczar: This catch represents a case when we failed trying to process the request. Returning and `INTERNAL_SERVER_ERROR` status in this method would make it indistinguishable from the situation when the called service returned a 500. I think we should not catch the exception, but instead let the global handler deal with it (including logging it).

src/smokeTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/JurisdictionAuthenticationTest.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class JurisdictionAuthenticationTest {
+
+    private static final String TEST_URL;
+
+    static {
+        TEST_URL = ConfigFactory.load().getString("test-url");
```
lgonczar: Does this initiation need to be split from definition?

src/smokeTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/JurisdictionAuthenticationTest.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    @DisplayName("Each configured jurisdiction should have valid credentials")
+    @Test
+    public void jurisdictions() {
```
lgonczar: DisplayName is nice.

I'd still follow our test method convention, too.

src/smokeTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/JurisdictionAuthenticationTest.java
```diff
@@ -0,0 +1,36 @@
...
+        Map<String, HttpStatus> response = RestAssured
+            .given()
+            .relaxedHTTPSValidation()
+            .baseUri(TEST_URL)
+            .header(SyntheticHeaders.SYNTHETIC_TEST_SOURCE, "Bulk Scan Orchestrator smoke test")
+            .get("/jurisdictions")
+            .as(new TypeRef<Map<String, HttpStatus>>() {});
+
+        assertThat(response.values()).containsOnly(HttpStatus.OK);
```
lgonczar: Maybe we should compare the whole maps, so that when there's a problem with one of the jurisdictions, the test will tell us which one.

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpointTest.java
```diff
@@ -0,0 +1,109 @@
...
+    @DisplayName("Should respond NOT FOUND for specific jurisdiction which does not exist")
+    @Test
+    public void singleNonExistingJurisdiction() {
+        assertThat(endpoint.jurisdiction("not give access")).isEqualTo(HttpStatus.NOT_FOUND);
+        verify(idamClient, never()).authenticateUser(anyString(), anyString());
+    }
+
+    @DisplayName("Should respond statuses of all configured jurisdictions")
+    @Test
+    public void jurisdictions() {
```
lgonczar: I'm ok with DisplayName being more descriptive than method name, as some kind of addition to it, but method name should also be descriptive.

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpointTest.java
```diff
@@ -0,0 +1,109 @@
...
+    public void singleValidJurisdiction() {
+        willReturn("token").given(idamClient).authenticateUser("username", "password");
+
+        assertThat(endpoint.jurisdiction(VALID_JURISDICTION)).isEqualTo(HttpStatus.OK);
+    }
+
+    @DisplayName("Should respond INTERNAL_SERVER_ERROR for specific jurisdiction which is incorrectly configured")
+    @Test
+    public void singleInvalidJurisdiction() {
+        willThrow(new RuntimeException("oh no")).given(idamClient).authenticateUser("user", "pass");
```
lgonczar: This test represents the case when idamClient throws some runtime exception. Is this what would happen if user credentials were invalid? Looking at code, I guess FeignException with the right status would be appropriate for this scenario.

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpointTest.java
```diff
@@ -0,0 +1,109 @@
...
+        JurisdictionToUserMapping mapping = new JurisdictionToUserMapping();
+        mapping.setUsers(USERS);
+
+        endpoint = new JurisdictionEndpoint(mapping, idamClient);
+    }
+
+    @DisplayName("Should respond OK for specific jurisdiction which is correctly configured")
+    @Test
+    public void singleValidJurisdiction() {
+        willReturn("token").given(idamClient).authenticateUser("username", "password");
```
lgonczar: Just a suggestion: `username`, `password`, `user`, `pass` could be referred to through variables describing when these values represent.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpoint.java
doncem: I've reverted these changes as following reviewer had different concerns. And the size of concerns already raises the suggestion we should discuss in person what team desires to respond in each situation

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpoint.java
doncem: This is configuration based rest endpoint. Since people have trouble with existing ones and I have a pretty strong position of **not** including it into health indicator list, it is beneficial to use custom actuator endpoint to achieve what this ticket has defined

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpoint.java
doncem: Please give more input as personally it is pretty clear

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpointTest.java
doncem: There is a scenario with `FeignException` ;). This is mimicking the situation when it is something else

doncem:
Rebase mainly addressed @lgonczar views as they are more on the topic. In terms of status codes please find the time and grab the team to discuss this in quick catch-up.
The rest I'm pretty sure is more than sufficient

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpoint.java
```diff
@@ -0,0 +1,93 @@
...
+                Entry::getKey,
+                entry -> checkCredentials(entry.getKey(), entry.getValue())
+            ));
+    }
+
+    @ReadOperation
+    public HttpStatus jurisdiction(@Selector String jurisdiction) {
+        try {
+            return checkCredentials(jurisdiction, credentialProvider.apply(jurisdiction));
+        } catch (NoUserConfiguredException exception) {
```
rkondratowicz: `checkCredentials` never throws an exception (it has a `catch (Exception exception)` inside), therefore this code is never reached.

rkondratowicz: also, missing test I guess...

doncem: You are correct `checkCredentials` doesn't throw an exception. But `credentialProvider.apply` does though:

```java
// assignment
credentialProvider = mapping::getUser;

// implementation
public Credential getUser(String jurisdiction) {
    return users.computeIfAbsent(jurisdiction.toLowerCase(), this::throwNotFound);
}

private Credential throwNotFound(String jurisdiction) {
    throw new NoUserConfiguredException(jurisdiction);
}
```

And the test is called `Should respond UNAUTHORISED for specific jurisdiction which does not exist` which verifies `idamClient` is never called.

To summarise: `checkCredentials` function should not throw an exception but always provide some status so the `/jurisdictions` endpoint always responds with success. Furthermore it parses all configured jurisdictions so you are safe from the one pointed out here ^. On the other hand `/jurisdictions/{value}` is free text and therefore can throw extra exception of jurisdiction not being present in config.

P.S. In case missed, coverage reports all lines/branches are satisfied:
> The diff coverage is `100%`.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpoint.java
rkondratowicz: Store `JurisdictionToUserMapping` and use it.
So instead of 
```
private final Map<String, Credential> jurisdictions;
private final Function<String, Credential> credentialProvider;
private final IdamClient idamClient;

public JurisdictionEndpoint(
    JurisdictionToUserMapping mapping,
    IdamClient idamClient
) {
    jurisdictions = mapping.getUsers();
    credentialProvider = mapping::getUser;
    this.idamClient = idamClient;
}
```
you can simply do
```
private JurisdictionToUserMapping jurisdictionToUserMapping;
private final IdamClient idamClient;

public JurisdictionEndpoint(
    JurisdictionToUserMapping mapping,
    IdamClient idamClient
) {
    this.jurisdictionToUserMapping = jurisdictionToUserMapping;
    this.idamClient = idamClient;
}
```

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpoint.java
rkondratowicz: I'm not suggesting a custom health indicator, just a controller.
Since we've configured the base path for management endpoints to be `/` I think there's no difference between a controller and `@Endpoint`.

src/smokeTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/JurisdictionAuthenticationTest.java
```diff
@@ -0,0 +1,50 @@
...
+        mapper.readTree(response).elements().forEachRemaining(responseStatus -> {
+            JurisdictionConfigurationStatus status = new JurisdictionConfigurationStatus(
+                responseStatus.get("jurisdiction").asText(),
+                responseStatus.get("is_correct").asBoolean(),
+                responseStatus.get("error_description").asText()
+            );
+
+            assertThat(status)
+                .withFailMessage(failMessage, status.jurisdiction, status.errorDescription)
+                .isEqualToComparingFieldByField(
```
rkondratowicz: we're only interested in `is_correct`

doncem: if it is incorrect, the failure message will be present in smoke test log - can potentially help out to speed track the failure instead of going to portal app insights figuring out the cause, here's an example:

```bash
java.lang.AssertionError: Misconfigured local jurisdiction, error description: Connection refused (Connection refused) executing POST http://localhost:8080/oauth2/authorize?response_type=code&client_id=bsp&redirect_uri=http%3A%2F%2Flocalhost%2Freceiver. Check the logs for more details
```

rkondratowicz: that's the fail message configured in line above, right?
I'm saying that in assert we should check all items have `is_corrent == true`

src/smokeTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/JurisdictionAuthenticationTest.java
```diff
@@ -0,0 +1,50 @@
...
+                    new JurisdictionConfigurationStatus(status.jurisdiction, true)
```
rkondratowicz: is this asserting that error description is null?

src/smokeTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/JurisdictionAuthenticationTest.java
doncem: we do. along with jurisdiction. it's no biggy nor complicated. not sure what you are trying to achieve here as test is pretty obvious and straight-forward

rkondratowicz: it's checking the jurisdiction by comparing `status.jurisdiction` to `status.jurisdiction` though

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/JurisdictionEndpointTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+        endpoint = new JurisdictionEndpoint(mapping, idamClient);
+    }
+
+    @DisplayName("Should respond OK for specific jurisdiction which is correctly configured")
+    @Test
+    public void should_respond_200_when_specific_jurisdiction_is_correctly_configured() {
+        willReturn("token").given(idamClient).authenticateUser("username", "password");
+
+        assertThat(endpoint.jurisdiction(VALID_JURISDICTION)).isEqualToComparingFieldByField(VALID_RESPONSE);
```
rkondratowicz: should we not use `mockMvc` as we do for all other REST endpoints?

doncem: `mockMvc` is very much like integration test. and am actually using it in one

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/IdamConfigStatusEndpointTest.java
```diff
@@ -0,0 +1,128 @@
...
+            );
+
+        willThrow(exception).given(idamClient).authenticateUser("user", "pass");
+
+        // not asserting an object as depends on feign version what message it has inside.
+        // plus it is not of importance for unit test.
+        // we have other cases covering the fact description is provided as expected.
+        assertThat(endpoint.jurisdiction(INVALID_JURISDICTION))
+            .extracting("jurisdiction", "isCorrect")
+            .containsExactly(tuple(INVALID_JURISDICTION, false).toArray());
```
rkondratowicz: either test name is outdated or there's something missing here

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/IdamConfigStatusEndpointTest.java
```diff
@@ -0,0 +1,49 @@
...
+@IntegrationTest
+@AutoConfigureMockMvc
+public class IdamConfigStatusEndpointTest {
+
+    @Autowired
+    private MockMvc mvc;
+
+    @DisplayName("Should have BULKSCAN dummy jurisdiction configured properly")
+    @Test
+    public void should_have_bulkscan_configured_properly() throws Exception {
```
nitinprabhu: should it be not for all configured juridictions ?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/IdamConfigStatusEndpointTest.java
```diff
@@ -0,0 +1,49 @@
...
+            .getResponse()
+            .getContentAsByteArray();
+
+        JsonNode responseNode = mapper.readTree(response);
+
+        assertThat(responseNode.isArray()).isTrue();
+        assertThat(responseNode.elements()).hasSize(1);
+
+        JurisdictionConfigurationStatus actual = new JurisdictionConfigurationStatus(
+            responseNode.get(0).get("jurisdiction").asText(),
```
nitinprabhu: should we not assert all the configured jurisdictions and not just bulk scan ?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/ConfigurationTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+@RunWith(SpringRunner.class)
+@IntegrationTest
+public class ConfigurationTest {
+
+    @Autowired
+    private ApplicationContext context;
+
+    @Test
+    public void should_load_custom_actuator_endpoint_for_jurisdictions() {
```
nitinprabhu: A more better integration test would be to use wire mock but that could be done as a separate PR

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/IdamConfigStatusEndpoint.java
```diff
@@ -0,0 +1,83 @@
...
+        } catch (NoUserConfiguredException exception) {
+            return new JurisdictionConfigurationStatus(jurisdiction, false, exception.getMessage());
+        }
+    }
+
+    private JurisdictionConfigurationStatus checkCredentials(String jurisdiction, Credential credential) {
+        try {
+            idamClient.authenticateUser(credential.getUsername(), credential.getPassword());
+
+            log.debug("Successful authentication");
```
nitinprabhu: we can include jurisdiction in the log

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/IdamConfigStatusEndpointTest.java
doncem: it is integration test. so whatever is configured as per dev environment aka `integration` [profile (see the yaml here)](https://github.com/hmcts/bulk-scan-orchestrator/tree/master/src/test/resources)

everything is checked in smoke test

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/IdamConfigStatusEndpointTest.java
doncem: same - we should not care about all at this place. everything is checked in smoke test

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/IdamConfigStatusEndpoint.java
doncem: ok. done

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/IdamConfigStatusEndpointTest.java
```diff
@@ -0,0 +1,128 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.tuple;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.BDDMockito.never;
+import static org.mockito.BDDMockito.verify;
+import static org.mockito.BDDMockito.willReturn;
+import static org.mockito.BDDMockito.willThrow;
+
+@ExtendWith(MockitoExtension.class)
+public class IdamConfigStatusEndpointTest {
```
KevinJCross: surely this should be an integration style test ?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/IdamConfigStatusEndpointTest.java
aliveni: @nitinprabhu All the configured jurisdictions are being verified in the smoke test. I don't think we  need to verify them here

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/IdamConfigStatusEndpointTest.java
```diff
@@ -0,0 +1,49 @@
...
+            responseNode.get(0).get("is_correct").asBoolean(),
+            responseNode.get(0).get("error_description").isNull() ? null : "must be null - failure"
+        );
+
+        assertThat(actual).isEqualToComparingFieldByField(
```
nitinprabhu: It would be nice if we could assert as below or something similar 
```
assertThat(fellowshipOfTheRing).extracting("name", "age", "race.name")
                               .contains(tuple("Boromir", 37, "Man"),
                                         tuple("Sam", 38, "Hobbit"),
                                         tuple("Legolas", 1000, "Elf"));
```

src/smokeTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/IdamConfigStatusTest.java
```diff
@@ -0,0 +1,51 @@
...
+            .given()
+            .relaxedHTTPSValidation()
+            .baseUri(TEST_URL)
+            .header(SyntheticHeaders.SYNTHETIC_TEST_SOURCE, "Bulk Scan Orchestrator smoke test")
+            .get("/idam-config-status")
+            .andReturn()
+            .asByteArray();
+        ObjectMapper mapper = new ObjectMapper();
+
+        mapper.readTree(response).elements().forEachRemaining(responseStatus -> {
```
nitinprabhu: I would prefer bind the response to a list of JurisdictionConfigurationStatus objects and then can assert similar to below
```
assertThat(fellowshipOfTheRing).extracting("name", "age", "race.name")
                               .contains(tuple("Boromir", 37, "Man"),
                                         tuple("Sam", 38, "Hobbit"),
                                         tuple("Legolas", 1000, "Elf"));

```

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/endpoints/IdamConfigStatusEndpointTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+    private static final JurisdictionConfigurationStatus VALID_RESPONSE = new JurisdictionConfigurationStatus(
+        VALID_JURISDICTION, true
+    );
+
+    private static final JurisdictionConfigurationStatus INVALID_RESPONSE = new JurisdictionConfigurationStatus(
+        INVALID_JURISDICTION, false, "oh no"
+    );
+
+    private static final Map<String, Map<String, String>> USERS = ImmutableMap.of(
```
nitinprabhu: This can be defined inline as it is not used across the class

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/ConfigurationTest.java
```diff
@@ -0,0 +1,26 @@
...
+        String clazz = IdamConfigStatusEndpoint.class.getSimpleName();
+
+        assertThat(context.containsBean(clazz.substring(0, 1).toLowerCase() + clazz.substring(1))).isTrue();
```
nitinprabhu: If we can remove this test than that would be nice `IdamConfigStatusEndpointTest` anyways proves that the bean is loaded

PR #236, opened 2019-01-31
Updated idam url to point to idam test envionment

nitinprabhu:
### Change description ###

- Sprod cannot connect to Idam preproduction environment. Hence updated idam url to use the test environment.

```
[ ] Yes
[x ] No
```


infrastructure/sprod.tfvars
```diff
@@ -1,4 +1,4 @@
 capacity = "2"
-idam_api_url = "https://preprod-idamapi.reform.hmcts.net:3511"
+idam_api_url = "http://betadevbccidamapplb.reform.hmcts.net"
```
doncem: it's not even https?

nitinprabhu: https doesn't work anyways it would be soon replaced with strategic solution so should be fine for now.

PR #237, opened 2019-01-31
[DO NOT MERGE] BPS-416 Date fields values without milliseconds

aliveni:
Run functional tests to check if CCD creates Exception Record / add evidence to the case with Date format changes (in bulk-scan-processor PR https://github.com/hmcts/bulk-scan-processor/pull/466 )

### DIFF
--- a/src/functionalTest/resources/envelopes/new-envelope-with-evidence.json
+++ b/src/functionalTest/resources/envelopes/new-envelope-with-evidence.json
@@ -5,15 +5,15 @@
   "jurisdiction": "BULKSCAN",
   "classification": "NEW_APPLICATION",
   "zip_file_name": "zip-file-test.zip",
-  "delivery_date": "1970-01-01T00:00:00.000Z",
-  "opening_date": "1970-01-01T00:00:00.000Z",
+  "delivery_date": "1970-01-01T00:00:00Z",
+  "opening_date": "1970-01-01T00:00:00Z",
   "documents": [
     {
       "file_name": "certificate1.pdf",
       "control_number": "154565768",
       "type": "other",
       "subtype": null,
-      "scanned_at": "2018-01-01T10:30:00.000Z",
+      "scanned_at": "2018-01-01T10:30:00Z",
       "url": "http://dm-store:4460/documents/ee69aee8-1a33-40dd-9af9-d90da1b104ba"
     }
   ],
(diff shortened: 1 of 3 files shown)


lgonczar:
Why do we want to remove milliseconds from those dates? If we accept them with milliseconds, it seems to make most sense to represent them like that in the response.

aliveni:
> Why do we want to remove milliseconds from those dates? If we accept them with milliseconds, it seems to make most sense to represent them like that in the message. Otherwise we lose some information.

@lgonczar I don't want to remove milliseconds from the test data. I just want to run functional tests with the changes made in https://github.com/hmcts/bulk-scan-processor/pull/466
To verify if this date format is causing any issues in CCD.

If the metadata have date fields with time values as zeros eg: `1970-01-01T00:00:00.000Z`, `Instant/LocalDateTime` classes are ignoring the milliseconds part and convert it to "1970-01-01T00:00:00Z". 
It works same as before for date fields with non-zero time values eg: `2018-06-23T02:10:30.123Z`

If this can not agreed, we should probably keep it as Timestamp. (unless there is any other way representing the date values with milliseconds precision)

aliveni:
> Maybe we should configure Instant serialisation in the processor to keep the full `... .SSS'Z` format instead?

Sure, I'll try and update bulk-scan-processor PR

doncem:
not sure why such a hassle for tiny thing like passing around milliseconds. if it is strictly defined in parsers then yes - absolutely needed. on the other hand, tests will be in place and break the build.

it is normal to interpret time as is. object has the knowledge that "if i'm missing milliseconds - i will assume it's 0"

furthermore, iirc ccd also has `LocalDateTime`s so any requests will be interpreted accordingly and properly. it has been tested long time ago. the only thing one should be aware of is microseconds - there's an issue parsing milliseconds bit out of microseconds string signature

in any case, i understand your point of view as well

lgonczar:
> not sure why such a hassle for tiny thing like passing around milliseconds
> ...

My point is that if we require milliseconds on the input then we should pass them to CCD, as well, and test data should reflect that.

aliveni:
> > not sure why such a hassle for tiny thing like passing around milliseconds
> > ...
> 
> My point is that if we require milliseconds on the input (metadata file) then we should pass them to CCD, as well, and test data should reflect that.

We do pass milliseconds to the CCD if they are not zero in the input metadata file.

PR #240, opened 2019-02-06
Remove not needed mapper interface

rkondratowicz:
Interface was never really used.
Event publishers (`CreateExceptionRecord` and `AttachDocsToSupplementaryEvidence`) take concrete implementations.

More importantly, this will allow customisation of `SupplementaryEvidenceMapper` which will soon take existing documents as parameter, not just envelope from the queue.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/mappers/DocumentMapper.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/mappers/DocumentMapper.java
@@ -1,11 +1,9 @@
 package uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd.mappers;
 
-import uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd.CaseData;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd.CcdCollectionElement;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd.CcdDocument;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd.ScannedDocument;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.model.Document;
-import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.model.Envelope;
 
 import java.time.Instant;
 import java.time.LocalDateTime;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/mappers/SupplementaryEvidenceMapper.java
```diff
@@ -4,14 +4,15 @@
...
-public class SupplementaryEvidenceMapper implements ModelMapper<SupplementaryEvidence> {
+public class SupplementaryEvidenceMapper {
 
     public SupplementaryEvidenceMapper() {
         // empty mapper construct
     }
 
-    @Override
     public SupplementaryEvidence mapEnvelope(Envelope envelope) {
         return new SupplementaryEvidence(mapDocuments(envelope.documents));
```
rkondratowicz: As you can it only needs the documents.

After this PR is merge and interface is gone, I'll change this method to something like
```
SupplementaryEvidence map(List<Document> existingDocument, List<Document> envelopeDocuments)
```

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/mappers/DocumentMapper.java
```diff
@@ -14,19 +12,21 @@
 import java.util.List;
 import java.util.stream.Collectors;
 
-public interface ModelMapper<T extends CaseData> {
+public class DocumentMapper {
```
doncem: utility classes are final classes

lgonczar:
I'd prefer this to be an injected dependency with instance methods, but this can be done in the future.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/mappers/DocumentMapper.java
rkondratowicz: Just did a quick search on `// util class` and existing helper classes in this repo are not final.
As pmd and/or sonar don't complain, I'll leave it as is and spare myself 20min ci build waiting time ;)

doncem: ᕕ( ᐛ )ᕗ

PR #247, opened 2019-02-07
Clarify data used in test

rkondratowicz:


src/functionalTest/resources/envelopes/supplementary-evidence-envelope.json
```diff
@@ -14,7 +14,7 @@
       "type": "other",
       "subtype": null,
       "scanned_at": "2018-01-01T11:20:00.000Z",
-      "url": "http://dm-store:4460/documents/ee69aee8-1a33-40dd-9af9-d90da1b104babc"
+      "url": "==THIS WILL BE OVERWRITTEN IN TEST=="
```
aliveni: Is there any better way of doing this? 
we replace document url in this test file also `new-envelope-with-evidence.json`

PR #271, opened 2019-02-12
Update tidam  url to use sidam

nitinprabhu:


nitinprabhu:
Merging this as there is no impact on prod build as saat tfvars have been changed. AKS has issues and Sidam build needs to be asap hence merging.

PR #280, opened 2019-02-14
Remove case retriever

doncem:
### Change description ###

Same API is used in the callback service. Only referenced in `EnvelopeEventProcessor`. There's a minor difference that callback does handle the exception whereas processor needs to bypass it. Created `getCaseOptionally` which perhaps will raise doubt to reviewers. Feel free to suggest something. Take multiple review attempts if needed

```
[ ] Yes
[x] No
```


doncem:
AKS :(

PR #286, opened 2019-02-20
Remove DelegatePublisher

rkondratowicz:
# Context
There are 2 classes that handle envelopes and make changes in CCD:
- `AttachDocsToSupplementaryEvidence`
- `CreateExceptionRecord`

Both extend the same base class which implements `EventPublisher` interface.

There's also `EventPublisherContainer` which is called by the main job and returns the correct 'handler' appropriate for given envelope.

The main job then calls `.publish(envelope)` on it.

## Before
Because attaching doc to supplementary requires extra information, a `DelegatePublisher` class is returned from the container. It exposes the same `publish(Envelope env)` method, but calls `publish(Envelope envelope, String caseTypeId)` on `AttachDocsToSupplementaryEvidence` under the surface.
The base class implements the interface, so there are two methods on it, with default implementations:
```
    default void publish(Envelope envelope) {
        throw new NotImplementedException();
    }

    default void publish(Envelope envelope, String caseTypeId) {
        throw new NotImplementedException();
    }
```
Note that  **only the first one** is called by envelope processor.

## After
`AttachDocsToSupplementaryEvidence` and `CreateExceptionRecord` no longer share a common interface.
`EventPublisher` interface has only one method:
```
void publish(Envelope envelope);
```
`DelegatePublisher` is no longer needed, `EventPublisherContainer` returns a function that is then called by event processor.


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EventPublisherContainer.java
```diff
@@ -16,6 +16,7 @@
  *     <li>use resource in {@link this#getPublisher(Classification, Supplier)})}</li>
  * </ul>
  */
+@SuppressWarnings("squid:S1612")
```
rkondratowicz: lambdas vs method references

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EventPublisherContainer.java
```diff
@@ -40,11 +41,11 @@ public EventPublisher getPublisher(
                 CaseDetails caseDetails = caseRetrieval.get();
 
                 return caseDetails == null
-                    ? exceptionRecordCreator
-                    : new DelegatePublisher(attachDocsPublisher, caseDetails.getCaseTypeId());
+                    ? envelope -> exceptionRecordCreator.publish(envelope)
+                    : envelope -> attachDocsPublisher.publish(envelope, caseDetails.getCaseTypeId());
```
doncem: _or_ could make a single lambda like

```java
return envelope -> caseDetails.get()
    ? exceptionRecordCreator.publish(envelope)
    : attachDocsPublisher.publish(envelope, caseDetails.getCaseTypeId());
```

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EventPublisherContainer.java
doncem: my vote is for reference 🙃

rkondratowicz: I went with lambda for readability reasons:
```
return caseDetails == null
    ? envelope -> exceptionRecordCreator.publish(envelope)
    : envelope -> attachDocsPublisher.publish(envelope, caseDetails.getCaseTypeId());
```
vs
```
return caseDetails == null
    ? exceptionRecordCreator::publish
    : envelope -> attachDocsPublisher.publish(envelope, caseDetails.getCaseTypeId());
```
in the 1st version we can see what's passed to the function and the similarities between the two functions.

lgonczar: it reads better if it's consistent

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EventPublisherContainer.java
lgonczar: Not a bad idea with retrieving the case in the publisher. This way the method would just get the publisher, as advertised, without taking extra steps. That might affect exception handling, though.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EventPublisherContainer.java
doncem: i remember the position, just stating my vote. all good and approved ;)

PR #287, opened 2019-02-21
Updated idam java client to 0.0.5

nitinprabhu:

https://tools.hmcts.net/jira/browse/BPS-558

### Change description ###

- This PR uses the latest idam java client library which sends requests to idam using request body and not using requests param. 

```
[ ] Yes
[x ] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -184,7 +184,7 @@ dependencies {
   compile group: 'uk.gov.hmcts.reform', name: 'java-logging', version: versions.reformLogging
   compile group: 'uk.gov.hmcts.reform', name: 'java-logging-appinsights', version: versions.reformLogging
   compile group: 'uk.gov.hmcts.reform', name: 'service-auth-provider-client', version: '2.0.0'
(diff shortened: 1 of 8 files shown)


build.gradle
```diff
@@ -184,7 +184,7 @@ dependencies {
   compile group: 'uk.gov.hmcts.reform', name: 'java-logging', version: versions.reformLogging
   compile group: 'uk.gov.hmcts.reform', name: 'java-logging-appinsights', version: versions.reformLogging
   compile group: 'uk.gov.hmcts.reform', name: 'service-auth-provider-client', version: '2.0.0'
-  compile group: 'uk.gov.hmcts.reform', name: 'idam-client', version: '0.0.2'
+  compile group: 'uk.gov.hmcts.reform', name: 'idam-client', version: '0.0.4'
   compile group: 'uk.gov.hmcts.reform', name: 'core-case-data-store-client', version: '4.5.0'
```
doncem: Would be good to update `core-case-data-store-client` on a same go as well - it has lower version of feigns in there - jump from 9 -> 10 could potentially cause problems. Same goes for `service-auth-provider-client`

PR #295, opened 2019-02-22
Refactor 02: breakdown envelope processor

doncem:
### Change description ###

1st PR: #291 

This PR moves handler itself to new package. This way inner package classes can "communicate" package-privately as there's no need to disclose this information anyway.

In order to smoothly breakdown the single `process` method I'm adding some basic composition of `MessageProcessingResult`. It's kind of flatmap with fail fast. Fast-forwarding to the last (6th) branch the `process` holds 3 distinct steps to complete the processing. 4th one is message finaliser - keeping it separate as is now. Steps:

1. parse envelope
2. publish envelope
3. notify processor
4. finalise the message

or if you prefer more bullet points 😆:

1. process message:
1.1. parse envelope
1.2. publish envelope
1.3. notify processor
2. finalise the message

last one kind of explains the separation mentioned above

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/handler/EnvelopeEventProcessor.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/handler/EnvelopeEventProcessor.java
@@ -1,4 +1,4 @@
-package uk.gov.hmcts.reform.bulkscan.orchestrator.services;
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.handler;
 
 import com.google.common.base.Strings;
 import com.microsoft.azure.servicebus.ExceptionPhase;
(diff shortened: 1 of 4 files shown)


doncem:
Although it should have been more than enough insight of what's coming I have pushed two more PRs for your consideration. In further development additional functions introduced are actually simplified. If in need of investigations - can push fully for sneak peek review. For now:

- [2 - 3](https://github.com/hmcts/bulk-scan-orchestrator/compare/refactor/02-breakdown-envelope-processor...refactor/03-breakdown-envelope-processor?expand=1)
- [3 - 4](https://github.com/hmcts/bulk-scan-orchestrator/compare/refactor/03-breakdown-envelope-processor...refactor/04-breakdown-envelope-processor?expand=1)
- [master - 4](https://github.com/hmcts/bulk-scan-orchestrator/compare/master...refactor/04-breakdown-envelope-processor?expand=1)

lgonczar:
The overall change, which this PR is a part of, contains some bits that I'm ok with, such as extracting message finalisation into a separate class, but also elements that I don't quite agree with, such as putting logic into `MessageProcessingResult` and building some sort of a single-use DSL in it (`andThen`, `andThenWithEnvelope`). This change starts with current pull request and is extended later. At the top level of message processing we're dealing with extremely simple logic:
- parse envelope
- process the envelope

It can be expressed in two lines of code somewhere in currently existing `process` method, without having to add this extra logic that together with its tests increases the amount of code by approx 100 lines.

I also think process-related logging should take place in the processor, which "knows" what it's trying to do with the message, instead of dispersing it between processor and `MessageProcessingResult`.

As the author of most of the original code I may be biased, so I'm leaving it to other reviewers.

PR #304, opened 2019-03-04
Prevent attaching exception record to a case if it's already attached

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-539

### Change description ###

Prevent attaching exception record to a case if it's already attached to any case

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -130,4 +149,12 @@ private String createEventSummary(
             theCase.getId()
         );
     }
+
+    private boolean isExceptionRecordAttachedToCase(CaseDetails exceptionRecordDetails) {
+        Object attachToCaseReference = exceptionRecordDetails
```
rkondratowicz: what if `exceptionRecordDetails` is null?

doncem: it is not null - assured by api call in previous step

PR #315, opened 2019-03-07
Make `AbstractEventPublisher` abstract from `Envelope`

doncem:
### Change description ###

Second major revamp of `AbstractEventPublisher` following #314. This PR includes complete abstraction of event it is trying to publish. Introducing `Event` kind of object which is wrapper-alike and contains logs. Previously event publisher was tightly coupled with `Envelope` which mostly was useful(was used?) for logging reasons. The rest properties could originate from anywhere.

There is another different object which also "uses" event publisher technique and is entirely unaware of any envelopes

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/AbstractEventPublisher.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/AbstractEventPublisher.java
@@ -53,10 +53,32 @@ abstract class AbstractEventPublisher {
     }
 
     protected void publish(Envelope envelope, String caseTypeId, String eventTypeId, String eventSummary) {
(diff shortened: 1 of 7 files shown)


lgonczar:
I think the best course of action would be moving away from AbstractEventPublisher towards composition. Maybe we should have a team meeting to discuss this area, because there are already some ideas in the team.

doncem:
Since opinions are expressed I would gladly contribute it is not complicated at all

Another point would be, and it should be obvious after investigation of numbers and series of pull requests is that favourable input is not java alike input. It's okay. It is still very good and, especially nowadays, doable input

Without further ado, this PR is clearly a "good first issue". No doubt can easily transition to composition from here now. That being said, either now or in next PR

Furthermore, the course of actions being taking in these series were much more insidious and complicated. Rather clear that away and proceed with other improvements alongside. Any one can make such initiative and it is more than encouraging

Thank you for the input

doncem:
release resources, not the and no time at the moment to discuss this approach

PR #329, opened 2019-03-14
Disable dlq messages deletion scheduled job on prod

aliveni:
Disable deleting the messages from envelopes dlq on `prod`

```
[ ] Yes
[x ] No
```


lgonczar:
I wonder if we could just remove this line.

aliveni:
> I wonder if we could just remove this line.

@lgonczar we could remove it.. it doesn't make any difference. 
Once the scheduled job is tested, I'll have to change it to true

PR #332, opened 2019-03-15
Feature: manually dead letter

doncem:

[DLQ Notification](https://tools.hmcts.net/jira/browse/BPS-532)

### Change description ###

Using output of max delivery count introduced since hmcts/bulk-scan-shared-infrastructure#78.

Joining two placements of deadlettering into one method and introducing `AppInsights` custom event so that querying azure will be separate from tons of traces. Efficient and more info can be added. This is first sample of what could be pushed and enough to have as ground base information

```
[ ] Yes
[x] No
```

### DIFF
--- a/config/pmd/ruleset.xml
+++ b/config/pmd/ruleset.xml
@@ -12,8 +12,6 @@
   </rule>
   <rule ref="category/java/codestyle.xml">
     <exclude name="AtLeastOneConstructor" />
-    <exclude name="CommentDefaultAccessModifier" />
-    <exclude name="DefaultPackage" />
     <exclude name="LocalVariableCouldBeFinal" />
     <exclude name="LongVariable" />
     <exclude name="MethodArgumentCouldBeFinal" />
--- a/config/pmd/ruleset.xml
+++ b/config/pmd/ruleset.xml
@@ -12,6 +12,8 @@
   </rule>
   <rule ref="category/java/codestyle.xml">
     <exclude name="AtLeastOneConstructor" />
+    <exclude name="CommentDefaultAccessModifier" />
+    <exclude name="DefaultPackage" />
     <exclude name="LocalVariableCouldBeFinal" />
     <exclude name="LongVariable" />
     <exclude name="MethodArgumentCouldBeFinal" />
--- a/infrastructure/variables.tf
+++ b/infrastructure/variables.tf
@@ -70,3 +70,9 @@ variable "delete_envelopes_dlq_messages_cron" {
 variable "delete_envelopes_dlq_messages_ttl" {
   default = "72h"
 }
+
+variable "envelopes_queue_max_delivery_count" {
+  type = "string"
+  default = "10"
+  description = "Application message consumption limit. Will be favoured in case server value is higher"
+}
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/logging/AppInsights.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/logging/AppInsights.java
@@ -1,0 +1,32 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.logging;
+
+import com.google.common.collect.ImmutableMap;
+import com.microsoft.applicationinsights.TelemetryClient;
+import com.microsoft.azure.servicebus.IMessage;
+import org.springframework.stereotype.Component;
+import uk.gov.hmcts.reform.logging.appinsights.AbstractAppInsights;
+
+@Component
+public class AppInsights extends AbstractAppInsights {
+
+    static final String DEAD_LETTER_EVENT = "DeadLetter";
+
+    public AppInsights(TelemetryClient client) {
+        super(client);
+    }
+
+    public void trackDeadLetteredMessage(IMessage message, String queue, String reason, String description) {
+        telemetry.trackEvent(
+            DEAD_LETTER_EVENT,
+            ImmutableMap.of(
+                "reason", reason,
+                "description", description,
+                "messageId", message.getMessageId(),
+                "queue", queue
+            ),
+            ImmutableMap.of(
+                "deliveryCount", (double) (message.getDeliveryCount() + 1)
+            )
+        );
+    }
+}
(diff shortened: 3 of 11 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
```diff
@@ -126,21 +134,31 @@ private void finaliseProcessedMessage(
...
             case POTENTIALLY_RECOVERABLE_FAILURE:
-                // do nothing - let the message lock expire
-                log.info(
-                    "Allowing message with ID {} to return to queue (delivery attempt {})",
-                    message.getMessageId(),
-                    message.getDeliveryCount() + 1
-                );
+                int deliveryCount = (int) message.getDeliveryCount() + 1;
+
+                if (deliveryCount < maxDeliveryCount - 5) {
```
rkondratowicz: looks like actual max is not `maxDeliveryCount`

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
```diff
@@ -149,6 +167,18 @@ private void finaliseProcessedMessage(
...
     }
 
+    private void deadLetterTheMessage(
+        IMessage message,
+        String reason,
+        String description
+    ) throws InterruptedException, ServiceBusException {
+        messageOperations.deadLetter(message.getLockToken(), reason, description);
+
+        log.info("Message with ID {} has been dead-lettered", message.getMessageId());
```
rkondratowicz: should this not be an error?

config/pmd/ruleset.xml
```diff
@@ -12,6 +12,8 @@
   </rule>
   <rule ref="category/java/codestyle.xml">
     <exclude name="AtLeastOneConstructor" />
+    <exclude name="CommentDefaultAccessModifier" />
+    <exclude name="DefaultPackage" />
```
rkondratowicz: why?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/logging/AppInsights.java
```diff
@@ -0,0 +1,32 @@
...
+        telemetry.trackEvent(
+            DEAD_LETTER_EVENT,
+            ImmutableMap.of(
+                "reason", reason,
+                "description", description,
+                "messageId", message.getMessageId(),
+                "queue", queue
+            ),
+            ImmutableMap.of(
+                "deliveryCount", (double) (message.getDeliveryCount() + 1)
```
rkondratowicz: why +1?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
doncem: This idea was proposed initially: increase delivery count and dead letter prior that. For fairly low value like `5` application application makes sure it gets to the point. Would not like to trust azure service bus for handling situation "whether I did push X times yet or not" since 1st delivery is marked `0`. Even though we know and can inspect such behaviour in our logs.

Consider it as safety measure. Default configuration is 10 deliveries so it'll die out sooner. The bigger the count - the more attempts. Hence the increase which is linked in the description.

If we're happy with `deliveryCount < maxDeliveryCount` instead - let's agree in conversation among the team ;)

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
doncem: Was info before - the line has just been moved

config/pmd/ruleset.xml
doncem: Since usage of newer libraries, there is no need to provide `public` accessors for certain methods like `Test` cases or `setUp` methods. PMD is not aware of that and complains in vain

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/logging/AppInsights.java
doncem: Library delivers messages from `0`. Counter is increased once lock is expired and message can be re-consumed

config/pmd/ruleset.xml
rkondratowicz: So only in tests then?
Perhaps this can be addressed across the project separately from this use story, after it's discussed?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/logging/AppInsights.java
rkondratowicz: Maybe it be named something else than `deliveryCount` if it's not message's actual delivery count?

doncem: It is delivery count. We use same wording and value in other logs too. It is rationally correct: when we read the logs, when we will read the events

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
```diff
@@ -126,21 +134,31 @@ private void finaliseProcessedMessage(
...
+                    log.info(
+                        "Allowing message with ID {} to return to queue (delivery attempt {})",
+                        message.getMessageId(),
+                        deliveryCount
+                    );
+                } else {
+                    deadLetterTheMessage(
+                        message,
+                        "Too many deliveries",
+                        "Breached the limit of message delivery count of " + deliveryCount
```
rkondratowicz: was it supposed to be 'Reached'?

infrastructure/variables.tf
```diff
@@ -70,3 +70,9 @@ variable "delete_envelopes_dlq_messages_cron" {
 variable "delete_envelopes_dlq_messages_ttl" {
   default = "72h"
 }
+
+variable "envelopes_queue_max_delivery_count" {
```
lgonczar: Maybe this value should just be calculated as the max set at queue level minus some constant, e.g. 5. Otherwise it may be easier to misconfigure it.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
doncem: It wasn't. But can amend

infrastructure/variables.tf
doncem: It is dealt in `main.tf` - whichever is the smallest value  is considered as max delivery count

PR #334, opened 2019-03-15
Fix/fix dlq scheduler

aliveni:

### Change description ###
Added the missing `@Component` annotation on Dlq Scheduler task.
Added an integration test to make sure the Dlq scheduler task is running 
according to the specified interval in the config.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/config/IntegrationTestInitializer.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/config/IntegrationTestInitializer.java
@@ -25,8 +25,6 @@ import static org.springframework.util.SocketUtils.findAvailableTcpPort;
 @Profile({"integration", "nosb"}) // no servicebus queue handler registration
 class IntegrationTestInitializer implements ApplicationContextInitializer<ConfigurableApplicationContext> {
 
-    private IMessageReceiver messageReceiver = mock(IMessageReceiver.class);
-
     @Override
     public void initialize(ConfigurableApplicationContext applicationContext) {
         System.setProperty("wiremock.port", Integer.toString(findAvailableTcpPort()));
@@ -64,11 +62,7 @@ class IntegrationTestInitializer implements ApplicationContextInitializer<Config
 
     @Bean
     public Supplier<IMessageReceiver> dlqReceiverProvider() {
-        return new Supplier<IMessageReceiver>() {
-            @Override
-            public IMessageReceiver get() {
-                return messageReceiver;
-            }
-        };
+        IMessageReceiver messageReceiver = mock(IMessageReceiver.class);
+        return () -> messageReceiver;
     }
 }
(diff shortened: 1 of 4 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/config/IntegrationTestInitializer.java
```diff
@@ -23,6 +26,8 @@
 @Profile({"integration", "nosb"}) // no servicebus queue handler registration
 class IntegrationTestInitializer implements ApplicationContextInitializer<ConfigurableApplicationContext> {
 
+    private IMessageReceiver messageReceiver = Mockito.mock(IMessageReceiver.class);
```
doncem: This can be moved to actual bean

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/config/IntegrationTestInitializer.java
```diff
@@ -57,4 +62,14 @@ public void deadLetter(UUID lockToken, String reason, String description) {
...
     }
+
+    @Bean
+    public Supplier<IMessageReceiver> dlqReceiverProvider() {
+        return new Supplier<IMessageReceiver>() {
+            @Override
+            public IMessageReceiver get() {
+                return messageReceiver;
+            }
+        };
```
doncem: This can be simplified to `return () -> messageReceiver;`

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTaskTest.java
```diff
@@ -0,0 +1,48 @@
...
+import uk.gov.hmcts.reform.bulkscan.orchestrator.config.IntegrationTest;
+
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+import static org.awaitility.Awaitility.await;
+import static org.mockito.Mockito.verify;
+
+@RunWith(SpringRunner.class)
```
doncem: We should try to convert everything to Junit5. Most of the tests are already using it

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTaskTest.java
```diff
@@ -0,0 +1,48 @@
...
+    Supplier<IMessageReceiver> dlqReceiverProvider;
+
+    private CleanupEnvelopesDlqTask dlqTask;
+
+    @Value("${scheduling.task.delete-envelopes-dlq-messages.ttl}")
+    private Duration ttl;
+
+    @Before
+    public void setUp() {
+        dlqTask = new CleanupEnvelopesDlqTask(dlqReceiverProvider, ttl);
```
doncem: There is no need to keep task in the test. It will be registered in application and schedule will kick off whenever it's ready so the test below will still validate

src/test/resources/application-integration.yaml
```diff
@@ -13,3 +13,10 @@ idam:
     bulkscan:
       username: bulkscanorchestrator+systemupdate@gmail.com
       password: Password12
+
+scheduling:
+  task:
+    delete-envelopes-dlq-messages:
+      enabled: true
+      cron: "*/10 * * * * *"
```
doncem: 1s is enough

aliveni: This is to run every 10 seconds

doncem: yea, but for integration tests we don't really need to wait for 10 seconds. 1 should be enough

PR #346, opened 2019-03-19
Feature: manually dead letter

doncem:
### Change description ###

Reference: #332 

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -60,7 +60,7 @@ locals {
     S2S_SECRET  = "${data.azurerm_key_vault_secret.s2s_secret.value}"
 
     ENVELOPES_QUEUE_CONNECTION_STRING           = "${data.terraform_remote_state.shared_infra.envelopes_queue_primary_listen_connection_string}"
-    ENVELOPES_QUEUE_MAX_DELIVERY_COUNT          = "${min(data.terraform_remote_state.shared_infra.envelopes_queue_max_delivery_count, var.envelopes_queue_max_delivery_count)}"
+    ENVELOPES_QUEUE_MAX_DELIVERY_COUNT          = "${data.terraform_remote_state.shared_infra.envelopes_queue_max_delivery_count}"
     PROCESSED_ENVELOPES_QUEUE_CONNECTION_STRING = "${data.terraform_remote_state.shared_infra.processed_envelopes_queue_primary_send_connection_string}"
 
     IDAM_API_URL              = "${var.idam_api_url}"
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -1,5 +1,5 @@
 provider "azurerm" {
-  version = "1.23.0"
+  version = "=1.22.1"
 }
 
 # Make sure the resource group exists
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -60,7 +60,7 @@ locals {
     S2S_SECRET  = "${data.azurerm_key_vault_secret.s2s_secret.value}"
 
     ENVELOPES_QUEUE_CONNECTION_STRING           = "${data.terraform_remote_state.shared_infra.envelopes_queue_primary_listen_connection_string}"
-    ENVELOPES_QUEUE_MAX_DELIVERY_COUNT          = "${data.terraform_remote_state.shared_infra.envelopes_queue_max_delivery_count}"
+    ENVELOPES_QUEUE_MAX_DELIVERY_COUNT          = "${data.terraform_remote_state.shared_infra.envelopes_queue_max_delivery_count - 5}"
     PROCESSED_ENVELOPES_QUEUE_CONNECTION_STRING = "${data.terraform_remote_state.shared_infra.processed_envelopes_queue_primary_send_connection_string}"
 
     IDAM_API_URL              = "${var.idam_api_url}"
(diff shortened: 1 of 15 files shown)


infrastructure/main.tf
```diff
@@ -60,6 +60,7 @@ locals {
     S2S_SECRET  = "${data.azurerm_key_vault_secret.s2s_secret.value}"
 
     ENVELOPES_QUEUE_CONNECTION_STRING           = "${data.terraform_remote_state.shared_infra.envelopes_queue_primary_listen_connection_string}"
+    ENVELOPES_QUEUE_MAX_DELIVERY_COUNT          = "${min(data.terraform_remote_state.shared_infra.envelopes_queue_max_delivery_count, var.envelopes_queue_max_delivery_count)}"
```
lgonczar: I think it would be ideal to have this property based on the queue setting, i.e. make it the queue max delivery count minus some small constant, e.g. 3. We always want the setting in the application to be smaller than that in the queue (to make sure it's the app that dead-letters the message) and it will be easiest to manage it this way. That's what you had before, except it would be calculated in terraform instead of application code.

src/main/resources/application.yaml
```diff
@@ -20,6 +20,7 @@ azure:
     envelopes:
       connection-string: ${ENVELOPES_QUEUE_CONNECTION_STRING}
       queue-name: envelopes
+      max-delivery-count: ${ENVELOPES_QUEUE_MAX_DELIVERY_COUNT:10}
```
lgonczar: It would be best to have no default value, so that if we forget to configure it in any environment, the app won't even start there.

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessorTest.java
```diff
@@ -158,12 +172,18 @@ public void should_dead_letter_the_message_when_notification_sending_fails() thr
             eq(DEAD_LETTER_REASON_PROCESSING_ERROR),
             eq(exceptionMessage)
         );
+        verify(appInsights).trackDeadLetteredMessage(
+            someMessage,
+            "envelopes",
```
lgonczar: Worth referencing the queue name through a constant. Just to consider, up to you.

doncem:
Those codacy alerts are a bit too much. I've introduced a suppression commit but as per request - reverted to be reviewed separately

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/logging/AppInsights.java
```diff
@@ -0,0 +1,32 @@
...
+public class AppInsights extends AbstractAppInsights {
+
+    static final String DEAD_LETTER_EVENT = "DeadLetter";
+
+    public AppInsights(TelemetryClient client) {
+        super(client);
+    }
+
+    public void trackDeadLetteredMessage(IMessage message, String queue, String reason, String description) {
+        telemetry.trackEvent(
```
nitinprabhu: could we add envelope id and zip file name please ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
```diff
@@ -126,21 +134,31 @@ private void finaliseProcessedMessage(
...
-                log.info("Message with ID {} has been dead-lettered", message.getMessageId());
                 break;
             case POTENTIALLY_RECOVERABLE_FAILURE:
-                // do nothing - let the message lock expire
-                log.info(
-                    "Allowing message with ID {} to return to queue (delivery attempt {})",
-                    message.getMessageId(),
-                    message.getDeliveryCount() + 1
-                );
+                int deliveryCount = (int) message.getDeliveryCount() + 1;
```
nitinprabhu: I assume we are adding 1 as the count starts from 0 comment might be worth here

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/logging/AppInsights.java
```diff
@@ -0,0 +1,32 @@
...
+            DEAD_LETTER_EVENT,
+            ImmutableMap.of(
+                "reason", reason,
+                "description", description,
+                "messageId", message.getMessageId(),
+                "queue", queue
+            ),
+            ImmutableMap.of(
+                "deliveryCount", (double) (message.getDeliveryCount() + 1)
```
nitinprabhu: I assume we are adding 1 as the count starts from 0 comment might be worth here

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
```diff
@@ -149,6 +167,18 @@ private void finaliseProcessedMessage(
...
     }
 
+    private void deadLetterTheMessage(
+        IMessage message,
+        String reason,
+        String description
+    ) throws InterruptedException, ServiceBusException {
+        messageOperations.deadLetter(message.getLockToken(), reason, description);
+
+        log.info("Message with ID {} has been dead-lettered", message.getMessageId());
```
nitinprabhu: can we also include zip file name and jurisdiction here ?

doncem: This would require changes to `MessageProcessingResult` and would be beneficial indeed. But rather as a separate PR.

P.S. This would've be so much simpler and already ready to use if stream of `refactor/(\d){2}-breakdown-envelope-processor` would've gone through. Just saying

lgonczar: This additional information isn't crucial here - we can get all that from logs.

PR #348, opened 2019-03-22
BPS-538 Updated aat config for dlq scheduler 

aliveni:


infrastructure/aat.tfvars
```diff
@@ -4,3 +4,5 @@ idam_client_redirect_uri = "https://rpe-bulk-scan-processor-sandbox.service.core
 supported_jurisdictions = ["SSCS", "BULKSCAN", "DIVORCE", "PROBATE"]
 
 delete_envelopes_dlq_messages_enabled = "true"
+delete_envelopes_dlq_messages_cron = "0 0 * * * *"
+delete_envelopes_dlq_messages_ttl = "1h"
```
nitinprabhu: we can keep it very low like say 1 min

aliveni: @nitinprabhu I'll set the ttl value to 5 minutes.

PR #353, opened 2019-03-25
BPS-538 Added functional test for Dlq Cleanup task

aliveni:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/BPS-538

### Change description ###
Added functional test to verify if Dlq scheduler task is deleting the messages.

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[ ] No
```

PR #355, opened 2019-03-26
Updated dlq scheduler to log messageId for invalid messages

aliveni:
### Change description ###
When there is an invalid message in envelopes dlq, log the message id and delete the message.
Updated aat config to run the dlq scheduler every miniute and the ttl is set to 30 seconds (for functional test which is added in https://github.com/hmcts/bulk-scan-orchestrator/pull/353)

```
[ ] Yes
[x ] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -194,7 +194,6 @@ dependencies {
   
   compile group: 'com.microsoft.azure', name: 'azure-servicebus', version: '1.2.8'
   compile group: 'io.vavr', name: 'vavr', version: '0.10.0'
-  compile group: 'org.json', name: 'json', version: '20180813'
 
   annotationProcessor group: 'org.springframework.boot', name: 'spring-boot-configuration-processor'
 
--- a/build.gradle
+++ b/build.gradle
@@ -194,6 +194,7 @@ dependencies {
   
   compile group: 'com.microsoft.azure', name: 'azure-servicebus', version: '1.2.8'
   compile group: 'io.vavr', name: 'vavr', version: '0.10.0'
+  compile group: 'org.json', name: 'json', version: '20180813'
 
   annotationProcessor group: 'org.springframework.boot', name: 'spring-boot-configuration-processor'
 
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
@@ -83,7 +83,7 @@ public class CleanupEnvelopesDlqTask {
                 envelope.caseRef
             );
         } catch (InvalidMessageException e) {
-            // Not logging the InvalidMessageException as it prints the sensitive information from the envelope
+            // Not logging the exception as it prints the sensitive information from the envelope
             log.error("An error occurred while parsing the dlq message with Message Id: {}",
                 msg.getMessageId());
         }
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
@@ -49,7 +49,8 @@ public class CleanupEnvelopesDlqTask {
             IMessage message = messageReceiver.receive();
             while (message != null) {
                 if (canBeDeleted(message)) {
-                    deleteMessage(messageReceiver, message);
+                    logMessage(message);
+                    messageReceiver.completeAsync(message.getLockToken());
                 }
                 message = messageReceiver.receive();
             }
@@ -66,25 +67,23 @@ public class CleanupEnvelopesDlqTask {
         }
     }
 
-    private void deleteMessage(IMessageReceiver messageReceiver, IMessage message)
-        throws ServiceBusException, InterruptedException {
+    private void logMessage(IMessage msg) {
         try {
-            Envelope envelope = EnvelopeParser.parse(message.getBody());
+            Envelope envelope = EnvelopeParser.parse(msg.getBody());
 
             log.info(
                 "Deleting message ID: {}, Envelope ID: {}, File name: {}, Jurisdiction: {},"
                     + " Classification: {}, Case: {}",
-                message.getMessageId(),
+                msg.getMessageId(),
                 envelope.id,
                 envelope.zipFileName,
                 envelope.jurisdiction,
                 envelope.classification,
                 envelope.caseRef
             );
-            messageReceiver.complete(message.getLockToken());
         } catch (InvalidMessageException e) {
             log.error("An error occurred while parsing the dlq message with Message Id: {}",
-                message.getMessageId());
+                msg.getMessageId());
         }
     }
 
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
@@ -50,10 +50,11 @@ public class CleanupEnvelopesDlqTask {
             while (message != null) {
                 if (canBeDeleted(message)) {
                     logMessage(message);
-                    messageReceiver.completeAsync(message.getLockToken());
+                    messageReceiver.complete(message.getLockToken());
                 }
                 message = messageReceiver.receive();
             }
+
         } catch (ConnectionException e) {
             log.error("Unable to connect to envelopes dead letter queue", e);
         } finally {
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
@@ -3,24 +3,25 @@ package uk.gov.hmcts.reform.bulkscan.orchestrator.tasks;
 import com.microsoft.azure.servicebus.IMessage;
 import com.microsoft.azure.servicebus.IMessageReceiver;
 import com.microsoft.azure.servicebus.primitives.ServiceBusException;
-import org.json.JSONException;
-import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.scheduling.annotation.Scheduled;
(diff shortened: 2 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
```diff
@@ -48,12 +48,11 @@ public void deleteMessagesInEnvelopesDlq() throws ServiceBusException, Interrupt
...
             while (message != null) {
                 if (canBeDeleted(message)) {
-                    logMessage(message);
-                    messageReceiver.complete(message.getLockToken());
+                    deleteMessage(messageReceiver, message);
                 }
                 message = messageReceiver.receive();
             }
-
+            log.info("Finished processing messages in envelopes Dead letter queue.");
```
lgonczar: This is not what this PR is about and it can be done later, but it would be nice to have the number of deleted messages in the log.

lgonczar:
Why don't we want to delete invalid messages? They may contain sensitive data, too.

build.gradle
```diff
@@ -194,6 +194,7 @@ dependencies {
   
   compile group: 'com.microsoft.azure', name: 'azure-servicebus', version: '1.2.8'
   compile group: 'io.vavr', name: 'vavr', version: '0.10.0'
+  compile group: 'org.json', name: 'json', version: '20180813'
```
doncem: Could we double check it's not clashing with anything similar already present in the scope?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
```diff
@@ -67,18 +66,25 @@ public void deleteMessagesInEnvelopesDlq() throws ServiceBusException, Interrupt
...
+            JSONObject msgContent = new JSONObject(new String(message.getBody(), defaultCharset()));
+            log.info(
+                "Deleting message. ID: {} "
+                    + "Envelope ID: {}, File name: {}, Jurisdiction: {}, Classification: {}, Case: {}",
+                message.getMessageId(),
+                msgContent.get("id"),
+                msgContent.get("zip_file_name"),
+                msgContent.get("jurisdiction"),
+                msgContent.get("classification"),
+                msgContent.get("case_ref")
```
doncem: i presume it's `null` if not found?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
```diff
@@ -67,18 +66,25 @@ public void deleteMessagesInEnvelopesDlq() throws ServiceBusException, Interrupt
...
+            );
+            messageReceiver.complete(message.getLockToken());
```
doncem: Just a thought, perhaps better would be to `completeAsync` and then compose the log message, what do you think?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
```diff
@@ -3,24 +3,24 @@
...
 import java.time.Duration;
 import java.time.Instant;
 import java.util.function.Supplier;
 
+import static java.nio.charset.Charset.defaultCharset;
+
 /**
  * Deletes messages from envelopes Dead letter queue.
  */
-@Component
```
doncem: So it is not needed anymore? 🤔

lgonczar: Good spot - without the annotation the schedule won't work.

build.gradle
aliveni: I'm removing this dependency and leave the log message as it was before (using `EnvelopeParser`).

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
aliveni: yes, it would be null if not found. But I'm reverting this change as I mentioned above

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
aliveni: yeah, good spot! I didn't notice that I removed this annotation 😳

aliveni:
> Why don't we want to delete invalid messages? They may contain sensitive data, too.

I discussed with @nitinprabhu about how we handle invalid messages. He suggested that we have manually handle invalid messages. we can discuss tomorrow

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
aliveni: @lgonczar removed the log message. I'll add it in a separate PR and will include the number of messages deleted

build.gradle
doncem: so can we remove it?

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTaskTest.java
```diff
@@ -120,4 +125,26 @@ public void should_not_process_messages_when_connection_exception_is_thrown() {
...
+        cleanupDlqTask.deleteMessagesInEnvelopesDlq();
+
+        //then
+        assertThat(capture.toString())
+            .containsPattern(
+                ".+ ERROR \\[.*\\].*"
+                    + CleanupEnvelopesDlqTask.class.getSimpleName()
+                    + " An error occurred while parsing the dlq message with Message Id: " + messageId
+            );
+    }
```
doncem: let's hope this test works as logging library is v4 😳

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
```diff
@@ -49,11 +50,10 @@ public void deleteMessagesInEnvelopesDlq() throws ServiceBusException, Interrupt
             while (message != null) {
                 if (canBeDeleted(message)) {
                     logMessage(message);
-                    messageReceiver.complete(message.getLockToken());
+                    messageReceiver.completeAsync(message.getLockToken());
```
lgonczar: Why async? I think it's best to leave `complete`, have all the exceptions thrown if something goes wrong and not start a new message until the previous one has been dealt with.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
```diff
@@ -49,11 +50,10 @@ public void deleteMessagesInEnvelopesDlq() throws ServiceBusException, Interrupt
...
                 }
                 message = messageReceiver.receive();
             }
-
         } catch (ConnectionException e) {
```
lgonczar: Not a part of this PR, but we should catch and log all exceptions.

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTaskTest.java
```diff
@@ -120,4 +125,26 @@ public void should_not_process_messages_when_connection_exception_is_thrown() {
         assertThat(exception).isNull();
     }
 
+    @Test
+    public void should_log_messageId_for_invalid_messages() throws Exception {
```
lgonczar: As it just verifies log output, this test is optional.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
aliveni: @lgonczar we are catching the `ServiceBusException` and `InterruptedException` in `DlqReceiverProvider` class (from where we are getting the message reciever for dlq)

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTaskTest.java
aliveni: I've deleted the test.

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTaskTest.java
aliveni: @doncem removed the test

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
aliveni: @lgonczar Changed it back to `complete`

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
lgonczar: It's still worth having some generic error logging, in case a runtime exception is thrown. Not necessarily in this PR.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTask.java
```diff
@@ -68,17 +69,23 @@ public void deleteMessagesInEnvelopesDlq() throws ServiceBusException, Interrupt
...
+                    + " Classification: {}, Case: {}",
+                msg.getMessageId(),
+                envelope.id,
+                envelope.zipFileName,
+                envelope.jurisdiction,
+                envelope.classification,
+                envelope.caseRef
+            );
+        } catch (InvalidMessageException e) {
+            log.error("An error occurred while parsing the dlq message with Message Id: {}",
```
lgonczar: It would be good to add `e` as the last argument.

PR #356, opened 2019-03-27
Refactor EventPublisherContainer to process envelopes in CCD

lgonczar:
### Change description ###

Refactor EventPublisherContainer to process envelopes in CCD.

Before the change, it returned classification-specific functions, which were executed by the caller immediately after. This change also takes case retrieval off EnvelopeEventProcessor's hands, reducing the number of its dependencies.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/CcdUpdater.java
```diff
@@ -0,0 +1,50 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.events;
+
+import com.google.common.base.Strings;
+import org.springframework.stereotype.Service;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.CaseRetriever;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.model.Envelope;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+
+@Service
+public class CcdUpdater {
```
lgonczar: This used to be `EventPublisherContainer`. Code structure is pretty much the same, but now, instead of returning functions that update CCD, it updates CCD itself.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/CcdUpdater.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private final AttachDocsToSupplementaryEvidence attachDocsPublisher;
+    private final CreateExceptionRecord exceptionRecordCreator;
+    private final CaseRetriever caseRetriever;
+
+    public CcdUpdater(
```
nitinprabhu: This name needs changing CaseHandler or something similar might be better because this class handles update as well as create

lgonczar: After quick discussion we agreed on `EnvelopeHandler`

lgonczar:
Applied the suggestion. Needs to be re-approved.

PR #365, opened 2019-04-03
Handle user errors separately when attaching exception record to a case

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-491

- add new exceptions to distinguish from internal errors
- log user errors using 'warning' level (to prevent alerts)
- in case of an internal error, send a fixed generic message back to CCD


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -111,7 +116,7 @@ private void verifyExceptionRecordIsNotAttachedToCase(
         );
 
         if (isExceptionRecordAttachedToCase(fetchedExceptionRecord)) {
-            throw new CallbackException("Exception record is already attached to a case");
+            throw new AlreadyAttachedToCaseException("Exception record is already attached to a case");
```
rkondratowicz: I'll improve this message by adding the case ref on a separate PR

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -28,6 +28,8 @@
 @Service
 public class AttachCaseCallbackService {
 
+    public static final String INTERNAL_ERROR_MSG = "An error occurred";
```
lgonczar: Because this message will be shown to the caseworker, maybe we should make it more descriptive, to make it clear it's the server error and maybe that it happened when trying to assign the exception record to the case.

PR #369, opened 2019-04-04
Add extra info to error

rkondratowicz:
When user tries to attach exception record to a case when it's already attached
provide the ref number of that case.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
@@ -115,8 +115,10 @@ public class AttachCaseCallbackService {
             exceptionRecordJurisdiction
         );
 
-        if (isExceptionRecordAttachedToCase(fetchedExceptionRecord)) {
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -115,8 +115,10 @@ private void verifyExceptionRecordIsNotAttachedToCase(
             exceptionRecordJurisdiction
         );
 
-        if (isExceptionRecordAttachedToCase(fetchedExceptionRecord)) {
-            throw new AlreadyAttachedToCaseException("Exception record is already attached to a case");
+        Object attachToCaseRef = fetchedExceptionRecord.getData().get(ATTACH_CASE_REFERENCE_FIELD_NAME);
```
nitinprabhu: should we typecast this to string ?

rkondratowicz: I guess. Just moved the existing code from lines 158 - 164 so that the change is minimal

PR #370, opened 2019-04-04
Update dlq config in demo to run the dlq scheduler every minute

aliveni:

### Change description ###
Update dlq config in demo to run the dlq scheduler every minute

```
[ ] Yes
[ x] No
```


aliveni:
> We could also disable the job in integration tests

As discussed earlier, I'll disable the scheduler in integration test if we could test the dlq scheduler in functional test

PR #376, opened 2019-04-06
Combine dependencies caught by @dependabot

doncem:
### Change description ###

Combine #373 and #374 

```
[ ] Yes
[x] No
```


doncem:
static tests all pass and this change only relates to that. Bogus owasp step is faulty

PR #383, opened 2019-04-08
Create a task for consuming envelopes queue

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-563

### Change description ###

Create a task for consuming envelopes queue. The task only does it when no jurisdiction account is locked in IDAM.

This is a part of a bigger change that also makes EnvelopeEventProcessor responsible for reading messages from the queue. The complete intended change can be found here: https://github.com/hmcts/bulk-scan-orchestrator/compare/feature/introduce-circuit-breaker

The task is not a service, yet, so this change has no effect on how the application works (needs `@Service` annotation to be added).

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
@@ -51,6 +51,16 @@ public class EnvelopeEventProcessor implements IMessageHandler {
         this.appInsights = appInsights;
     }
 
+    /**
+     * Reads and processes next message from the queue.
+     *
+     * @return false if there was no message to process. Otherwise true.
+     */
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
```diff
@@ -51,6 +51,16 @@ public EnvelopeEventProcessor(
...
     }
 
+    /**
+     * Reads and processes next message from the queue.
+     *
+     * @return false if there was no message to process. Otherwise true.
+     */
+    public boolean processNextMessage() throws ServiceBusException, InterruptedException {
+        // TODO: implement
+        return false;
```
lgonczar: Implementation has been extracted (https://github.com/hmcts/bulk-scan-orchestrator/compare/feature/introduce-circuit-breaker), as this is a big enough change alone.

PR #385, opened 2019-04-09
Prevent processing the queue when a jurisdiction account is locked in IDAM

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-563

### Change description ###

Prevent processing the queue when a jurisdiction account is locked in IDAM.

The most important part of this change is `EnvelopeEventProcessor` becoming responsible for reading messages, as opposed to having them passed from the caller. The old way didn't allow for applying circuit breaker, as there was no way to pause receiving messages.

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/tasks/CleanupEnvelopesDlqTaskTest.java
```diff
@@ -14,7 +14,7 @@
 
 import java.time.Duration;
 import java.time.LocalDateTime;
-import java.time.ZoneOffset;
+import java.util.TimeZone;
 import java.util.TimeZone;
```
rkondratowicz: duplicated?

lgonczar:
> Would be nice if possible to split this PR into smaller chunks. I think there's no harm in disabling message handler since we are not pushing to master. So if you could prepare necessary changes _just_ for handler first, then enable with attaching `@Service` to the task with extra tests if needed. And old code (especially config part) can be removed as third/cleanup PR. This could help us focus better

This PR is pretty much just the handler. Enabling the service is two lines of code - import and annotation. Can't do this bit in a later PR, as that will mean the orchestrator will not be processing the queue anymore. I can extract the deletion of no-longer-used code (which is most of the changes), but I don't think it would be a great split, as there shouldn't be unused code at any time.

PR #394, opened 2019-04-12
Include IDAM response in jurisdiction-specific IDAM status

lgonczar:
### Change description ###

Include IDAM response in jurisdiction-specific IDAM status. Currently, the status contains a generic FeignException message.

```
[ ] Yes
[x] No
```


lgonczar:
> Do we need to update smoke tests as it uses it ?

No need - smoke tests expect success and this message is only for when IDAM returns errors.

PR #396, opened 2019-04-15
Include IDAM response in jurisdiction-specific IDAM status

lgonczar:
### Change description ###

This PR is a copy of already approved https://github.com/hmcts/bulk-scan-orchestrator/pull/394, which was closed because of not being able to delete namespace, which in turn caused builds to fail.

Include IDAM response in jurisdiction-specific IDAM status. Currently, the status contains a generic FeignException message.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/AuthenticationCheckerTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/AuthenticationCheckerTest.java
@@ -94,7 +94,7 @@ class AuthenticationCheckerTest {
         assertThat(status.jurisdiction).isEqualTo(LOCKED_ACCOUNT_JURISDICTION);
         assertThat(status.isCorrect).isFalse();
         assertThat(status.errorResponseStatus).isEqualTo(HttpStatus.LOCKED.value());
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/AuthenticationCheckerTest.java
```diff
@@ -156,7 +156,7 @@ private FeignException createFeignException(int httpStatus) {
             .errorStatus("method1", Response
                 .builder()
                 .request(mock(Request.class))
-                .body(new byte[0])
+                .body("Error response from IDAM".getBytes())
```
rkondratowicz: where is 'Error response from IDAM' coming from?

lgonczar: It's just an example response. I'm giving it some value here to check if whatever's here ends up in the response of the IDAM status endpoint.

rkondratowicz: oops, it's setup stage, thought it was verification

PR #401, opened 2019-04-16
Updated dlq provider supplier to throw exception 

aliveni:

### Change description ###
The `DlqReceiverrProvider` is returning null in case `InterruptedException`. 
Updated `DlqReceiverrProvider` to throw an exception to avoid `NullPointerException`

```
[ ] Yes
[x ] No
```


rkondratowicz:
More context would be appreciated

aliveni:
> More context would be appreciated

@rkondratowicz Updated the decription

PR #404, opened 2019-04-16
Add dlq environment variables to aat values template file

aliveni:


aliveni:
Combined these changes with pr-402 as they both should be in the same pr

PR #411, opened 2019-04-23
Include library class/interface methods

doncem:
### Change description ###

Theoretically, these classes should be recorded to AppInsights. Let's see

```
[ ] Yes
[x] No
```


rkondratowicz:
please provide clear commit message when merging

doncem:
on the progress updating servicebus library. could be latest will be automatically tracked by app insights. otherwise will review more internet

PR #424, opened 2019-05-08
Update build tools

doncem:
### Change description ###

Main change is junit5 jump 5.3.2 -> 5.4.2 as dependencies/configuration differ. Just follow send letter setup as it has latest.

The rest is minor tool upgrades.

Left out checkstyle as for some reason 8.19 -> 8.20 was not trivial jump and not lingering for this one

```
[ ] Yes
[x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -120,11 +120,11 @@ sonarqube {
   }
 }
 
-dependencyUpdates.resolutionStrategy = {
+dependencyUpdates.resolutionStrategy {
   componentSelection { rules ->
     rules.all { ComponentSelection selection ->
-      boolean rejected = ['alpha', 'beta', 'rc', 'cr', 'm'].any { qualifier ->
-        selection.candidate.version ==~ /(?i).*[.-]${qualifier}[.\d-]*/
+      boolean rejected = ['alpha', 'beta', 'rc', 'cr', 'm', 'preview', 'b', 'ea'].any { qualifier ->
+        selection.candidate.version ==~ /(?i).*[.-]$qualifier[.\d-+]*/
       }
       if (rejected) {
         selection.reject('Release candidate')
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -216,33 +223,38 @@ dependencies {
...
   smokeTestCompile sourceSets.test.runtimeClasspath
+
+  smokeTestImplementation group: 'org.junit.jupiter', name: 'junit-jupiter', version: versions.junit
 }
 
 mainClassName = 'uk.gov.hmcts.reform.bulkscan.orchestrator.Application'
 
 bootJar {
-  archiveName = 'bulk-scan-orchestrator.jar'
+  getArchiveFileName().set(provider {
```
nitinprabhu: do we need to set it like this ? Previous one looks simpler.

build.gradle
```diff
@@ -107,6 +107,13 @@ jacocoTestReport {
     html.destination file("${buildDir}/reports/jacoco/html")
     xml.destination file("${project.buildDir}/reports/jacoco/test/jacocoTestReport.xml")
   }
+  afterEvaluate {
+    getClassDirectories().from = getClassDirectories().files.collect {
```
nitinprabhu: why we need this for ?

build.gradle
doncem: usage of `archiveName` has been deprecated and planned to be removed in next major jump. the set is copied from what it was/is doing at the moment with `archiveName = 'some.jar'`

build.gradle
doncem: just to be at the same page with sonar config which skips that folder

nitinprabhu: we have it in sonar to exclude code coverage so do we need it in pmd ?

doncem: not pmd, jacoco ;)

nitinprabhu: we have this `property "sonar.coverage.exclusions", "**/config/**"`  configured when sonar will use jacoco reports it will exclude it anyways so I think we don't need the config added in this PR. What are your thoughts ?

doncem: hm, true. on the other hand, jacoco report is used by other CIs too: codacy and codecov which both are configured for this project

PR #433, opened 2019-05-20
Create configuration bits for mapping services to case type IDs and jurisdictions

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-584

### Change description ###

Create configuration section for mapping services to jurisdictions and case type IDs. This config is necessary for finding cases by their legacy IDs and it's a part of that change.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -46,7 +46,7 @@ idam:
       password: ${IDAM_USERS_BULKSCAN_PASSWORD:Password12}
 
 service-config:
-  # Services and all their case type IDs.
+  # Services and all their case type IDs, excluding exception records.
   # These mappings are used by Bulk Scan to determine which case type IDs to search across
   # when looking for a case by its legacy ID. Those CTIDs that are not present here, will
   # not be included in the search.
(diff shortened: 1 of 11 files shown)


src/main/resources/application.yaml
```diff
@@ -45,6 +45,43 @@ idam:
       username: ${IDAM_USERS_BULKSCAN_USERNAME:bulkscan+ccd@gmail.com}
       password: ${IDAM_USERS_BULKSCAN_PASSWORD:Password12}
 
+service-config:
```
lgonczar: Service teams will be requested to complete their lists of case type IDs.

src/main/resources/application.yaml
```diff
@@ -45,6 +45,43 @@ idam:
...
+    - service: bulkscan # service name corresponds to the name of the container in Blob Storage
+      jurisdiction: BULKSCAN
+      case-type-ids:
+        - Bulk_Scanned
+        - BULKSCAN_ExceptionRecord
+
+    - service: sscs
+      jurisdiction: SSCS
+      case-type-ids:
+        - SSCS_ExceptionRecord
```
nitinprabhu: I think we should not configure exception record case types ? We should only have service case types as we are attaching scanned documents to service cases.

lgonczar: Of course. They won't even have legacy IDs, anyway. Will remove.

PR #438, opened 2019-05-24
Create validation methods required for attaching case by legacy ID

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-584

### Change description ###

Depending on which callback endpoint is used, the case will be validated by either:
- `attachToCaseReference` - the old one, which only contains CCD case references and currently serves a search field and a reference field
- `searchCaseReference` + `searchCaseReferenceType` - fields used for search purposes only. For this kind of search we also need to retrieve the service name from case type ID

This PR makes all the changes on the validation side for the above searches to be possible.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
@@ -6,7 +6,9 @@ import org.slf4j.LoggerFactory;
 import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
 
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
+import java.util.Optional;
 import javax.annotation.Nonnull;
 
 import static io.vavr.control.Validation.invalid;
@@ -14,6 +16,9 @@ import static io.vavr.control.Validation.valid;
 import static java.lang.String.format;
 
 final class CallbackValidations {
+
+    private static final String CASE_TYPE_ID_SUFFIX = "_ExceptionRecord";
+
     private static final Logger log = LoggerFactory.getLogger(CallbackValidations.class);
 
     private static final CaseReferenceValidator caseRefValidator = new CaseReferenceValidator();
@@ -45,8 +50,18 @@ final class CallbackValidations {
     }
 
     @Nonnull
-    static Validation<String, String> hasCaseReference(CaseDetails theCase) {
-        return caseRefValidator.validate(theCase);
+    static Validation<String, String> hasSearchCaseReferenceType(CaseDetails theCase) {
+        return caseRefValidator.validateCaseReferenceType(theCase);
+    }
+
+    @Nonnull
+    static Validation<String, String> hasSearchCaseReference(CaseDetails theCase) {
+        return caseRefValidator.validateSearchCaseReference(theCase);
+    }
+
+    @Nonnull
+    static Validation<String, String> hasAttachToCaseReference(CaseDetails theCase) {
+        return caseRefValidator.validateAttachToCaseReference(theCase);
     }
 
     @Nonnull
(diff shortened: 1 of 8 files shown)


src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidationsTest.java
```diff
@@ -14,38 +14,125 @@
...
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.TestCaseBuilder.caseWithCcdSearchCaseReference;
 import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.TestCaseBuilder.caseWithDocument;
-import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.TestCaseBuilder.caseWithReference;
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.TestCaseBuilder.caseWithExternalSearchCaseReference;
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.TestCaseBuilder.caseWithSearchCaseReference;
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.TestCaseBuilder.caseWithSearchCaseReferenceType;
 import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.TestCaseBuilder.createCaseWith;
 import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.TestCaseBuilder.document;
 
+@SuppressWarnings("checkstyle:LineLength")
```
lgonczar: Some test cases take more than 120 characters. I came to the conclusion that it's better to suppress line length warning than split each case into multiple lines. Alternatively, I could split this test class by method and then multiline representations of each case would probably be fine. What do you think?

lgonczar:
Those codacy issues are just the result of me following the existing convention in those files.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CaseReferenceValidator.java
```diff
@@ -4,40 +4,97 @@
...
+                    }
+                }
+            ).orElseGet(() -> invalid("No case reference type supplied"));
+    }
+
+    private Validation<String, String> validateCcdCaseRef(Object caseRef) {
+        return Optional.of(caseRef)
+            .filter(ref -> ref instanceof String)
+            .map(ref -> ((String) ref).replaceAll("[^0-9]", ""))
+            .filter(ref -> !Strings.isNullOrEmpty(ref))
```
rkondratowicz: I know it's not new code, but... is it me, or is this validation not only unclear but also wrong?
Perhaps worth adding a todo comment and/or jira ticket.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/JurisdictionToUserMapping.java
```diff
@@ -27,7 +28,7 @@ public void setUsers(Map<String, Map<String, String>> users) {
     }
 
     private Entry<String, Credential> createEntry(Entry<String, Map<String, String>> entry) {
-        String key = entry.getKey().toLowerCase();
+        String key = entry.getKey().toLowerCase(Locale.getDefault());
```
rkondratowicz: unrelated, but ok

lgonczar: codacy complained

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -57,6 +72,28 @@ private CallbackValidations() {
...
+    @Nonnull
+    static Validation<String, String> hasServiceNameInCaseTypeId(CaseDetails theCase) {
+        return Optional
+            .ofNullable(theCase)
+            .map(CaseDetails::getCaseTypeId)
+            .filter(caseTypeId -> caseTypeId != null)
+            .map(caseTypeId -> {
+                if (caseTypeId.endsWith(CASE_TYPE_ID_SUFFIX)) {
+                    int serviceNameLength = caseTypeId.length() - CASE_TYPE_ID_SUFFIX.length();
+                    String serviceName = caseTypeId.substring(0, serviceNameLength).toLowerCase(Locale.getDefault());
```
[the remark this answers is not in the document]
lgonczar: It wouldn't be exactly the same as what we've got, but I guess we can rule out the possibility of the suffix appearing anywhere in the service's name.  Will apply this, as it's easier to read.

PR #450, opened 2019-06-03
Fix the test which is failing due to timestamp serialisation

aliveni:
https://tools.hmcts.net/jira/browse/BPS-647

### Change description ###
Fix the test which is failing intermittently due to the timestamp format mismatch.

```
[ ] Yes
[x ] No
```


src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/EnvelopeParserTest.java
```diff
@@ -235,7 +234,7 @@ private JSONObject toJson(Document doc) throws Exception {
             .put("control_number", doc.controlNumber)
             .put("type", doc.type)
             .put("subtype", doc.subtype)
-            .put("scanned_at", toIso8601(doc.scannedAt))
+            .put("scanned_at", doc.scannedAt)
```
rkondratowicz: I think this needs to stay so that we can actually check whether mapping is working as expected

aliveni: The test should've failed when the timestamp format is changed. but it doesn't. Also, we are not converting the date fields in the envelope (`deliveryDate`, `openingDate`). I'm not sure why only scannedDate fields are converted to ISO8601 format

rkondratowicz: The test checks whether `EnvelopeParser` can parse json string to an object of type `Envelope`.

`toIso8601(...)` returns a string, so that we exactly know what we put in a json that we want to parse (this could also be a .json file read from resources).

The test fails when it's comparing the date from `Envelope` object (created after parsing the json string) to a pre-existing `private Instant scannedAt = Instant.now();`

For some reason `scannedAt` on Travis has 6 decimal places for seconds.

If anything needs changing in this test it's the assertion, not building the json string to parse.

rkondratowicz: Or maybe instead of using current timestamp
```
private Instant scannedAt = Instant.now();
```
we could build the instant 'manually':
```
private Instant scannedAt = LocalDateTime.of(..., ...).toInstant(...);
```

rkondratowicz: Perhaps it's because Travis uses Java 11:
```
$ java -Xmx32m -version
openjdk version "11.0.2" 2019-01-15
```

aliveni: > Perhaps it's because Travis uses Java 11:
> 
> ```
> $ java -Xmx32m -version
> openjdk version "11.0.2" 2019-01-15
> ```

@rkondratowicz I don't see java11 in Travis build. I see java 8 in travis logs for this PR
https://travis-ci.com/hmcts/bulk-scan-orchestrator/builds/114113178
```
$ java -Xmx32m -version
java version "1.8.0_151"
```

aliveni: > Or maybe instead of using current timestamp
> 
> ```
> private Instant scannedAt = Instant.now();
> ```
> 
> we could build the instant 'manually':
> 
> ```
> private Instant scannedAt = LocalDateTime.of(..., ...).toInstant(...);
> ```

I'll try this 👍

aliveni: Travis is using jdk 11 now. I'm sure it was jdk8 in yesterday's build. 
I'll update travis config to force using jdk8

PR #457, opened 2019-06-05
Replace a potentially existing test email with a non-existing one

lgonczar:
### Change description ###

Replace a potentially existing test email with a non-existing one

```
[ ] Yes
[x] No
```


lgonczar:
Merging, as the only change here is in a unit test, which is passing. We've reached the namespaces limit.

PR #462, opened 2019-06-10
Add previous service case reference to Envelope

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-588

### Change description ###

Add previous service case reference to Envelope.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/model/Envelope.java
```diff
@@ -24,6 +25,7 @@
     public Envelope(
         @JsonProperty(value = "id", required = true) String id,
         @JsonProperty(value = "case_ref", required = true) String caseRef,
+        @JsonProperty("previous_service_case_ref") String previousServiceCaseRef,
```
rkondratowicz: `required=false` for clarity?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/model/Envelope.java
```diff
@@ -11,6 +11,7 @@
 
     public final String id;
     public final String caseRef;
+    public final String previousServiceCaseRef;
```
rkondratowicz: might be good to add a note here clarifying that it's not the 'previous ServiceCaseRef', it's the 'previous service caseRef'.

rkondratowicz: or rename the field altogether

lgonczar: This "previous service case reference" is a convention that is shared with Exela and CCD, so I'd rather not rename it. I'll add a comment.

lgonczar: Maybe I should call it legacyId, as that's how we normally refer to it?

lgonczar: `legacyCaseRef` it is

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/model/Envelope.java
lgonczar: Idea complains about redundant parameter if I do

rkondratowicz: people > intellij ;)

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/model/Envelope.java
```diff
@@ -24,6 +25,7 @@
     public Envelope(
         @JsonProperty(value = "id", required = true) String id,
         @JsonProperty(value = "case_ref", required = true) String caseRef,
+        @JsonProperty("previous_service_case_ref") String legacyCaseRef,
```
rkondratowicz: are there any changes made in the processor yet?
If not, maybe we could have `legacy_case_ref` in json?

lgonczar: The processor is already passing this field. I'll change the name as the last thing in this story.

PR #474, opened 2019-06-19
Replace CaseRetriever with CaseFinder and CcdApi

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-588

### Change description ###

`CaseRetriever` used to duplicate some of `CcdApi`'s functionality. This PR removes that class. The calls to it have either been replaced with calls to `CcdApi` or, in case of attaching supplementary evidence, with a call to `CaseFinder`. As a result, supplementary evidence can now be attached by either CCD ID or legacy ID (if CCD ID is either missing or the search doesn't find a case by it).

Also, add a missing integration test related with searching for a case by legacy ID.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CaseFinder.java
```diff
@@ -114,6 +114,10 @@ private void logCaseRetrievalResultBasedOnLegacyIdSearch(
         try {
             return Optional.of(ccdApi.getCase(ccdCaseRef, jurisdiction));
         } catch (CaseNotFoundException e) {
+            log.info("Case wasn't found by CCD ID {}", ccdCaseRef);
+            return Optional.empty();
+        } catch (InvalidCaseIdException e) {
```
nitinprabhu: Can we please add a test case for this ?

lgonczar: done

PR #531, opened 2019-08-20
Bump servicebus 1.2.14 -> 1.2.15

doncem:
### Change description ###

[Change "log"](https://github.com/Azure/azure-service-bus-java/compare/1.2.14...1.2.15)

Fixes possible connection loss after being idle

```
[ ] Yes
[x] No
```


doncem:
This version bump is bugfix. Not sure api version needs a change or ever needed in this repository. Unless you are referring to secrets? But shouldn't be the case either way.

P.S. Same version jump can be applied in processor too - without worry ;)

PR #532, opened 2019-08-21
Improve logging around processing envelopes queue

lgonczar:
### Change description ###

Improve logging around processing envelopes queue

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
```diff
@@ -64,8 +64,11 @@ public boolean processNextMessage() throws ServiceBusException, InterruptedExcep
         IMessage message = messageReceiver.receive();
 
         if (message != null) {
+            log.info("Started processing message with ID {}", message.getMessageId());
```
rkondratowicz: this exact thing is logged inside of `process()`

lgonczar: Interesting. I moved this `log.info` here, not just added it. Must have been a merge issue resulting with the removal of line deletion in `process()`. I'll add it again.

PR #550, opened 2019-09-02
Configure concurrent jobs

rkondratowicz:
### Summary
By default, spring boot scheduler uses single thread and only one job can run at a time.
This change configures a thread pool which will allow running multiple jobs at the same time.

### Background
We currently have 2 jobs:
- `CleanupEnvelopesDlqTask`
- `EnvelopesQueueConsumeTask`

Due to a bug, the dlq cleanup job kept running for a long time, preventing the envelopes job from starting.

https://stackoverflow.com/a/35017579
https://crmepham.github.io/spring-boot-multi-thread-scheduling/

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/config/SchedulerConfig.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/config/SchedulerConfig.java
@@ -1,0 +1,21 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.config;
+
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.annotation.SchedulingConfigurer;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/config/SchedulerConfig.java
```diff
@@ -0,0 +1,21 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.config;
+
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.annotation.SchedulingConfigurer;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
+import org.springframework.scheduling.config.ScheduledTaskRegistrar;
+
+@Configuration
+public class SchedulerConfig implements SchedulingConfigurer {
```
doncem: `@EnableScheduling` should go together with `SchedulingConfigurer` implementation 😉

PR #551, opened 2019-09-04
Populate CCD exception record with OCR data validation warnings

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-757

### Change description ###

Populate CCD exception record with OCR data validation warnings

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/ExceptionRecord.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/ExceptionRecord.java
@@ -26,6 +26,9 @@ public class ExceptionRecord implements CaseData {
 
     public final List<CcdCollectionElement<String>> ocrDataValidationWarnings;
 
+    // Yes/No field indicating if there are warnings to show
+    public final String hasWarnings;
+
     public ExceptionRecord(
         String classification,
         String poBox,
@@ -34,7 +37,8 @@ public class ExceptionRecord implements CaseData {
         LocalDateTime openingDate,
         List<CcdCollectionElement<ScannedDocument>> scannedDocuments,
         List<CcdCollectionElement<CcdKeyValue>> ocrData,
-        List<CcdCollectionElement<String>> ocrDataValidationWarnings
+        List<CcdCollectionElement<String>> ocrDataValidationWarnings,
+        String hasWarnings
     ) {
         this.classification = classification;
         this.poBox = poBox;
@@ -44,5 +48,6 @@ public class ExceptionRecord implements CaseData {
         this.scannedDocuments = scannedDocuments;
         this.ocrData = ocrData;
         this.ocrDataValidationWarnings = ocrDataValidationWarnings;
+        this.hasWarnings = hasWarnings;
     }
 }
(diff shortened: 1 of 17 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/model/Envelope.java
```diff
@@ -21,7 +21,7 @@
     public final Classification classification;
     public final List<Document> documents;
     public final List<OcrDataField> ocrData;
-    public final List<String> ocrValidationWarnings;
+    public final List<String> ocrDataValidationWarnings;
```
rkondratowicz: why? now it doesn't match the json property

lgonczar: The field in the queue message is called `ocr_validation_warnings`, which is a bit against our naming convention for OCR data. I'll change the element name to `ocr_data_validation_warnings`, but that requires a change in processor and can be done separately, in steps.

rkondratowicz: I'd suggest sticking to populating CCD exception record with validation warnings in this PR

lgonczar: I don't understand what you mean

rkondratowicz: revert this change

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/ExceptionRecord.java
```diff
@@ -24,14 +24,21 @@
...
         String classification,
         String poBox,
         String jurisdiction,
         LocalDateTime deliveryDate,
         LocalDateTime openingDate,
         List<CcdCollectionElement<ScannedDocument>> scannedDocuments,
-        List<CcdCollectionElement<CcdKeyValue>> ocrData
+        List<CcdCollectionElement<CcdKeyValue>> ocrData,
+        List<CcdCollectionElement<String>> ocrDataValidationWarnings,
+        String hasWarnings
```
rkondratowicz: This can lead to creating ER with warning and this field being set to "No".
Or even "Hello".
Perhaps `hasWarnings` could be a getter?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/model/Envelope.java
lgonczar: I'll make this change in a previous PR. Otherwise, I'd create more inconsistencies.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/ExceptionRecord.java
lgonczar: Tests prevent this scenario. I made it this way so that the model would remain without logic.

nitinprabhu: should we rename hasWarnings to warningsIndicator or something similar as it is not a boolean type ?

lgonczar: We should treat it like a boolean, though. This is a YesOrNo field in CCD, which corresponds to boolean. Unfortunately, CCD doesn't support boolean in its normal form.

nitinprabhu: 🤦‍♂

doncem:
If so, make it clear by labelling and make sure definitions updates are not kept back for who knows how long. E.g. my non-breaking changes never got picked up as far as I am aware

PR #552, opened 2019-09-04
Introduce new validator for event id

doncem:

[Create Case: pre-validation of callback request](https://tools.hmcts.net/jira/browse/BPS-738)

### Change description ###

Requirement to validate new event id screams for some unification together with existing one. Introducing separate validator just for event id. Event ID is the most important instant rejection required validation so makes sense to extract it out to be stand-alone

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EventIdValidator.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EventIdValidator.java
@@ -20,12 +20,12 @@ public final class EventIdValidator {
 
     @Nonnull
     static Validation<String, Void> isAttachToCaseEvent(String eventId) {
-        return hasValidEventId(EVENT_ID_ATTACH_TO_CASE::equalsIgnoreCase, eventId);
+        return hasValidEventId(EVENT_ID_ATTACH_TO_CASE::equals, eventId);
     }
 
     @Nonnull
     static Validation<String, Void> isCreateCaseEvent(String eventId) {
-        return hasValidEventId(EVENT_ID_CREATE_CASE::equalsIgnoreCase, eventId);
+        return hasValidEventId(EVENT_ID_CREATE_CASE::equals, eventId);
     }
 
     @Nonnull
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EventIdValidator.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EventIdValidator.java
@@ -1,0 +1,37 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd;
+
+import io.vavr.control.Validation;
+
+import java.util.function.Function;
+import javax.annotation.Nonnull;
+
+import static io.vavr.control.Validation.invalid;
+import static io.vavr.control.Validation.valid;
+import static java.lang.String.format;
+
+public final class EventIdValidator {
+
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EventIdValidator.java
```diff
@@ -0,0 +1,37 @@
...
+        return hasValidEventId(EVENT_ID_ATTACH_TO_CASE::equalsIgnoreCase, eventId);
+    }
+
+    @Nonnull
+    static Validation<String, Void> isCreateCaseEvent(String eventId) {
+        return hasValidEventId(EVENT_ID_CREATE_CASE::equalsIgnoreCase, eventId);
+    }
+
+    @Nonnull
+    private static Validation<String, Void> hasValidEventId(Function<String, Boolean> isValid, String eventId) {
```
rkondratowicz: 'valid' or 'expected'?
```
isCreateCaseEvent(EVENT_ID_ATTACH_TO_CASE)`
```
will say `EVENT_ID_ATTACH_TO_CASE` is not supported. But it very much is.

When will this check be used in general?
Perhaps include usage in the PR

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EventIdValidator.java
```diff
@@ -0,0 +1,37 @@
...
+    private static final String EVENT_ID_ATTACH_TO_CASE = "attachToExistingCase";
+    private static final String EVENT_ID_CREATE_CASE = "createCase";
+
+    private EventIdValidator() {
+        // utility class constructor
+    }
+
+    @Nonnull
+    static Validation<String, Void> isAttachToCaseEvent(String eventId) {
+        return hasValidEventId(EVENT_ID_ATTACH_TO_CASE::equalsIgnoreCase, eventId);
```
nitinprabhu: can't ignore case for this check :)

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EventIdValidator.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @Nonnull
+    static Validation<String, Void> isCreateCaseEvent(String eventId) {
+        return hasValidEventId(EVENT_ID_CREATE_CASE::equalsIgnoreCase, eventId);
```
nitinprabhu: can't ignore case for this check :)

nitinprabhu: I know it exists currently but it looks like a bug

doncem: lol, breaking change compared to previous implementation. at least for attach to case event

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EventIdValidator.java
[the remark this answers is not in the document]
nitinprabhu: Do we need a function as an argument here ?

doncem: This is private method. And in source code. So no questions about `isCreateCase(AttachCaseEvent)` integrity possible here nor `expected` - keyword is used in test cases

PR #553, opened 2019-09-04
Replace current use of event id validator with new one

doncem:

[Create Case: pre-validation of callback request](https://tools.hmcts.net/jira/browse/BPS-738)

### Change description ###

Replace validation of `attachToExistingCase` event id validation to new validator. `createCase` event validation is coming in next PR when new service for new callback is introduced

```
[ ] Yes
[x] No
```


doncem:
Strange pipeline error for the time being. Closing to clear resources. Will just re-create as a separate PR

doncem:
Retrying again. Apparently pipeline was a bit slow

PR #565, opened 2019-09-11
Disable scheduled tasks in functional tests

lgonczar:
### Change description ###

Disable scheduled tasks in functional tests, to that they don't affect tests and don't produce logs.

```
[ ] Yes
[x] No
```


rkondratowicz:
why are functional starting the app in the first place? 😱

lgonczar:
> why are functional test starting the app in the first place? 😱

This requires a whole separate discussion.

PR #569, opened 2019-09-12
Feature: create case callback endpoint

doncem:

### Change description ###

Introduce `/callback/create-case` endpoint to handle CCD callbacks. Semi implemented - no happy path as only portion of service is implemented. Hence `NotImplementedException` is thrown

```
[ ] Yes
[x] No
```


doncem:
postponing that in (personal) favour of #572 - will continue work from there. this will not go away just going to be introduced a bit later

PR #573, opened 2019-09-13
Require transformation URL to be configured

doncem:

### Change description ###

Require transformation URL to be configured prior doing any `CaseDetails` validation. This is fail fast operation and expanding previously (#572) extracted component to accept any number of "fail fast" operations. Not using stream api as it might as well evaluate all suppliers.

Such expansion as well proves the mentioning of it in previous PR - and there are more to come.

Lot of branches introduced hence so many changes here. Main goal: introduce mandatory requirement of callback: service in question has to be configured with transformation url in orchestrator

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
@@ -1,25 +1,34 @@
 package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd;
 
+import com.google.common.base.Strings;
 import io.vavr.collection.Seq;
 import io.vavr.control.Either;
+import io.vavr.control.Try;
+import io.vavr.control.Validation;
 import org.springframework.stereotype.Service;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.client.transformation.model.request.ExceptionRecord;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.config.ServiceConfigItem;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.callback.CreateCaseValidator;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.config.ServiceConfigProvider;
 import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
 
 import java.util.List;
 
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.CallbackValidations.hasServiceNameInCaseTypeId;
 import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.EventIdValidator.isCreateCaseEvent;
 
 @Service
 public class CreateCaseCallbackService {
 
     private final CreateCaseValidator validator;
+    private final ServiceConfigProvider serviceConfigProvider;
 
     public CreateCaseCallbackService(
-        CreateCaseValidator validator
+        CreateCaseValidator validator,
+        ServiceConfigProvider serviceConfigProvider
     ) {
         this.validator = validator;
+        this.serviceConfigProvider = serviceConfigProvider;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackServiceTest.java
```diff
@@ -15,30 +22,110 @@
...
-        Either<List<String>, ExceptionRecord> output = SERVICE.process(null, "some event");
+        Either<List<String>, ExceptionRecord> output = service.process(null, "some event");
 
         assertThat(output.isLeft()).isTrue();
         assertThat(output.getLeft()).containsOnly("The some event event is not supported. Please contact service team");
+        verify(serviceConfigProvider, never()).getConfig(anyString());
+    }
+
+    @Test
+    void should_not_allow_to_process_callback_in_case_wrong_case_type_id_is_missing() {
```
aliveni: in_case_<s>**wrong_**</s>case_type_id_is_missing?

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackServiceTest.java
```diff
@@ -15,30 +22,110 @@
...
+        // when
+        Either<List<String>, ExceptionRecord> output = service.process(caseDetails, EVENT_ID);
+
+        assertThat(output.isLeft()).isTrue();
+        assertThat(output.getLeft()).containsOnly("No case type ID supplied");
+        verify(serviceConfigProvider, never()).getConfig(anyString());
+    }
+
+    @Test
+    void should_not_allow_to_process_callback_in_case_wrong_case_type_id_is_empty() {
```
aliveni: same as above

aliveni: do we need another test for wrong case type id?

doncem: "wrong case type id" is covered with both: empty case type id and the unit test which asserts "Transformation URL is not configured" message.

any string is firstly asserting the correct suffix and then returns any service (as a prefix of case type) found. then it's down to service config provider to determine service is configured or not. so both - incorrect case type id and transformation url branches are covered

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackServiceTest.java
```diff
@@ -194,4 +291,10 @@ void should_report_errors_when_ocr_data_is_invalid() {
             .asString()
             .matches(match);
     }
+
+    private void setUpTransformationUrl() {
+        ServiceConfigItem configItem = new ServiceConfigItem();
+        configItem.setTransformationUrl("url");
```
nitinprabhu: Doesn't matter but it would be nice to have at least correct url format `http://servicehost/transform-exception-record`

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -28,12 +37,28 @@ public CreateCaseCallbackService(
...
+
+    private Either<List<String>, Void> assertAllowToAccess(CaseDetails caseDetails, String eventId) {
+        return validator.mandatoryPrerequisites(
+            () -> isCreateCaseEvent(eventId),
+            () -> getServiceConfig(caseDetails).map(item -> null)
+        );
+    }
+
+    private Validation<String, ServiceConfigItem> getServiceConfig(CaseDetails caseDetails) {
+        return hasServiceNameInCaseTypeId(caseDetails).flatMap(service -> Try
```
nitinprabhu: we should be retrieving transformation url based on jurisdiction and not service name as CCD case detail object would know about jurisdiction and not service(internal to orchestrator)

```
 - service: sscs
      jurisdiction: SSCS

```

nitinprabhu: I am aware you can still use service name but when it can be simply done using jurisdiction look up then we should use it :)

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackServiceTest.java
doncem: yes, it absolutely doesn't matter

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
doncem: and what about finrem service? jurisdiction is divorce

nitinprabhu: Good point so we can’t do it per jurisdiction.

PR #576, opened 2019-09-17
Add transformation client to `createCase` callback service

doncem:

### Change description ###

Just a placement to call transformation client. This PR solves one TBD - returns a real object to caller which does not exists at the moment.

Next: add s2s

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
@@ -8,6 +8,7 @@ import io.vavr.control.Either;
 import io.vavr.control.Try;
 import io.vavr.control.Validation;
 import org.springframework.stereotype.Service;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.client.transformation.TransformationClient;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.client.transformation.model.request.ExceptionRecord;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.config.ServiceConfigItem;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.callback.CreateCaseValidator;
@@ -27,13 +28,16 @@ public class CreateCaseCallbackService {
 
     private final CreateCaseValidator validator;
     private final ServiceConfigProvider serviceConfigProvider;
+    private final TransformationClient transformationClient;
 
     public CreateCaseCallbackService(
         CreateCaseValidator validator,
-        ServiceConfigProvider serviceConfigProvider
+        ServiceConfigProvider serviceConfigProvider,
+        TransformationClient transformationClient
     ) {
         this.validator = validator;
         this.serviceConfigProvider = serviceConfigProvider;
+        this.transformationClient = transformationClient;
     }
 
     /**
@@ -48,6 +52,7 @@ public class CreateCaseCallbackService {
                 .combine(getServiceConfig(caseDetails).mapError(Array::of))
                 .ap(this::createExceptionRecord)
                 .mapError(errors -> errors.flatMap(Function.identity()))
+                .flatMap(Function.identity())
                 .toEither()
                 .mapLeft(Seq::asJava)
             );
@@ -70,7 +75,22 @@ public class CreateCaseCallbackService {
             .getOrElse(Validation.invalid("Transformation URL is not configured"));
     }
 
-    private Map<String, Object> createExceptionRecord(ExceptionRecord exceptionRecord, ServiceConfigItem configItem) {
-        return ImmutableMap.of("caseReference", UUID.randomUUID());
+    private Validation<Seq<String>, Map<String, Object>> createExceptionRecord(
+        ExceptionRecord exceptionRecord,
+        ServiceConfigItem configItem
+    ) {
+        try {
+            // log
+            transformationClient.transformExceptionRecord(
+                configItem.getTransformationUrl(),
+                exceptionRecord,
+                "s2s token"
+            );
+
+            return Validation.valid(ImmutableMap.of("caseReference", UUID.randomUUID()));
+        } catch (Exception exception) {
+            // log
+            return Validation.invalid(Array.of("Internal error. " + exception.getMessage()));
+        }
     }
 }
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
@@ -10,6 +10,7 @@ import io.vavr.control.Validation;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.client.transformation.InvalidCaseDataException;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.client.transformation.TransformationClient;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.client.transformation.model.request.ExceptionRecord;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.config.ServiceConfigItem;
@@ -102,6 +103,9 @@ public class CreateCaseCallbackService {
             );
 
             return Validation.valid(ImmutableMap.of("caseReference", UUID.randomUUID()));
+        } catch (InvalidCaseDataException exception) {
+            // let controller deal with this. it is 422 or 400
+            throw exception;
         } catch (Exception exception) {
             log.error(
                 "Failed to create exception for service {} and exception record {}",
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
@@ -54,7 +54,7 @@ public class CreateCaseCallbackService {
             .flatMap(theVoid -> validator
                 .getValidation(caseDetails)
                 .combine(getServiceConfig(caseDetails).mapError(Array::of))
-                .ap((exceptionRecord, configItem) -> createExceptionRecord(
+                .ap((exceptionRecord, configItem) -> createNewCase(
                     exceptionRecord,
                     configItem,
                     caseDetails.getId()
@@ -83,7 +83,7 @@ public class CreateCaseCallbackService {
             .getOrElse(Validation.invalid("Transformation URL is not configured"));
     }
 
-    private Validation<Seq<String>, Map<String, Object>> createExceptionRecord(
+    private Validation<Seq<String>, Map<String, Object>> createNewCase(
         ExceptionRecord exceptionRecord,
         ServiceConfigItem configItem,
         long caseId
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
@@ -54,7 +54,11 @@ public class CreateCaseCallbackService {
             .flatMap(theVoid -> validator
                 .getValidation(caseDetails)
                 .combine(getServiceConfig(caseDetails).mapError(Array::of))
-                .ap(this::createExceptionRecord)
+                .ap((exceptionRecord, configItem) -> createExceptionRecord(
+                    exceptionRecord,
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -61,4 +82,26 @@ public CreateCaseCallbackService(
             .filter(item -> !Strings.isNullOrEmpty(item.getTransformationUrl()))
             .getOrElse(Validation.invalid("Transformation URL is not configured"));
     }
+
+    private Validation<Seq<String>, Map<String, Object>> createExceptionRecord(
```
aliveni: is this method supposed to just transform the exception record to case format? The method name is confusing, creating an Exception record from ExceptionRecord?

doncem: fair point. will think of a better name unless provided earlier. since this PR is preview of what's coming and therefore based from the one before - #573

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -61,4 +82,26 @@ public CreateCaseCallbackService(
...
+        try {
+            log.info("Start creating exception record for {} {}", configItem.getService(), caseId);
+
+            transformationClient.transformExceptionRecord(
+                configItem.getTransformationUrl(),
+                exceptionRecord,
+                "s2s token"
+            );
+
+            return Validation.valid(ImmutableMap.of("caseReference", UUID.randomUUID()));
```
aliveni: I assume this will change in the later prs?

doncem: Yep, as soon as CCD API is included - the actual value is stuck in. Now just keeping the format so actually from this point endpoint can be developed. Quick note - will change the method signature to include idam token and user id coming from callback request. Not yet needed - not including

apereverzin: A better name maybe should be validateExceptionRecordTransformation

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
apereverzin: A better name maybe should be validateExceptionRecordTransformation

doncem: this method will wrap the whole thing:

transform, create case, return new case id

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -61,4 +82,26 @@ public CreateCaseCallbackService(
...
             .getOrElse(Validation.invalid("Transformation URL is not configured"));
     }
+
+    private Validation<Seq<String>, Map<String, Object>> createNewCase(
+        ExceptionRecord exceptionRecord,
+        ServiceConfigItem configItem,
+        long caseId
+    ) {
+        try {
+            log.info("Start creating exception record for {} {}", configItem.getService(), caseId);
```
lgonczar: Worth adding `service` and `case ID` words here and in the error log to make the message obvious.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -61,4 +82,26 @@ public CreateCaseCallbackService(
...
+                configItem.getTransformationUrl(),
+                exceptionRecord,
+                "s2s token"
+            );
+
+            return Validation.valid(ImmutableMap.of("caseReference", UUID.randomUUID()));
+        } catch (Exception exception) {
+            log.error("Failed to create exception for {} {}", configItem.getService(), caseId, exception);
+
+            return Validation.invalid(Array.of("Internal error. " + exception.getMessage()));
```
lgonczar: There are two possible error scenarios:
- 422 from transformation API - we should return a 200 response with those errors, so that they can be displayed to the caseworker
- any other error (5xx, timeouts, our internal errors), including 400 - the orchestrator should return a 500 response and CCD will display a generic error

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
lgonczar: It would be best to get that dependency (CCD API stuff) in first.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
doncem: Good catch. Transformation client already handles 422 and 400 altogether. Just about to extract to separate catch block and we can proceed from there as I'm not entirely sure whether re-throw or return some internal. But that would be similar to `Exception` handle

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
doncem: not really. ccd api dependency needs what's coming out from transformation api

lgonczar: What I meant was that it's best to start with the deepest dependencies and move towards the root rather than the other way. The benefit is that whatever pieces of code we write, they're complete from the beginning and there's no need to "fix" them later (of course refactoring happens all the time). They can also be properly tested from the start.

This is kind of a general suggestion - I don't want to stop this piece of work if it's problematic to add that other thing first at this point.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -61,4 +83,38 @@ public CreateCaseCallbackService(
...
+            );
+
+            transformationClient.transformExceptionRecord(
+                configItem.getTransformationUrl(),
+                exceptionRecord,
+                "s2s token"
+            );
+
+            return Validation.valid(ImmutableMap.of("caseReference", UUID.randomUUID()));
+        } catch (InvalidCaseDataException exception) {
```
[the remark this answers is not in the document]
lgonczar: At this point 422 and 400 responses are indistinguishable from each other. Whatever handles this exception needs to have that information, because a 422 response will result in us returning a 200 response to CCD, with errors and warnings (presented to caseworker), while 400 means there's an error at service level (orchestrator can't properly talk to the transformation service) and it's going to be a 500 for CCD.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -61,4 +83,38 @@ public CreateCaseCallbackService(
             .filter(item -> !Strings.isNullOrEmpty(item.getTransformationUrl()))
             .getOrElse(Validation.invalid("Transformation URL is not configured"));
     }
+
+    private Validation<Seq<String>, Map<String, Object>> createNewCase(
```
lgonczar: What does the sequence represent in the validation? Is it errors or warnings?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
doncem: Then I guess blocking this ticket so validation can go in first was incorrect order, isn't it? From entry point of this ticket - validation was already present. Now going backwards ccd would need s2s and response from transformation api. Could, of course, just provide nulls or dummy things..

At this point it shouldn't really matter. No serious implications in code changes, service not in use, humanly you review a step-by-step approach.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
doncem: As per current design orchestrator is using - it is always responding with errors. Warnings are not used. Transformation will return warnings on successful response so this would mean successful response + warnings. slightly different behaviour. Errors + warnings never used nor distinguished. Nor adapted in orchestrator callback at any point in the past

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -1,45 +1,67 @@
...
     }
 
     /**
      * Create case record from exception case record.
      *
-     * @return Either TBD - not yet implemented
+     * @return Either list of errors or map of changes - new case reference
      */
-    public Either<List<String>, ExceptionRecord> process(CaseDetails caseDetails, String eventId) {
+    public Either<List<String>, Map<String, Object>> process(CaseDetails caseDetails, String eventId) {
```
lgonczar: Why is it not going to be an `ExceptionRecord` anymore? Something will still have to prepare it, based on the result from transformation API. This class seems to be the best place for it.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
lgonczar: Transformation API can return a combination of errors and warnings so that they can be presented to the caseworker as such. We can't ignore warnings.

Note that not all calls to our create case API will result in a new case - it depends on `ignoreWarnings` flag. On the first attempt, if there are warnings, we want to present them to the caseworker and let them decide what to do next.

Errors and warnings have been working together in attaching exception record to a case.

doncem: > Errors and warnings have been working together in attaching exception record to a case.

I don't see warnings being there:
https://github.com/hmcts/bulk-scan-orchestrator/blob/0ec25b7f8601f7d36e197d6e7884642fbe6f3ab4/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CcdCallbackController.java#L42-L57

it's always an error

transformation api either throws an exception or returns case details with warnings. that is / should be handled in next pr. or perhaps include in the mapping - at least that's what ccd api branch currently does

lgonczar: Regarding attaching exception records, you're right - in practice we don't provide values for warnings here. I thought we did at some point, remembering that we tested whether we could see both at the same time, but looks like I was wrong about adding any warnings. But it doesn't matter - we're building a new endpoint, which has its own requirements.

> transformation api either throws an exception or returns case details with warnings. that is / should be handled in next pr. or perhaps include in the mapping - at least that's what ccd api branch currently does

This is what our java client might be doing at the moment, but transformation API returns either case details with warnings or errors with warnings, as per spec: https://tools.hmcts.net/confluence/pages/viewpage.action?pageId=1064666568#TechnicalprerequisitesandinformationforServiceon-boardingwithPhase2Bulkscanning-2.4.5OpenAPI(swagger)specificationforexceptionrecordtransformationAPI

We need to reflect that in our responses to CCD.

doncem: ok. Can I suggest return object TBD in the future then and leave this as is? I do have warnings accompanied later on which I was meant to discuss with the team when reaching such point. But I think this counts as a discussion and I'm aware of future development now

lgonczar: Up to you. My recommendation, given it's a new method and we know that it won't look like this when the story is complete, is to prepare dependencies first:
- adjustment of transformation client, so that we can distinguish between 422/400
- preparation of the part where we talk to CCD to create the case

Then you could easily complete this one without having to stub things and leaving it for further modifications.

doncem:
Everything discussed. Merging as is

PR #579, opened 2019-09-20
Add S2S dependency to case creation service

doncem:

### Change description ###

Previous PR #576 misses S2S intentionally. This makes a very tiny change in the chain

Next: ~add ccd api. Big change~ add endpoint and fix service return object as per discussion originated in #576

```
[ ] Yes
[x] No
```


aliveni:
oops, yeah I think we need the config

doncem:
Not really. We do not validate token, we need to generate one for transformation API. And the config is already in and used in all interactions with CCD

PR #577, opened 2019-09-19
Include envelope ID in exception records

lgonczar:
### JIRA link (if applicable) ###

https://tools.hmcts.net/jira/browse/BPS-490

### Change description ###

Include envelope ID in exception records, but only if the envelope's jurisdiction supports duplicate exception record prevention (i.e. is listed as such in the config)

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #581, opened 2019-09-20
Include envelope ID in exception records

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-490

### Change description ###

**This is a copy of already approved #577, plus config fix**

Include envelope ID in exception records, but only if the envelope's jurisdiction supports duplicate exception record prevention (i.e. is listed as such in the config)

```
[ ] Yes
[x] No
```


lgonczar:
done

PR #582, opened 2019-09-20
Added checks for classification and OCRs

apereverzin:
BPS-746: https://tools.hmcts.net/jira/browse/BPS-746

### Change description ###
Updated validation of case creation:
- classification cannot be supplementary_evidence
- OCR data is not mandatory for new_application and is mandatory for exception

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -203,17 +234,10 @@ static boolean hasOcr(CaseDetails theCase) {
...
     }
 
-    /**
-     * Used in createCase callback only. Can be reviewed later and with BPS-746 - revised the usage.
-     * @param theCase from CCD.
-     * @return Validation of it
-     */
     @SuppressWarnings("unchecked")
-    public static Validation<Object, List<Map<String, Object>>> getOcrData(CaseDetails theCase) {
+    public static Optional<List<Map<String, Object>>> getOcrData(CaseDetails theCase) {
```
lgonczar: Doesn't the change of this method affect the validation? As far as I understand from the code, if OCR data is missing then `getOcrDataFields` will return `valid(null)`, instead of `invalid("Missing OCR data")`.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -192,9 +192,40 @@ static boolean hasOcr(CaseDetails theCase) {
             .map(validation -> validation
                 .mapError(throwable -> "Invalid journeyClassification. Error: " + throwable.getMessage())
             )
+            .map(validation -> validateClassification(validation, theCase))
```
lgonczar: How about doing this mapping inside the previous map call (before `mapError`)? Then `validateClassification` would have to worry about the result of previous validation steps and could just accept classification and case.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
lgonczar: Given OCR is not always required, nice one.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
doncem: good call by doubting this orchestration. let's see, we could do something like following:

```java
map(validation -> validation
    .mapError(throwable -> "...")
    .flatMap(classification -> validateClassification(theCase, classification))
)
```

solves the nesting of validation objects used in `validateClassification`.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -192,9 +192,40 @@ static boolean hasOcr(CaseDetails theCase) {
...
+            return invalid(format(
+                "Event %s not allowed for the current journey classification %s",
+                "createCase",
+                classification
+            ));
+        }
+
+        if (EXCEPTION.equals(classification)) {
+            Optional<List<Map<String, Object>>> ocrDataOpt = getOcrData(theCase);
+            if (!ocrDataOpt.isPresent() || ocrDataOpt.get().isEmpty()) {
```
doncem: Looks similar as `hasOcrData`. Can we reuse it? after that it's nicer to join both ifs into 1

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -192,9 +192,40 @@ static boolean hasOcr(CaseDetails theCase) {
...
+        CaseDetails theCase
+    ) {
+        if (validation.isInvalid()) {
+            return validation;
+        }
+        Classification classification = validation.get();
+        if (SUPPLEMENTARY_EVIDENCE.equals(classification)) {
+            return invalid(format(
+                "Event %s not allowed for the current journey classification %s",
+                "createCase",
```
doncem: no need to parameterise this (and in following log entry)

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/callback/CreateCaseValidator.java
```diff
@@ -100,9 +100,8 @@ public CreateCaseValidator() {
         return getOcrData(caseDetails)
             // following mapError should never happen as getting should be non-breaking
             // left side must be String
```
doncem: Comments above no longer valid

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/callback/CreateCaseValidator.java
```diff
@@ -113,7 +112,8 @@ public CreateCaseValidator() {
                     ))
                     .collect(toList())
                 ).toValidation().mapError(throwable -> "Invalid OCR data format. Error: " + throwable.getMessage())
-            );
+            )
+            .orElse(Validation.valid(null));
```
doncem: How about we use empty list here instead?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -191,10 +190,32 @@ static boolean hasOcr(CaseDetails theCase) {
...
             .orElse(invalid("Missing journeyClassification"));
     }
 
+    private static Validation<String, Classification> validateClassification(
+        Classification classification,
+        CaseDetails theCase
+    ) {
+        if (SUPPLEMENTARY_EVIDENCE.equals(classification)) {
+            return invalid(format(
+                "Event createCase not allowed for the current journey classification %s",
```
lgonczar: Having had some thoughts about it, I think we should get rid of the whole "Event createCase" part, as this code doesn't really know what event is being handled. Also, this message would be displayed to the caseworker, who doesn't understand this value. It's Ok to just say sth like "Classification X doesn't allow for creating new cases" or sth like that.

doncem: That's not what ticket suggests. Shall we continue with our BA before making a change?

lgonczar: Oh, you're right. Yes, let's have a quick chat.

lgonczar: Ok, this can always be refined.

lgonczar:
Please wait with merging this one until master build is green.

PR #585, opened 2019-09-23
Make service result object more sensible

doncem:

### Change description ###

It would make more sense to have a decent enough object as data class which holds relevant information useful to make a callback response - introducing one

This is potentially if not eventually last change to service's `process` method in terms of return object

```
[ ] Yes
[x] No
```


doncem:
Forgot to modify edge cases. Will amend the PR later today

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -100,16 +101,21 @@ public CreateCaseCallbackService(
...
+            return Validation.valid(new ProcessResult(
+                ImmutableMap.of("caseReference", UUID.randomUUID()),
+                transformationResponse.warnings
+            ));
         } catch (InvalidCaseDataException exception) {
-            // let controller deal with this. it is 422 or 400
-            throw exception;
+            return Validation.valid(new ProcessResult(
+                exception.getResponse().warnings,
+                exception.getResponse().errors
```
nitinprabhu: Is it not Service will return a successful response even if there are warning/errors ? service need not throw an exception if say one of the field is missing/has incorrect value

doncem: this PR does resolve exception throw and returns list of warnings/errors to the callback requester

PR #587, opened 2019-09-23
Add create case callback endpoint

doncem:

### Change description ###

This PR also stands for ground base for CCD API to come in - helps out integration testing. Unit testing has reached the pointlessness measure.

```
[ ] Yes
[x] No
```


src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
```diff
@@ -0,0 +1,127 @@
...
+    private static final String IDAM_TOKEN = "idam-token";
+    private static final String USER_ID = "user-id";
+
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+
+    @LocalServerPort
+    int serverPort;
+
+    @Test
+    void should_create_exception_record_if_classification_new_application_with_documents_and_ocr_data() {
```
nitinprabhu: do you mean `should_create_service_case` ?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+        postWithBody(getRequestBody("valid-new-application.json"))
+            .statusCode(OK.value())
+            .body("errors", empty())
+            .body("warnings", empty())
+            .body("data.caseReference", matchesPattern(CASE_ID_REGEX));
+    }
+
+    @Test
+    void should_create_exception_record_if_classification_new_application_with_documents_and_without_ocr_data() {
```
nitinprabhu: without OCR we cannot transform or create service case

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
doncem: good point. just copied/moved from unit tests. but yea - let me rename

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
doncem: and what about new application? this was addressed already with extra validation done on #582

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
```diff
@@ -0,0 +1,127 @@
...
+import static org.hamcrest.Matchers.hasItems;
+import static org.hamcrest.text.MatchesPattern.matchesPattern;
+import static org.springframework.http.HttpStatus.OK;
+
+@IntegrationTest
+class CreateCaseCallbackTest {
+
+    private static final String CASE_ID_REGEX =
+        // will be replaced with real one after ccd api
+        "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$";
```
lgonczar: Should case reference be with hyphens? When we talk to CCD, we normally have to use digits only. Also, CCD returns us case refs in the digit-only format. If this field is for display purposes only, I suppose it's good as it is, but worth checking.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CcdCallbackController.java
```diff
@@ -48,6 +54,28 @@ public CallbackResponse attachToCase(
         }
     }
 
+    @PostMapping(path = "/create-case")
+    public CallbackResponse createCase(
```
lgonczar: Because we're adding this endpoint, we should un-ignore functional tests for it as part of this PR.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
doncem: Using random UUID as a placeholder. Next PR introduces CCD at which point it'll become string of integers (as ccd id is type of long and updates are done with strings

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+    private static final String IDAM_TOKEN = "idam-token";
+    private static final String USER_ID = "user-id";
+
+    @LocalServerPort
+    int serverPort;
+
+    @Test
+    void should_create_case_if_classification_new_application_with_documents_and_ocr_data() {
```
lgonczar: there's no OCR data in the input json

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CcdCallbackController.java
doncem: CCD API not yet in. Introduced controller first as it completes integration testing efficiently without mocking

lgonczar: I think we should wait with merging this change until the CCD API part is done.

doncem: OK. Even though I agree and normally I would develop as we wish but it is easier to build up-to-standards testing framework against source code with integration testing once endpoint is in. Rather than doing CCD part and mocking lot's of things where mocks are already in place and unit test is 600+ lines already.

+! endpoint is not available anyway: no definition file points to it and even if you try - service configuration does not allow it - no transformation endpoint configured. Fail fast approach.

I know this is breaking the mindset, but really - faster approach to accomplishment

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
```diff
@@ -0,0 +1,136 @@
...
+        setUpTransformation(getTransformationResponseBody("ok-no-warnings.json"));
+
+        postWithBody(getRequestBody("valid-new-application-with-ocr.json"))
+            .statusCode(OK.value())
+            .body("errors", empty())
+            .body("warnings", empty())
+            .body("data.caseReference", matchesPattern(CASE_ID_REGEX));
```
nitinprabhu: I would just assert for not null and not have that regex in place as it will be removed soon.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    @Test
+    void should_create_case_if_classification_new_application_with_documents_and_without_ocr_data() {
```
nitinprabhu: this test is invalid I think as we can't create service case when there is no OCR data. Or we could change our test case to say when there is no ocr it is going to fail ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CcdCallbackController.java
```diff
@@ -48,6 +55,28 @@ public CallbackResponse attachToCase(
...
+        @RequestBody CcdCallbackRequest callbackRequest,
+        @RequestHeader(value = "Authorization", required = false) String idamToken,
```
nitinprabhu: any reason this is not required ? This would be required unless we are validating this at later point ?

rkondratowicz: so that we can return 401/403 I think. Making it required will cause Spring Boot handle it and return 400

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
doncem: It won't be removed, next PR is just modifying the regex - asserting correct case id

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
doncem: This needs a separate PR as it was suppose to be addressed within another ticket. In fact, it is in team sign-off. Will move back to in-progress so it can be handled appropriately

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CcdCallbackController.java
doncem: This follows same path as previous callback - always respond with 200 notifying what's the error. If path of this has changed - let's address in separate ticket stating what changes need to be done, how to respond back to ccd

PR #590, opened 2019-09-24
Extend io.vavr.control.Validation interface to flexibly apply larger sized object constructs

doncem:
### Change description ###

`ExceptionRecord` construct requires larger size of arguments to be passed in constructor. At the moment vavr is pmd compatible and my guess it'll stay like that for indefinite time so 8 arguments is a limit. Therefore extending with custom `HmctsValidation` and providing relevant builder (which is for 9 argument constructs). In case `ExceptionRecord` or any other object will require higher number of arguments - can simply bump to further iterations or add more if this one will still be used.

This is not Jira related but problem arose during [Add form type to Exception Record](https://tools.hmcts.net/jira/browse/BPS-819)

```
[ ] Yes
[x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -105,7 +105,8 @@ jacocoTestReport {
   afterEvaluate {
     getClassDirectories().from = getClassDirectories().files.collect {
       fileTree(dir: it, exclude: [
-        'uk/gov/hmcts/reform/bulkscan/orchestrator/config/**'
+        'uk/gov/hmcts/reform/bulkscan/orchestrator/config/**',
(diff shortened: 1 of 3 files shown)


lgonczar:
If we got to the point where we have too many arguments for the lib to support, I think the best action would be adjusting the way we use it. We don't need to call `Validation.combine` or `Validation.ap`. Instead, we can convert validations into a list of errors or exception record explicitly, which would add less code than this extension and make it trivial to add new validations, if required at some point in the future.

doncem:
this was for experimentation reasons. no longer needed

PR #596, opened 2019-09-26
WIP - Add payDcnProcessed to ExceptionRecord

aliveni:
https://tools.hmcts.net/jira/browse/BPS-814

### Change description ###
- Add `payDcnProcessed` field to ExceptionRecord model
- Set defaultValue as "No"  when creating ExceptionRecord
- Assert that ExceptionRecord case details contain `payDcnProcessed` field.

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/ExceptionRecordCreationTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/ExceptionRecordCreationTest.java
@@ -105,6 +105,8 @@ class ExceptionRecordCreationTest {
         // envelope ID from the JSON resource representing the test message
         String messageEnvelopeId = "743aeac9-1791-463a-b929-9526e285fe2e";
         assertThat(caseDetails.getData().get("envelopeId")).isEqualTo(messageEnvelopeId);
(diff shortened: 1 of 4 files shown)


lgonczar:
I'm not sure we have reached conclusion about this one. If our field answers the "is pay dcn processed?" question, then we don't know if we're waiting for it to be processed or whether there's nothing to process. Alternatively, if our field's name was `isAwaitingPayProcessing` or something like this, it would give us exactly the information we need.

What are your thoughts?

aliveni:
> I'm not sure we have reached conclusion about this one. If our field answers the "is pay dcn processed?" question, then we don't know if we're waiting for it to be processed or whether there's nothing to process. Alternatively, if our field's name was `isAwaitingPayProcessing` or something like this, it would give us exactly the information we need.
> 
> What are your thoughts?

I thought the field name was agreed and that's why the ticket was in `ready for sprint`. I should've confirmed with someone in the team.
Anyway, I'll add a TODO to rename the property and not merge this PR.
Also, I realised I missed an important point - set this field value only when envelope contains payments. I'll make this change and mark the pr not to merge.

aliveni:
Closing this PR as the changes are not valid. will create a new PR once payments are added to envelope

PR #598, opened 2019-09-26
Prevent creating multiple exception records from same envelope

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-490

### Change description ###

Don't create an exception record for the envelope if there exists one already, with the same envelopeId.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/CreateExceptionRecord.java
```diff
@@ -25,19 +28,50 @@
 
     private final ExceptionRecordMapper mapper;
     private final CcdApi ccdApi;
+    private final List<String> jurisdictionsWithDuplicatePrevention;
 
     public CreateExceptionRecord(
         ExceptionRecordMapper mapper,
-        CcdApi ccdApi
+        CcdApi ccdApi,
+        @Value("${jurisdictions-with-duplicate-er-prevention}") final List<String> jurisdictionsWithDuplicatePrevention
```
lgonczar: This config will go away as soon as all services have ElasticSearch enabled in prod.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/CreateExceptionRecord.java
```diff
@@ -25,19 +28,50 @@
...
+        this.jurisdictionsWithDuplicatePrevention = jurisdictionsWithDuplicatePrevention;
     }
 
     /**
-     * Creates an Exception Record from given envelope.
+     * Creates an exception record from given envelope, unless an exception record
+     * already exists for this envelope.
      */
-    public void createFrom(Envelope envelope) {
+    public void tryCreateFrom(Envelope envelope) {
```
apereverzin: 'try' prefix does not look good for me, it implies an exception to be thrown. Maybe something like 'createIfPossible' or 'createIfAllowed' would be a better name.

lgonczar: I had the opposite intention - try prefix suggests that the method may not succeed to do its thing and at the same time not throw exception. This convention is well-known in .NET, so I thought I'd apply it here.

apereverzin: ok

lgonczar: On a second thought, this name doesn't fully match the convention, as in case of a technical failure an exception will be thrown. How about `createIfNotNotExisting`?

rkondratowicz: The original method name was fine be me

lgonczar: The original method was just creating an exception method. Now it does it conditionally, so I think it would be ideal to reflect that in its name.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/CreateExceptionRecord.java
rkondratowicz: maybe the `-er-` is not needed? it's not present in parameter name or field name

lgonczar: This configuration exists already in the app. I'm just adding another reference here. ER is not present in parameter/field names in this class, because the class is all about exception records, so it provides context. The configuration property exists only for the transition period.

PR #599, opened 2019-09-26
Add CCD API to create case callback

doncem:

### Change description ###

Core implementation minimal version. As mentioned previously, unit testing was and would be insane if covering everything after including ccd. Therefore basing this on top of endpoint #587 

More to come:

- integration test preferred not unit test
- cleanup
- config (to include transformation urls
- definition files (if covered in the ticket)
- pray for acceptance (actually - praying right now)

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
@@ -10,6 +10,7 @@ import uk.gov.hmcts.reform.bulkscan.orchestrator.config.IntegrationTest;
 import java.io.IOException;
 
 import static com.github.tomakehurst.wiremock.client.WireMock.containing;
+import static com.github.tomakehurst.wiremock.client.WireMock.get;
 import static com.github.tomakehurst.wiremock.client.WireMock.givenThat;
 import static com.github.tomakehurst.wiremock.client.WireMock.okJson;
 import static com.github.tomakehurst.wiremock.client.WireMock.post;
@@ -27,9 +28,7 @@ import static org.springframework.http.HttpStatus.OK;
 @IntegrationTest
 class CreateCaseCallbackTest {
 
-    private static final String CASE_ID_REGEX =
-        // will be replaced with real one after ccd api
-        "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$";
+    private static final String CASE_ID_REGEX = "\\d+";
 
     private static final String IDAM_TOKEN = "idam-token";
     private static final String USER_ID = "user-id";
@@ -40,6 +39,10 @@ class CreateCaseCallbackTest {
     @Test
     void should_create_case_if_classification_new_application_with_documents_and_ocr_data() {
         setUpTransformation(getTransformationResponseBody("ok-no-warnings.json"));
+        setUpCcdCreateCase(
+            getCcdResponseBody("start-event.json"),
+            getCcdResponseBody("sample-case.json")
+        );
 
         postWithBody(getRequestBody("valid-new-application-with-ocr.json"))
             .statusCode(OK.value())
@@ -62,6 +65,10 @@ class CreateCaseCallbackTest {
     @Test
     void should_create_case_if_classification_exception_with_documents_and_ocr_data() {
         setUpTransformation(getTransformationResponseBody("ok-no-warnings.json"));
+        setUpCcdCreateCase(
+            getCcdResponseBody("start-event.json"),
+            getCcdResponseBody("sample-case.json")
+        );
 
         postWithBody(getRequestBody("valid-exception.json"))
             .statusCode(OK.value())
@@ -97,6 +104,26 @@ class CreateCaseCallbackTest {
         );
     }
 
+    private void setUpCcdCreateCase(String startResponseBody, String submitResponseBody) {
+        givenThat(
+            get(
+                // values from config + initial request body
+                "/caseworkers/" + USER_ID + "/jurisdictions/BULKSCAN/case-types/123/event-triggers/createCase/token"
+            )
+            .withHeader("ServiceAuthorization", containing("Bearer"))
+            .willReturn(okJson(startResponseBody))
+        );
+
+        givenThat(
+            post(
+                // values from config + initial request body
+                "/caseworkers/" + USER_ID + "/jurisdictions/BULKSCAN/case-types/123/cases?ignore-warning=true"
+            )
+                .withHeader("ServiceAuthorization", containing("Bearer"))
+                .willReturn(okJson(submitResponseBody))
+        );
+    }
+
     private byte[] getFileContents(String ccdCallbackSubFolders, String filename) {
         try {
             return toByteArray(getResource("ccd/callback/" + ccdCallbackSubFolders + filename));
@@ -111,6 +138,10 @@ class CreateCaseCallbackTest {
 
     private String getTransformationResponseBody(String filename) {
         return new String(getFileContents("transformation-client/response/", filename));
+    }
+
+    private String getCcdResponseBody(String filename) {
+        return new String(getFileContents("../response/", filename));
     }
 
     private ValidatableResponse postWithBody(byte[] body) {
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
@@ -19,16 +19,14 @@ import static com.google.common.io.Resources.toByteArray;
 import static io.restassured.RestAssured.given;
 import static org.hamcrest.Matchers.contains;
 import static org.hamcrest.Matchers.empty;
+import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.hasItem;
 import static org.hamcrest.Matchers.hasItems;
 import static org.hamcrest.Matchers.nullValue;
-import static org.hamcrest.text.MatchesPattern.matchesPattern;
 import static org.springframework.http.HttpStatus.OK;
 
 @IntegrationTest
 class CreateCaseCallbackTest {
-
-    private static final String CASE_ID_REGEX = "\\d+";
 
     private static final String IDAM_TOKEN = "idam-token";
     private static final String USER_ID = "user-id";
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -39,17 +43,20 @@
...
     private final AuthTokenGenerator s2sTokenGenerator;
+    private final CoreCaseDataApi feignCcdApi;
 
     public CreateCaseCallbackService(
         CreateCaseValidator validator,
         ServiceConfigProvider serviceConfigProvider,
         TransformationClient transformationClient,
-        AuthTokenGenerator s2sTokenGenerator
+        AuthTokenGenerator s2sTokenGenerator,
+        CoreCaseDataApi feignCcdApi
```
lgonczar: How about `ccdApi`?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -100,30 +109,50 @@ public CreateCaseCallbackService(
         ExceptionRecord exceptionRecord,
         ServiceConfigItem configItem,
         long caseId,
-        boolean ignoreWarnings
+        boolean ignoreWarnings,
+        String idamToken,
+        String userId
     ) {
+        String caseIdStringify = Long.toString(caseId);
```
lgonczar: Best to make it a nount, e.g. `caseIdString`, `caseIdAsString` or sth like this.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -100,30 +109,50 @@ public CreateCaseCallbackService(
...
+
+        log.info("Start creating exception record for {} {}", configItem.getService(), caseIdStringify);
```
lgonczar: Worth adding `service` and `case ID` into log message. I can see it was already like this, not sure why you changed it.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
doncem: Mixes up with our own CcdApi in codebase. This reeks for general review and unification. tl;dr - could not re-use CcdApi. But perhaps for this class purposes - can be renamed, np

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
doncem: Perhaps because code was done long since, just rebased. Sorry about that. Can add now but will review logging in future PR

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
```diff
@@ -27,9 +28,7 @@
 @IntegrationTest
 class CreateCaseCallbackTest {
 
-    private static final String CASE_ID_REGEX =
-        // will be replaced with real one after ccd api
-        "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$";
+    private static final String CASE_ID_REGEX = "\\d+";
```
lgonczar: Wy don't we verify that values are exactly the same?

doncem: Let me update this then. Was quickly renaming it but intension was to actually copy from json files. I'll do that and add a comment

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -100,30 +109,54 @@ public CreateCaseCallbackService(
...
 
             SuccessfulTransformationResponse transformationResponse = transformationClient.transformExceptionRecord(
                 configItem.getTransformationUrl(),
                 exceptionRecord,
-                s2sTokenGenerator.generate()
+                s2sToken
             );
 
             if (!ignoreWarnings && !transformationResponse.warnings.isEmpty()) {
                 log.warn("Transformation warnings: {}", String.join(", ", transformationResponse.warnings));
```
lgonczar: I know it's not this PR, but worth double checking if we can log warnings from transformation service. @nitinprabhu ?

nitinprabhu: actually warnings may include sensitive Info so it would be safe to not log it

doncem: gone 🐩

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -70,7 +77,9 @@ public CreateCaseCallbackService(
                     exceptionRecord,
                     configItem,
                     request.getCaseDetails().getId(),
-                    request.isIgnoreWarnings()
+                    request.isIgnoreWarnings(),
+                    idamToken,
```
nitinprabhu: hope we are passing this idamToken - one which we recieved from callback ?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
```diff
@@ -18,19 +19,15 @@
...
 import static org.hamcrest.Matchers.hasItem;
 import static org.hamcrest.Matchers.hasItems;
 import static org.hamcrest.Matchers.nullValue;
-import static org.hamcrest.text.MatchesPattern.matchesPattern;
 import static org.springframework.http.HttpStatus.OK;
 
 @IntegrationTest
 class CreateCaseCallbackTest {
 
-    private static final String CASE_ID_REGEX =
```
nitinprabhu: nice we didn't needed regex in first instance 👍

src/integrationTest/resources/ccd/response/start-event.json
```diff
@@ -0,0 +1,5 @@
+{
```
nitinprabhu: worth renaming this file to start-event-response but can be done later

doncem: it's already under `response` folder. no need to be too verbose here with resources

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -100,30 +109,53 @@ public CreateCaseCallbackService(
...
                 exceptionRecord,
-                s2sTokenGenerator.generate()
+                s2sToken
             );
 
             if (!ignoreWarnings && !transformationResponse.warnings.isEmpty()) {
-                log.warn("Transformation warnings: {}", String.join(", ", transformationResponse.warnings));
-
+                // do not log warnings
                 return Validation.invalid(Array.ofAll(transformationResponse.warnings));
```
lgonczar: Maybe not necessarily for this PR, but this method needs to start returning a different kind of result. Currently, it returns:
* valid when successful response
* valid when errors
* invalid when warnings

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
```diff
@@ -40,12 +37,16 @@
...
 
         postWithBody(getRequestBody("valid-new-application-with-ocr.json"))
             .statusCode(OK.value())
             .body("errors", empty())
             .body("warnings", empty())
-            .body("data.caseReference", matchesPattern(CASE_ID_REGEX));
+            .body("data.caseReference", equalTo("1539007368674134")); // from sample-case.json
     }
 
     @Test
```
lgonczar: I've just noticed that this test expects the body to contain warnings in `errors` field.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
doncem: There's inconsistency I agree, but on the ccd it ends up in a same way - error'ed. There's a tiny possibility to achieve errors + warnings with this new design, but essentially it'll end up as errors disregardingly. The way it was before with another callback endpoint

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
doncem: Let me return a warning then. Makes sense. It all comes down to this "proper way of handling response from service" matter

lgonczar:
There are two, in my opinion, important things to test somewhere:
* have we sent the right data to transformation service, i.e. consistent with the data that we received with the callback?
* have we created the case with data from transformation service?

Do we have the above tested somewhere?

doncem:
This is explicitly spec'ing out hardcoded data aka double hardcoding the verification. If it is a blocker - fair thing, can add. If it is not then all static tests cover edge cases where necessary.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -100,30 +110,56 @@ public CreateCaseCallbackService(
...
+                s2sToken
             );
 
             if (!ignoreWarnings && !transformationResponse.warnings.isEmpty()) {
-                log.warn("Transformation warnings: {}", String.join(", ", transformationResponse.warnings));
-
-                return Validation.invalid(Array.ofAll(transformationResponse.warnings));
-            } else {
+                // do not log warnings
                 return Validation.valid(new ProcessResult(
```
lgonczar: I think we don't have any `Validation.invalid` case anymore, so this method can return ProcessResult. Can be done in a future PR.

doncem: Yep. part of whichever cleanup I'm planning to do. Ta

doncem:
last changes only affect static tests. they passed. merging. 🕔

PR #602, opened 2019-09-27
Stop dead-lettering envelope messages when notification fails

lgonczar:
THIS PR CAN'T BE MERGED UNTIL DUPLICATE EXCEPTION RECORD PREVENTION IS IN PLACE FOR ALL SERVICES.

https://tools.hmcts.net/jira/browse/BPS-814

### Change description ###

Stop dead-lettering messages from `envelopes` queue when sending corresponding messages to `processed-envelopes` queue fails. Do nothing instead, so that messages will be reprocessed.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java
```diff
@@ -93,10 +93,7 @@ private MessageProcessingResult process(IMessage message) {
                 return new MessageProcessingResult(UNRECOVERABLE_FAILURE, ex);
             } catch (NotificationSendingException ex) {
                 logMessageProcessingError(message, envelope, ex);
-
-                // CCD changes have been made, so it's better to dead-letter the message and
-                // not repeat them, at least until CCD operations become idempotent
-                return new MessageProcessingResult(UNRECOVERABLE_FAILURE, ex);
+                return new MessageProcessingResult(POTENTIALLY_RECOVERABLE_FAILURE, ex);
```
lgonczar: I just realised I can remove this whole catch clause.

rkondratowicz: You were adding `ex` as parameter here and it's not being added in the last `catch` clause.
Not sure what's preferrend

lgonczar: Exception is only needed for `UNRECOVERABLE_FAILURE` results - the message would be in the reason in dead-lettered message. For recoverable failures we have no use for exceptions.

PR #606, opened 2019-09-30
Add PROBATE callback url for demo env

doncem:

### Change description ###

Including config so bug fishing can be started on demo environment while probate implements their forms.

Including chart config as default as we are not live with aks and this feature is ready / will be ready in blink of an eye anyway. If requested - can revert single commit addressing that

```
[ ] Yes
[x] No
```


doncem:
We will end up with too many env vars anyway :(. Each service will have to provide their own urls

doncem:
I believe it's ok to use default ones and env will be same as per boot - whoever will boot it up as a product. in terms of our own build in preview - we actually not using those urls

PR #618, opened 2019-10-02
Fix case data dates

doncem:

### Change description ###

Specs not showing `Z` being present in date time fields and it is mandatory for instant parser. Customising formatter so any format can be accepted. Modified integration tests to prove the point

This originates from 2 strange outcomes from ccd alone

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/resources/ccd/callback/create-case/request/valid-exception.json
+++ b/src/integrationTest/resources/ccd/callback/create-case/request/valid-exception.json
@@ -12,8 +12,8 @@
       "poBox": "PO 12345",
       "journeyClassification": "EXCEPTION",
       "formType": "B123",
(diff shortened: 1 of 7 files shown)


src/integrationTest/resources/ccd/callback/create-case/request/valid-exception.json
```diff
@@ -12,8 +12,8 @@
       "poBox": "PO 12345",
       "journeyClassification": "EXCEPTION",
       "formType": "B123",
-      "deliveryDate": "2018-01-02T12:34:56",
-      "openingDate": "2018-01-03T12:34:56",
+      "deliveryDate": "2018-01-02T12:34:56.123",
```
aliveni: missing `Z`

PR #619, opened 2019-10-03
Fix issue parsing datetime object for scanned docs

doncem:

### Change description ###

Previously #611 was looking in the wrong place assuming case details date was wrong. I think it was in the first place but then forgot/skipped scanned docs too - couple date fields are there as well with the same issue - not having millis not `Z`

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
@@ -34,7 +34,8 @@ public final class CallbackValidations {
 
     private static final Logger log = LoggerFactory.getLogger(CallbackValidations.class);
 
-    private static final DateTimeFormatter FORMATTER = new DateTimeFormatterBuilder()
+    // todo review usage
+    public static final DateTimeFormatter FORMATTER = new DateTimeFormatterBuilder()
         // date/time
         .append(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
         // optional offset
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/callback/CreateCaseValidator.java
```diff
@@ -158,8 +159,8 @@ private ScannedDocument mapScannedDocument(Map<String, Object> document) {
             ((Map<String, String>) document.get("url")).get("document_url"),
             (String) document.get("controlNumber"),
             (String) document.get("fileName"),
-            Instant.parse((String) document.get("scannedDate")),
-            Instant.parse((String) document.get("deliveryDate"))
+            Instant.from(FORMATTER.parse((String) document.get("scannedDate"))),
+            Instant.from(FORMATTER.parse((String) document.get("deliveryDate")))
```
rkondratowicz: no change in tests ? (or new tests)

doncem: let me have a look. don't want to just add tests which cover similar functionality with integration. i have success path unit tests handy amend to cover such cases

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -34,7 +34,8 @@
 
     private static final Logger log = LoggerFactory.getLogger(CallbackValidations.class);
 
-    private static final DateTimeFormatter FORMATTER = new DateTimeFormatterBuilder()
+    // todo review usage
+    public static final DateTimeFormatter FORMATTER = new DateTimeFormatterBuilder()
```
lgonczar: Not necessarily here, but because we're using this formatter not just from CallbackValidations, it's best if it goes to a more common place.

doncem: exactly. i'm trying to fix as quickly as possible. to get the happy path working. together with @apereverzin functional tests. looks like this one is last conceptually. now we are hitting the wall while respond gets back from a third party dependency :/

PR #625, opened 2019-10-04
Fix DateTime serialisation for RestTemplate requests

doncem:
### Change description ###

Kept on getting floating point and/or array of integers for datetime no matter what. This PR follows hmcts/send-letter-service#440 as this issue caused problems before and only today remembered it. The follow up didn't quite work and as you can see in the `HttpConfiguration`, some amendments to `RestTemplate` had to be done so it can identify relevant serialisation plugins

Planned to sneak in test upgrade for transformation client but it didn't seem trivial so parked it for the time being - the fix right now has higher priority

```
[ ] Yes
[x] No
```


doncem:
Breaks functional tests as highly dependant due to document management client being there and using rest template. Will attempt with less prone-y version

PR #630, opened 2019-10-07
Add request logging for CCD API client

doncem:

### Change description ###

Debug commit. Logging request for ccd callback to verify correctness.

Bonus - logging all errors as well as per story requirement

```
[ ] Yes
[x] No
```


lgonczar:
If this PR is going to be merged, it has to disable request logging for production first.

doncem:
This will never be logged as access to endpoint is enabled to probate on demo and bulkscan on aat

lgonczar:
> This will never be logged as access to endpoint is enabled to probate on demo and bulkscan on aat

I don't think we should rely on that. It's going to be easy to forget that we have this logging.

PR #644, opened 2019-10-09
Payments: Add method for sending update commands to the queue

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-824

Just adding the method, not used by orchestrator yet. (will be added in next PR).

Note that the ID for messages sent on queue is now a random UUID

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/payments/PaymentsPublisher.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/payments/PaymentsPublisher.java
@@ -10,8 +10,10 @@ import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.context.annotation.Profile;
 import org.springframework.stereotype.Service;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.domains.payments.model.CreatePaymentsCommand;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.domains.payments.model.UpdatePaymentsCommand;
 
 import java.time.Instant;
+import java.util.UUID;
 
 import static org.springframework.http.MediaType.APPLICATION_JSON;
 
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/payments/PaymentsPublisher.java
```diff
@@ -34,24 +36,30 @@ public PaymentsPublisher(
...
+    }
+
+    private void send(Object cmd, String label) {
+        try {
             IMessage message = new Message(
-                cmd.ccdReference,
-                messageBody,
+                UUID.randomUUID().toString(),
+                objectMapper.writeValueAsString(cmd),
                 APPLICATION_JSON.toString()
```
doncem: just out of nowhere question: is that a type of message saying it's gonna be json format and not byte array? i believe it goes into different parameter in service bus' message body object. i might be panicking in vain but - just asking

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/payments/PaymentsPublisher.java
```diff
@@ -34,24 +36,30 @@ public PaymentsPublisher(
...
+    public void send(UpdatePaymentsCommand cmd) {
+        send(cmd, Labels.UPDATE);
+    }
+
+    private void send(Object cmd, String label) {
+        try {
             IMessage message = new Message(
-                cmd.ccdReference,
-                messageBody,
+                UUID.randomUUID().toString(),
```
doncem: > Note that the ID for messages sent on queue is not a random UUID

🤔

rkondratowicz: typo, I wanted to say **NOW**

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/payments/PaymentsPublisher.java
rkondratowicz: Afaik it's just a regular property.
We never use it on the reading end btw.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/payments/PaymentsPublisher.java
```diff
@@ -34,24 +36,30 @@ public PaymentsPublisher(
...
 
-            LOG.info("Sent message to payments queue. CCD Reference: {}", cmd.ccdReference);
+            LOG.info("Sent message to payments queue. ID: {}, Label: {}", message.getMessageId(), message.getLabel());
+
         } catch (Exception ex) {
             throw new PaymentsPublishingException(
-                String.format(
-                    "An error occurred when trying to publish payments for CCD Ref %s", cmd.ccdReference
-                ),
+                "An error occurred when trying to publish message to payments queue.",
```
lgonczar: Why did we get rid of CCD Ref?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/payments/PaymentsPublisher.java
```diff
@@ -34,24 +36,30 @@ public PaymentsPublisher(
...
-        try {
-            String messageBody = objectMapper.writeValueAsString(cmd);
+        send(cmd, Labels.CREATE);
+    }
 
+    @Override
+    public void send(UpdatePaymentsCommand cmd) {
+        send(cmd, Labels.UPDATE);
+    }
+
```
apereverzin: IMHO should be better to fetch label from the command itself. There should be some interface defining getLabel() method implemented in each command class. This will end up in having just one send() method.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/payments/PaymentsPublisher.java
doncem: Down the line, theoretically, content type is there to distinguish what kind of body one is sending. ATM it still translates to byte array but in my opinion it's lack of restriction or implementation of service bus.

All good. Apparently was panicking too much as it ~looks like~ is translating to byte array anyway

PR #648, opened 2019-10-10
Make sure document type is passed lower case

doncem:

### Change description ###

Sample app already reads as plain string. CCD value has to be lower case. The creation of cases happening from `Envelope` message also has same object structure. Verifying lowercase existence with regex in integration test

```
[ ] Yes
[x] No
```


rkondratowicz:
let's discuss

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/transformation/model/request/DocumentType.java
```diff
@@ -1,8 +1,22 @@
...
+
+    DocumentType(String value) {
+        this.value = value;
+    }
+
+    @Override
+    @JsonValue
+    public String toString() {
+        return value;
+    }
```
apereverzin: We need to check that it work in a test.

doncem: it is checking as a regex providing only lower cases

PR #654, opened 2019-10-11
Simplify 'create case' service

rkondratowicz:
Service used to return `Either<List<String>, ProcessResult>` which is list of errors or result which can also have errors...

Also avoiding obscure methods and types that are result of calling them, like
```
Validation<Seq<Seq<String>>, Validation<Seq<String>, ProcessResult>>
```

Separated calling transformation from creating case in ccd.

Retrieving transformation is now done once, twice before (which was easy to miss)

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
@@ -61,7 +61,7 @@ class CreateCaseCallbackTest {
             .statusCode(OK.value())
             .body("errors", contains("Event " + EVENT_ID_CREATE_NEW_CASE + " not allowed "
                 + "for the current journey classification " + NEW_APPLICATION.name() + " without OCR"))
-            .body("warnings", nullValue())
+            .body("warnings", empty())
             .body("data", nullValue());
     }
 
@@ -71,7 +71,7 @@ class CreateCaseCallbackTest {
             .statusCode(OK.value())
             .body("errors", contains("Event " + EVENT_ID_CREATE_NEW_CASE + " not allowed "
                 + "for the current journey classification " + EXCEPTION.name() + " without OCR"))
-            .body("warnings", nullValue())
+            .body("warnings", empty())
             .body("data", nullValue());
     }
 
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/transformation/TransformationClient.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/transformation/TransformationClient.java
@@ -39,7 +39,7 @@ public class TransformationClient {
         String baseUrl,
         ExceptionRecord exceptionRecord,
         String s2sToken
-    ) throws IOException, CaseTransformationException {
+    ) {
         HttpHeaders headers = new HttpHeaders();
         headers.add("ServiceAuthorization", s2sToken);
 
@@ -64,13 +64,18 @@ public class TransformationClient {
             );
 
             if (ex.getStatusCode().equals(UNPROCESSABLE_ENTITY) || ex.getStatusCode().equals(BAD_REQUEST)) {
-                throw new InvalidCaseDataException(
-                    ex,
-                    objectMapper.readValue(ex.getResponseBodyAsByteArray(), TransformationErrorResponse.class)
-                );
+                try {
+                    TransformationErrorResponse error = objectMapper.readValue(
+                        ex.getResponseBodyAsByteArray(),
+                        TransformationErrorResponse.class
+                    );
+                    throw new InvalidCaseDataException(ex, error);
+                } catch (IOException e) {
+                    throw new InvalidCaseDataException(ex, null);
+                }
+            } else {
+                throw new CaseTransformationException(ex, ex.getResponseBodyAsString());
             }
-
-            throw new CaseTransformationException(ex, ex.getResponseBodyAsString());
         }
     }
 }
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
@@ -4,6 +4,7 @@ import com.google.common.collect.ImmutableMap;
 import io.vavr.control.Validation;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Service;
 import uk.gov.hmcts.reform.authorisation.generators.AuthTokenGenerator;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.client.transformation.InvalidCaseDataException;
@@ -108,7 +109,11 @@ public class CreateCaseCallbackService {
                     } catch (ServiceNotConfiguredException exc) {
                         return new ProcessResult(emptyList(), singletonList(exc.getMessage()));
                     } catch (InvalidCaseDataException exc) {
-                        return new ProcessResult(exc.getResponse().warnings, exc.getResponse().errors);
+                        if (exc.getStatus() == HttpStatus.BAD_REQUEST) {
+                            throw exc;
+                        } else {
+                            return new ProcessResult(exc.getResponse().warnings, exc.getResponse().errors);
+                        }
                     } catch (Exception exc) {
                         log.error("Error handling event", exc);
                         return new ProcessResult(emptyList(), singletonList("Internal error"));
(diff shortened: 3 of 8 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -64,136 +55,65 @@ public CreateCaseCallbackService(
...
-        } catch (Exception exception) {
-            log.error(
-                "Failed to create exception for service {} and exception record {}",
-                configItem.getService(),
-                caseIdAsString,
-                exception
-            );
-
-            return Validation.invalid(Array.of("Internal error. " + exception.getMessage()));
+        Validation<String, Void> event = isCreateNewCaseEvent(request.getEventId());
```
lgonczar: How about `eventValidation`?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -64,136 +55,65 @@ public CreateCaseCallbackService(
...
+        if (event.isInvalid()) {
+            return new ProcessResult(emptyList(), singletonList(event.getError()));
+        } else {
+            return hasServiceNameInCaseTypeId(request.getCaseDetails())
+                .map(serviceName -> {
+                    try {
+                        ServiceConfigItem serviceCfg = serviceConfigProvider.getConfig(serviceName);
+                        if (serviceCfg == null || serviceCfg.getTransformationUrl() == null) {
+                            return new ProcessResult(emptyList(), singletonList("Transformation URL is not configured"));
```
lgonczar: I know that it's not about what this PR is for, but this error should be returned as a 500 response and the caseworker should get a generic error.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -64,136 +55,65 @@ public CreateCaseCallbackService(
...
+                                            s2sTokenGenerator.generate()
+                                        );
+                                    long newCaseId = createNewCaseInCcd(
+                                        idamToken,
+                                        userId,
+                                        exceptionRecord.poBoxJurisdiction,
+                                        transformationResp.caseCreationDetails,
+                                        request.getCaseDetails().getId().toString()
+                                    );
+                                    return new ProcessResult(ImmutableMap.of(CASE_REFERENCE, Long.toString(newCaseId)));
```
lgonczar: This method is long and there's a lot of wrapping in it. At least this bit should be extracted out.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/transformation/TransformationClient.java
```diff
@@ -64,13 +64,18 @@ public SuccessfulTransformationResponse transformExceptionRecord(
...
-                    objectMapper.readValue(ex.getResponseBodyAsByteArray(), TransformationErrorResponse.class)
-                );
+                try {
+                    TransformationErrorResponse error = objectMapper.readValue(
+                        ex.getResponseBodyAsByteArray(),
+                        TransformationErrorResponse.class
+                    );
+                    throw new InvalidCaseDataException(ex, error);
+                } catch (IOException e) {
+                    throw new InvalidCaseDataException(ex, null);
```
lgonczar: This should be a 500 from the orchestrator

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/transformation/TransformationClient.java
```diff
@@ -64,13 +64,18 @@ public SuccessfulTransformationResponse transformExceptionRecord(
             );
 
             if (ex.getStatusCode().equals(UNPROCESSABLE_ENTITY) || ex.getStatusCode().equals(BAD_REQUEST)) {
```
lgonczar: This was supposed to be changed, but the behaviour is still incorrect - unprocessable entity should result in a 200 response with errors and warnings, but a bad request should result with a 500 response from us.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -64,136 +55,65 @@ public CreateCaseCallbackService(
...
+                                })
+                                .getOrElseGet(errors -> new ProcessResult(emptyList(), errors.asJava()));
+                        }
+                    } catch (ServiceNotConfiguredException exc) {
+                        return new ProcessResult(emptyList(), singletonList(exc.getMessage()));
```
lgonczar: That should also be a 500

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -64,136 +55,65 @@ public CreateCaseCallbackService(
...
+                    } catch (InvalidCaseDataException exc) {
+                        return new ProcessResult(exc.getResponse().warnings, exc.getResponse().errors);
+                    } catch (Exception exc) {
+                        log.error("Error handling event", exc);
+                        return new ProcessResult(emptyList(), singletonList("Internal error"));
```
lgonczar: 500, too

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/transformation/TransformationClient.java
apereverzin: Should not it be better to have a constructor of InvalidCaseDataException with 'String msg' parameter and use ex.getResponseBodyAsString() as this parameter? Rather than using null

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
rkondratowicz: won't help, you'd still have to pass many parameters and wrap.
See how it was before. with both transformation and case creation in single method.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
```diff
@@ -61,7 +61,7 @@ void should_not_create_case_if_classification_new_application_without_ocr_data()
             .statusCode(OK.value())
             .body("errors", contains("Event " + EVENT_ID_CREATE_NEW_CASE + " not allowed "
                 + "for the current journey classification " + NEW_APPLICATION.name() + " without OCR"))
-            .body("warnings", nullValue())
```
rkondratowicz: No idea why `null` was expected here.
Other tests have empty arrays.
I guess I accidentally fixed something.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
rkondratowicz: Let's keep those notes for later, but this PR is strictly about refactoring, keeping existing tests passing.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
lgonczar: This problem was created by this PR

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -64,136 +55,99 @@ public CreateCaseCallbackService(
...
+                                            Long.toString(newCaseId)
+                                        ));
+                                    }
+                                })
+                                .getOrElseGet(errors -> new ProcessResult(emptyList(), errors.asJava()));
+                        }
+                    } catch (ServiceNotConfiguredException exc) {
+                        return new ProcessResult(emptyList(), singletonList(exc.getMessage()));
+                    } catch (InvalidCaseDataException exc) {
+                        return new ProcessResult(exc.getResponse().warnings, exc.getResponse().errors);
```
lgonczar: I can see the old code was checking if the response was a 400 or not - after the change it doesn't. Only after 422 we should return a response like this. In all other cases it should be an exception thrown, so that CCD would return 500.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/transformation/TransformationClient.java
lgonczar: this is also new

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
lgonczar: The new method would take 5 parameters. Not great, but it's the same number as an already existing call here. In my opinion, the method would be much more readable if this whole bit was extracted. As of now, it's hard to tell what part the following `.getOrElseGet(..` corresponds with.

PR #655, opened 2019-10-11
Add containsPayments field to Exception record

aliveni:
https://tools.hmcts.net/jira/browse/BPS-784

### Change description ###
duplicate of #637 which was merged on demo for testing the definition changes.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/ExceptionRecord.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/ExceptionRecord.java
@@ -43,6 +43,10 @@ public class ExceptionRecord implements CaseData {
     @JsonProperty("awaitingPaymentDCNProcessing")
     public final String awaitingPaymentDcnProcessing;
 
+    // Yes/No field indicating if the exception record contains payments
+    @JsonProperty("containsPayments")
+    public final String containsPayments;
+
     public ExceptionRecord(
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/ExceptionRecord.java
```diff
@@ -43,6 +43,10 @@
     @JsonProperty("awaitingPaymentDCNProcessing")
     public final String awaitingPaymentDcnProcessing;
 
+    // Yes/No field indicating if the exception record contains payments
+    @JsonProperty("containsPayments")
+    public final String containsPayments;
```
rkondratowicz: can this be boolean?

aliveni: It can be, but we have to send Yes/No value in the CCD request

rkondratowicz: ok

aliveni: Actually, it can't be boolean. this is the JSON property type and it should be `string`.
I was talking about the construction parameter earlier.

PR #661, opened 2019-10-15
Simplify create case callback service signature

doncem:
### Change description ###

The idea looks like partially applied from #654. This PR meant to address the signature as wrapping with `Either<ListOfSomething, ProcessResult>` became really irrelevant.

What we benefit here?

- straight-forward conversion to response object for ccd
- logging is more to the topic
  - if error - error
  - if warning - warning

There could be more verbose things happening if we try to move away from vavr but ideally would keep it separately as it tends to blow. Plus there are still "tiny" validations should come in shortly:

- must have at least one scanned document
- make sure idam token and user id are not empty

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
@@ -4,7 +4,6 @@ import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMap;
 import io.vavr.collection.Array;
 import io.vavr.collection.Seq;
-import io.vavr.control.Either;
 import io.vavr.control.Try;
 import io.vavr.control.Validation;
 import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -86,15 +80,22 @@ public CreateCaseCallbackService(
...
-                log.warn("Warnings found during callback process:\n  - {}", String.join("\n  - ", warnings))
-            );
+            .getOrElseGet(errors -> new ProcessResult(emptyList(), errors));
+
+        if (!result.getWarnings().isEmpty()) {
+            log.warn("Warnings found during callback process:\n  - {}", String.join("\n  - ", result.getWarnings()));
+        }
+
+        if (!result.getErrors().isEmpty()) {
+            log.error("Errors found during callback process:\n  - {}", String.join("\n  - ", result.getErrors()));
```
nitinprabhu: could you please log ccd case reference and jurisdiction/service name

doncem: Can do reference for now. for service will need a further split as it might be one of the reasons callback failed so comes out from validation process/progress

doncem: oh. actually, ID comes from validation too

PR #662, opened 2019-10-15
Clear warnings after case creation from exception record

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-758

### Change description ###

Clear warnings in exception records when the case was successfully created from it.

```
[ ] Yes
[x] No
```


src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/CreateCaseTest.java
```diff
@@ -88,10 +92,30 @@ public void should_create_case_from_valid_exception_record() throws Exception {
         assertThat(createdCase.getData().get("email")).isEqualTo("hello@test.com");
     }
 
+    @Test
+    public void should_clear_exception_record_warnings() throws Exception {
```
lgonczar: The story mentions functional test, but having written it, I'm not entirely sure it's needed - we have it pretty much covered in the integration test and there's not much more we can check here.

PR #664, opened 2019-10-16
Simplify service method - no need to flatmap at all

doncem:

### Change description ###

Reducing the nesting of validations causing multiple flatmaps. The fact that service config is a mandatory prerequisite allows to extract and use directly in single `map` statement after initial validation. Job done how it is done before

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
@@ -90,11 +90,19 @@ public class CreateCaseCallbackService {
             .getOrElseGet(errors -> new ProcessResult(emptyList(), errors));
 
         if (!result.getWarnings().isEmpty()) {
-            log.warn("Warnings found during callback process:\n  - {}", String.join("\n  - ", result.getWarnings()));
+            log.warn(
+                "Warnings found for {} during callback process:\n  - {}",
+                serviceConfigItem.getService(),
+                String.join("\n  - ", result.getWarnings())
+            );
         }
 
         if (!result.getErrors().isEmpty()) {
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -66,36 +63,49 @@ public CreateCaseCallbackService(
...
-                .mapError(errors -> errors.flatMap(Function.identity()))
-                .flatMap(Function.identity())
-                .mapError(Seq::asJava)
-            )
+        Validation<String, Void> canAccess = assertAllowToAccess(request.getCaseDetails(), request.getEventId());
+
+        if (canAccess.isInvalid()) {
+            // log happens in assertion method
+            return new ProcessResult(emptyList(), singletonList(canAccess.getError()));
+        }
```
lgonczar: I like this change. As for what is returned, this case should be a 500 result, so throwing an exception would be ideal. Not necessarily for this PR.

lgonczar: Here or somewhere else, we could log these errors, as they come from us.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -183,7 +187,7 @@ public ProcessResult process(CcdCallbackRequest request, String idamToken, Strin
                 exception
             );
 
-            return Validation.invalid(Array.of("Internal error. " + exception.getMessage()));
+            return new ProcessResult(emptyList(), singletonList("Internal error. " + exception.getMessage()));
```
lgonczar: Probably not for this PR, but this should also result in a 500 response.

PR #667, opened 2019-10-17
Include exception record ID in transformation

doncem:

### Change description ###

Exception record is needed for references and for idempotency. Depends on these PRs:

- hmcts/bulk-scan-ccd-definitions#74
- hmcts/bulk-scan-ccd-event-handler-sample-app#83

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
@@ -81,7 +81,6 @@ public class CreateCaseCallbackService {
             .map(exceptionRecord -> createNewCase(
                 exceptionRecord,
                 serviceConfigItem,
-                request.getCaseDetails().getId(),
                 request.isIgnoreWarnings(),
                 idamToken,
                 userId
@@ -128,14 +127,15 @@ public class CreateCaseCallbackService {
     private ProcessResult createNewCase(
         ExceptionRecord exceptionRecord,
         ServiceConfigItem configItem,
-        long caseId,
         boolean ignoreWarnings,
         String idamToken,
         String userId
     ) {
-        String caseIdAsString = Long.toString(caseId);
-
-        log.info("Start creating new case for {} from exception record {}", configItem.getService(), caseIdAsString);
+        log.info(
+            "Start creating new case for {} from exception record {}",
+            configItem.getService(),
+            exceptionRecord.id
+        );
 
         try {
             String s2sToken = s2sTokenGenerator.generate();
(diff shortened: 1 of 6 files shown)


lgonczar:
We need to document the necessity to include this field in transformation response, so that it's obvious for product teams. I wonder if it would be even better if we set it ourselves.

It would be useful to log an error if we get a transformation response without exception record reference in it. Otherwise, we would be unaware of the problem that needs addressing.

doncem:
this will be documented (updated confluence on-boarding page)

if we set ourselves then breaks our goal here: not amend case data which comes from transformation and risk failing to create service case as it might not be configured

lgonczar:
> this will be documented (updated confluence on-boarding page)
> 
> if we set ourselves then breaks our goal here: not amend case data which comes from transformation and risk failing to create service case as it might not be configured

Ok, if we don't set it ourselves, then I'd log an error, so that we know there's an issue that needs to be addressed quickly.

doncem:
> Ok, if we don't set it ourselves, then I'd log an error, so that we know there's an issue that needs to be addressed quickly.

sounds very reasonable 👍 . will address that in my logging branch

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -220,22 +228,38 @@ private long createNewCaseInCcd(
...
-                    .description("Case created from exception record ref " + originalCaseId)
+                    .description("Case created from exception record ref " + exceptionRecord.id)
                     .build()
                 )
                 .eventToken(eventResponse.getToken())
                 .build()
         ).getId();
     }
+
+    private void checkBulkScanReferenceIsSet(Map<String, ?> caseData, String exceptionRecordId) {
```
nitinprabhu: Can we not set bulk scan reference and not rely on services ?

nitinprabhu: Also we would need to update phase 2 case field definition in Confluence so that doesn't get missed

doncem: absolutely. that's my next step

lgonczar: Sorry, which one have you just agreed on - setting it ourselves or verifying if the transformation service provided it?

nitinprabhu: @lgonczar  we agreed that we would not modify case data coming back from transformation. and if we send something which is not configured in ccd by them (services) then it’s our fault we broke the callback which we don't want to happen and hence logging would be better as it would make us aware.

PR #670, opened 2019-10-17
Review create case callback logging

doncem:

### Change description ###

Breaking change (see comment in `Yes` section at the bottom) but it is expected and handled by CCD.

Compiled previous comments about logging and exception throws from different PRs.

P.S. A lot less changes to review if you exclude whitespace

```
[x] Yes - callback responses may cause 500s now. ccd already handles them individually as we expect to happen in those cases introduced here
[ ] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/transformation/TransformationClient.java
```diff
@@ -58,19 +58,30 @@ public SuccessfulTransformationResponse transformExceptionRecord(
...
     }
+
+    private void tryParseResponseBody(HttpStatusCodeException exception) throws CaseTransformationException {
+        try {
+            TransformationErrorResponse errorResponse = objectMapper.readValue(
+                exception.getResponseBodyAsByteArray(),
+                TransformationErrorResponse.class
+            );
+
+            throw new InvalidCaseDataException(exception, errorResponse);
```
nitinprabhu: do you need this ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -94,17 +96,18 @@ public ProcessResult process(CcdCallbackRequest request, String idamToken, Strin
...
+                result.getWarnings().size()
             );
         }
 
         if (!result.getErrors().isEmpty()) {
-            log.error(
-                "Errors found for {} during callback process:\n  - {}",
+            // no need to error - it's informational log. specific logs will be error'ed already
+            log.warn(
+                "Errors found for {} during callback process: {}",
```
nitinprabhu: can we include exception record id here ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -94,17 +96,18 @@ public ProcessResult process(CcdCallbackRequest request, String idamToken, Strin
 
         if (!result.getWarnings().isEmpty()) {
             log.warn(
-                "Warnings found for {} during callback process:\n  - {}",
+                "Warnings found for {} during callback process: {}",
```
nitinprabhu: can we include exception record id here ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/transformation/TransformationClient.java
```diff
@@ -58,19 +58,30 @@ public SuccessfulTransformationResponse transformExceptionRecord(
...
+                exceptionRecord.id,
                 ex
             );
 
             if (ex.getStatusCode().equals(UNPROCESSABLE_ENTITY) || ex.getStatusCode().equals(BAD_REQUEST)) {
-                throw new InvalidCaseDataException(
-                    ex,
-                    objectMapper.readValue(ex.getResponseBodyAsByteArray(), TransformationErrorResponse.class)
-                );
+                tryParseResponseBody(ex);
```
apereverzin: Should be renamed to tryParseResponseBodyAndThrowException or smth like this, otherwise it is not clear that itsonly outcome is throwing exception

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/transformation/TransformationClient.java
doncem: as per specification, we should accept custom response object defined in there. applies for 400s and 422s. this is as before just "safer" - catch deserialisation issues and wrap with custom exception instead of forwarding it to the source

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
doncem: not yet possible - needs further refactor/extraction as ID comes from validation (gets included in exception record model)

doncem: applied

PR #679, opened 2019-10-18
Send payment update message after creating case from exception record

rkondratowicz:


rkondratowicz:
closing to start a new pipeline build

PR #712, opened 2019-10-28
Return the whole modified exception record to CCD

lgonczar:
### Change description ###

Make case creation callback endpoint return the whole modified exception record, instead of just the changed fields. This is a bug fix. Previous way resulted in overwriting the whole exception record data.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/callback/ProcessResult.java
```diff
@@ -11,26 +11,26 @@
  */
 public class ProcessResult {
 
-    private final Map<String, Object> modifiedFields;
+    private final Map<String, Object> exceptionRecordData;
```
lgonczar: Changing the name of this field, because now it has a different meaning.

PR #722, opened 2019-10-31
Update `jackson-databind` to be used latest from spring boot itself

doncem:
### Change description ###

Spring boot 2.2 shipped together with latest jackson. Customly bumped databind one in configs is not needed anymore

```
[ ] Yes
[x] No
```


doncem:
Merging as it is fixing OWASP and is already proven in other repositories. There is implementation clash and will raise a different PR to address it

PR #727, opened 2019-11-01
Fix/envelope case reference values

aliveni:
https://tools.hmcts.net/jira/browse/BPS-831

### Change description ###
When case reference values are not provided in the envelope, they should be set as null in the Exception Record.

```
[ ] Yes
[x] No
```


aliveni:
This pr can not be built, as the fields are not present in `aat` yet. 
Merging this pr, as this change is required to test the CCD definition on demo.

PR #733, opened 2019-11-04
Add showEnvelopeCaseReference and showEnvelopeLegacyCaseReference fields to exception record

aliveni:
https://tools.hmcts.net/jira/browse/BPS-831

### Change description ###
Added 2 new fields to exception record which are used to show/hide envelope case reference values on CCD UI.

```
[ ] Yes
[x] No
```


aliveni:
This change can't be built on AAT as the CCD definition is not updated in AAT. I'll raise a new PR once tested on demo.

PR #735, opened 2019-11-04
Updated flex volume env vars secret mappings

nitinprabhu:
### Change description ###

- Secrets are not correctly mapped as it needs to be separated using dot and not using underscores(they way flex volume will mount it).

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/resources/bootstrap.yaml
+++ b/src/main/resources/bootstrap.yaml
@@ -6,16 +6,16 @@ spring:
       aliases:
         idam-client-secret: IDAM_CLIENT_SECRET
         s2s-secret-bulk-scan-orchestrator: S2S_SECRET
-        idam-users-bulkscan-username: IDAM_USERS_BULKSCAN_USERNAME
-        idam-users-bulkscan-password: IDAM_USERS_BULKSCAN_PASSWORD
-        idam-users-cmc-username: IDAM_USERS_CMC_USERNAME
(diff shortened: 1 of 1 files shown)


src/main/resources/bootstrap.yaml
```diff
@@ -6,16 +6,16 @@ spring:
...
-        idam-users-bulkscan-username: IDAM_USERS_BULKSCAN_USERNAME
-        idam-users-bulkscan-password: IDAM_USERS_BULKSCAN_PASSWORD
-        idam-users-cmc-username: IDAM_USERS_CMC_USERNAME
-        idam-users-cmc-password: IDAM_USERS_CMC_PASSWORD
-        idam-users-div-username: IDAM_USERS_DIVORCE_USERNAME
-        idam-users-div-password: IDAM_USERS_DIVORCE_PASSWORD
-        idam-users-probate-username: IDAM_USERS_PROBATE_USERNAME
-        idam-users-probate-password: IDAM_USERS_PROBATE_PASSWORD
-        idam-users-sscs-username: IDAM_USERS_SSCS_USERNAME
-        idam-users-sscs-password: IDAM_USERS_SSCS_PASSWORD
```
doncem: reminds me of the note to us having too much "customisation" with env vars. i guess moving to aks and bootstraping those secrets makes references to `IDAM_USERS_*` obsolete

PR #745, opened 2019-11-07
Set envelope case reference values empty

aliveni:
### Change description ###
Fix master build (Create case functional tests)
Envelope case references fields are set to "null" when values are provided in the envelope. 
This is causing issues when creating a case as we are using the immutable map in some places.
Immutable Map does not allow "null" values and causing problems when creating a case.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/mappers/ExceptionRecordMapper.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/mappers/ExceptionRecordMapper.java
@@ -58,7 +58,7 @@ public class ExceptionRecordMapper {
             envelope.id,
             CollectionUtils.isEmpty(envelope.payments) ? NO : YES,
             CollectionUtils.isEmpty(envelope.payments) ? NO : YES,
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/mappers/ExceptionRecordMapper.java
```diff
@@ -58,8 +58,9 @@ public ExceptionRecord mapEnvelope(Envelope envelope) {
             envelope.id,
             CollectionUtils.isEmpty(envelope.payments) ? NO : YES,
             CollectionUtils.isEmpty(envelope.payments) ? NO : YES,
-            isBlank(envelope.caseRef) ? null : envelope.caseRef,
-            isBlank(envelope.legacyCaseRef) ? null : envelope.legacyCaseRef,
+            // Setting the value as "", as null values are causing issues when used in ImmutableMap
```
nitinprabhu: I would update the comment to say null values are not supported in ImmutableMap

aliveni:
The PR build is still failing as the change is still not deployed on AAT. Merging this PR should fix the build.

PR #765, opened 2019-11-21
Create payments for all exception records

rkondratowicz:
Currently, payments are only created for exception record when its classification is `NEW_APPLICATION`.

This PR changes this behaviour - payments are created for all exception record classifications.

https://tools.hmcts.net/jira/browse/BPS-865


rkondratowicz:
@lgonczar this is covered in the tests for `PaymentsProcessor`

lgonczar:
> @lgonczar this is covered in the tests for `PaymentsProcessor`

Then we should verify we call payments processor

rkondratowicz:
> > @lgonczar this is covered in the tests for `PaymentsProcessor`
> 
> Then we should verify we call payments processor

@lgonczar it's already in this PR

PR #767, opened 2019-11-21
Defined Update URL properties

apereverzin:
[Orchestrator to update cases with Supplementary Evidence](https://tools.hmcts.net/jira/browse/BPS-833)

### Change description ###
Define properties for Update URL in client services.
It is separate from transformation URL (though at the moment the values are the same) to make it possible for the services to implement transform and update functionality in a flexible way, potentially on different servers.

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/bulk-scan-orchestrator/values.preview.template.yaml
+++ b/charts/bulk-scan-orchestrator/values.preview.template.yaml
@@ -10,6 +10,7 @@ java:
     CORE_CASE_DATA_API_URL: "http://ccd-data-store-api-aat.service.core-compute-aat.internal"
     DOCUMENT_MANAGEMENT_URL: "http://dm-store-aat.service.core-compute-aat.internal"
     TRANSFORMATION_URL_BULKSCAN: "http://bulk-scan-sample-app-aat.service.core-compute-aat.internal"
+    UPDATE_URL_BULKSCAN: "http://bulk-scan-sample-app-aat.service.core-compute-aat.internal"
     ENVELOPES_QUEUE_CONNECTION_STRING: "$(SB_CONN_STRING);EntityPath=envelopes"
     ENVELOPES_QUEUE_MAX_DELIVERY_COUNT: "10"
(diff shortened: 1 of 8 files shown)


infrastructure/variables.tf
```diff
@@ -79,6 +79,10 @@ variable "transformation_url_probate" {
   default = ""
 }
 
+variable "update_url_bulkscan" {
+  default = "http://bulk-scan-sample-app-aat.service.core-compute-aat.internal"
```
doncem: there's no need for this variable - can just simply assign directly in main app settings

rkondratowicz: it's consistent with transformation though

doncem: yes. it is. and it has no value in it. we moved to aks, aks defined in all envs. this terraform change is legacy only applied in old demo environment. just keep in consistent and define in main.tf. if endpoint doesn't work it'll just be smth azure cloud provides ;)

charts/bulk-scan-orchestrator/values.preview.template.yaml
```diff
@@ -10,6 +10,7 @@ java:
     CORE_CASE_DATA_API_URL: "http://ccd-data-store-api-aat.service.core-compute-aat.internal"
     DOCUMENT_MANAGEMENT_URL: "http://dm-store-aat.service.core-compute-aat.internal"
     TRANSFORMATION_URL_BULKSCAN: "http://bulk-scan-sample-app-aat.service.core-compute-aat.internal"
+    UPDATE_URL_BULKSCAN: "http://bulk-scan-sample-app-aat.service.core-compute-aat.internal"
```
lgonczar: Consistent with `TRANSFORMATION_URL_...`, so this PR is good, but I think we will need another PR to change the naming of these environment variables to sth like `CASE_TRANSFORMATION_URL_{service}` and `CASE_UPDATE_URL_{service}` to make it clearer what they're about.

PR #773, opened 2019-11-26
get ready for helm version upgrade

mdayican:
### Change description ###

get ready for helm version upgrade

ApiVersion is mandatory in helm chart 
Conditions in Helm charts , indentation of the condition needs to inside the base chart.

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/bulk-scan-orchestrator/requirements.yaml
+++ b/charts/bulk-scan-orchestrator/requirements.yaml
@@ -5,4 +5,4 @@ dependencies:
   - name: servicebus
     version: ~0.2.2
     repository: '@hmctspublic'
-    condition: bso.servicebus.enabled
+    condition: servicebus.enabled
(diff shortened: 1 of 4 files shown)


charts/bulk-scan-orchestrator/requirements.yaml
```diff
@@ -5,4 +5,4 @@ dependencies:
   - name: servicebus
     version: ~0.2.2
     repository: '@hmctspublic'
-    condition: bso.servicebus.enabled
+    condition: servicebus.enabled
```
lgonczar: Why are we removing the first component?

mdayican: they are added to make them unique, with new version  "indentation of the condition needs to inside the base chart." we do not need

nitinprabhu: they were initially there so that it is not applicable globally and now since they are inlined we don't them as such

mdayican:
it is stuck in pipeline, will create new PR.

PR #778, opened 2019-11-28
Callback validations

apereverzin:

### Change description ###
Defined hasId method returning CaseDetails unlike hasAnId returning only id. It will be used in AttachCaseCallbackService.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
@@ -113,6 +113,14 @@ public final class CallbackValidations {
     }
 
     @Nonnull
+    public static Validation<String, CaseDetails> hasId(CaseDetails theCase) {
+        return theCase != null
+            && theCase.getId() != null
+            ? valid(theCase)
+            : invalid("Exception case has no Id");
+    }
+
+    @Nonnull
     static Validation<String, String> hasServiceNameInCaseTypeId(CaseDetails theCase) {
         return Optional
             .ofNullable(theCase)
@@ -163,7 +171,7 @@ public final class CallbackValidations {
                             return !hasOcr(theCase)
                                 ? Validation.<String, Void>valid(null)
                                 : Validation.<String, Void>invalid(
-                                    format("The 'attach to case' event is not supported for exception records with OCR")
+                                format("The 'attach to case' event is not supported for exception records with OCR")
                             );
                         default:
                             return Validation.<String, Void>invalid(
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
@@ -116,7 +116,7 @@ public final class CallbackValidations {
     }
 
     @Nonnull
-    public static Validation<Seq<String>, CaseDetails> hasValidData(
+    public static Validation<Seq<String>, CaseDetails> hasValidDetailsForAttachingToCase(
         boolean useSearchCaseReference,
         CaseDetails theCase
     ) {
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
@@ -191,16 +191,14 @@ public final class CallbackValidations {
                             return hasOcr(theCase)
                                 ? Validation.<String, Void>valid(null)
                                 : Validation.<String, Void>invalid(
-                                format(
                                     "The 'attach to case' event is not supported for supplementary evidence with OCR "
                                         + "but not containing OCR data"
-                                )
                             );
                         case CLASSIFICATION_EXCEPTION:
                             return !hasOcr(theCase)
                                 ? Validation.<String, Void>valid(null)
                                 : Validation.<String, Void>invalid(
-                                format("The 'attach to case' event is not supported for exception records with OCR")
+                                    "The 'attach to case' event is not supported for exception records with OCR"
                             );
                         default:
                             return Validation.<String, Void>invalid(
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
@@ -118,8 +118,6 @@ public final class CallbackValidations {
     @Nonnull
     public static Validation<Seq<String>, CaseDetails> hasValidData(
         boolean useSearchCaseReference,
-        String requesterIdamToken,
-        String requesterUserId,
         CaseDetails theCase
     ) {
         Validation<String, String> caseReferenceTypeValidation = useSearchCaseReference
@@ -132,8 +130,6 @@ public final class CallbackValidations {
         Validation<String, String> serviceNameInCaseTypeIdValidation = hasServiceNameInCaseTypeId(theCase);
         Validation<String, Long> idValidation = hasAnId(theCase);
         Validation<String, List<Map<String, Object>>> scannedRecordValidation = hasAScannedRecord(theCase);
-        Validation<String, String> idamTokenValidation = hasIdamToken(requesterIdamToken);
-        Validation<String, String> userIdValidation = hasUserId(requesterUserId);
 
         Seq<Validation<String, ?>> validations = Array.of(
             caseReferenceTypeValidation,
@@ -141,9 +137,7 @@ public final class CallbackValidations {
             jurisdictionValidation,
             serviceNameInCaseTypeIdValidation,
             idValidation,
-            scannedRecordValidation,
-            idamTokenValidation,
-            userIdValidation
+            scannedRecordValidation
         );
 
         Seq<String> errors = validations
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
@@ -1,5 +1,7 @@
 package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd;
 
+import io.vavr.collection.Array;
+import io.vavr.collection.Seq;
 import io.vavr.control.Try;
 import io.vavr.control.Validation;
 import org.apache.commons.collections4.CollectionUtils;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -112,6 +112,14 @@ private CallbackValidations() {
             : invalid("Exception case has no Id");
     }
 
+    @Nonnull
+    public static Validation<String, CaseDetails> hasId(CaseDetails theCase) {
+        return theCase != null
+            && theCase.getId() != null
+            ? valid(theCase)
+            : invalid("Exception case has no Id");
+    }
```
rkondratowicz: all other validation methods in this class follow convention `hasX` returns `X`.
There's already a method that validates whether the case has Id.

Can you just call it and use `CaseDetails` object that you pass to that method afterwards?
The validation method does not change it so you can use it.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -163,7 +171,7 @@ private CallbackValidations() {
                             return !hasOcr(theCase)
                                 ? Validation.<String, Void>valid(null)
                                 : Validation.<String, Void>invalid(
-                                    format("The 'attach to case' event is not supported for exception records with OCR")
+                                format("The 'attach to case' event is not supported for exception records with OCR")
```
lgonczar: Previous indentation seemed right

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
lgonczar: What will happen to validation result? It can't come back to the caseworker, as it's related with CCD doing something wrong.

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidationsTest.java
```diff
@@ -258,6 +258,62 @@ void noCaseIdTest() {
         );
     }
 
+    @Test
+    void caseIdTest() {
```
lgonczar: I realise these method names are following a convention started for this class, but maybe this PR would be an opportunity to fix the convention. Ideally, method names should provide short information on:
- what's being tested (unless the class specifies it)
- conditions
- expected outcome

lgonczar: @rkondratowicz `[x_]should_y_when_z()` is pretty much what we've got everywhere else. What case does `xTest()` cover?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -112,6 +115,40 @@ private CallbackValidations() {
             : invalid("Exception case has no Id");
     }
 
+    @Nonnull
+    public static Validation<Seq<String>, CaseDetails> hasValidData(
```
lgonczar: This is a general validation class, called `CallbackValidations`. This method returns validation for an exception record that the caseworker wants to attach to a case. I think we should name it something like `hasValidDetailsForAttachingToCase`. Not as nice, but leaves less room for misunderstanding.

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidationsTest.java
```diff
@@ -258,6 +261,79 @@ void noCaseIdTest() {
...
+    void hasInvalidDataTest() {
+        CaseDetails caseDetails = createCaseWith(
+            b -> b
+                .id(null)
+                .caseTypeId("SERVICE_ExceptionRecord")
+                .jurisdiction("BULKSCAN")
+        );
+        Validation<Seq<String>, CaseDetails> res =
+            CallbackValidations.hasValidData(true, caseDetails);
+        assertThat(res.isValid()).isEqualTo(false);
```
lgonczar: I think we should make it a test per error or verify that all possible errors are included in validation result.

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidationsTest.java
```diff
@@ -258,6 +261,79 @@ void noCaseIdTest() {
...
+    }
+
+    @Test
+    void hasValidData_use_search_case_reference_true() {
```
lgonczar: Regarding `useSearchCaseReference`, I think we need the following four scenarios covered:
- `useSearchCaseReference` == `true` and `searchCaseReference` is present => valid
- `useSearchCaseReference` == `true` and `searchCaseReference` is missing => invalid
- `useSearchCaseReference` == `false` and `attachToCaseReference` is present => valid
- `useSearchCaseReference` == `false` and `attachToCaseReference` is missing => invalid

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidationsTest.java
```diff
@@ -258,6 +261,79 @@ void noCaseIdTest() {
...
+
+        CaseDetails caseDetails = createCaseWith(
+            b -> b
+                .id(1L)
+                .caseTypeId("SERVICE_ExceptionRecord")
+                .jurisdiction("BULKSCAN")
+                .data(caseData)
+        );
+        Validation<Seq<String>, CaseDetails> res =
+            CallbackValidations.hasValidData(true, caseDetails);
```
lgonczar: If this was changed to false, I guess the test would fail.

apereverzin: Fixed

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidationsTest.java
apereverzin: Done

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidationsTest.java
apereverzin: Added check for all validation results

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
apereverzin: Done

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -112,6 +115,40 @@ private CallbackValidations() {
             : invalid("Exception case has no Id");
     }
 
+    @Nonnull
+    public static Validation<Seq<String>, CaseDetails> hasValidDetailsForAttachingToCase(
```
nitinprabhu: is this used for validating prerequisites for updating case with OCR + Scanned docs ?
can we name it accordingly `hasValidDetailsForAttachingOCRWithSuppEvidence` ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -112,6 +115,40 @@ private CallbackValidations() {
...
+    ) {
+        Validation<String, String> caseReferenceTypeValidation = useSearchCaseReference
+            ? hasSearchCaseReferenceType(theCase)
+            : valid(CCD_CASE_REFERENCE);
+        Validation<String, String> caseReferenceValidation = useSearchCaseReference
+            ? hasSearchCaseReference(theCase)
+            : hasAttachToCaseReference(theCase);
+        Validation<String, String> jurisdictionValidation = hasJurisdiction(theCase);
+        Validation<String, String> serviceNameInCaseTypeIdValidation = hasServiceNameInCaseTypeId(theCase);
+        Validation<String, Long> idValidation = hasAnId(theCase);
```
nitinprabhu: has service case id or exception case id ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
nitinprabhu: no changes here ?

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidationsTest.java
nitinprabhu: `caseIdTest` is not making that much sense as in what is expected in this test

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
apereverzin: The whole point of this PR is to introduce this method which combines all checks called from AttachCaseCallbackService::getValidation method. In the existing code all this check can be invoked separately because to just attach documents we need these documents and case ID. But for this user story (update case) we need the whole CaseDetails object (i.e. exception record) to send it to the service Update endpoint. None of existing validation methods returns the whole CaseDetails object. This is why this new method has been introduced - it combines all validations and return the whole CaseDetails object.

After this PR is merged the new validation method will be called from AttachCaseCallbackService::getValidation. And AttachCaseCallbackService::AttachToCaseEventData will contain the whole CaseDetails object (returned by the new validation method) instead of just exceptionRecordId.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
apereverzin: hasAnId is an existing method, should not be renamed in this PR. It is checking ID of the exception record

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
apereverzin: No, it is not specific to OCR, it is for any supplementary Evidence

PR #780, opened 2019-11-29
Introduced CcdCaseUpdater

apereverzin:

### Change description ###
Introduced CcdCaseUpdater.  

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -0,0 +1,209 @@
...
+                exceptionRecord.id,
+                exception
+            );
+
+            throw exception;
+        }
+    }
+
+    private void checkBulkScanReferenceIsSet(Map<String, ?> caseData, String exceptionRecordId) {
+        if (!caseData.containsKey(EXCEPTION_RECORD_REFERENCE)) {
```
nitinprabhu: why can't orchestrator set the bulk scan reference . Services may or may not set but we could always set it even they have set we can override.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdNewCaseCreator.java
```diff
@@ -112,11 +112,10 @@ public ProcessResult createNewCase(
                 exceptionRecordFinalizer.finalizeExceptionRecord(exceptionRecordData.getData(), newCaseId)
             );
         } catch (InvalidCaseDataException exception) {
-            if (BAD_REQUEST.equals(exception.getStatus())) {
-                throw new CallbackException("Failed to transform exception record", exception);
-            } else {
-                return new ProcessResult(exception.getResponse().warnings, exception.getResponse().errors);
-            }
+            return ExceptionHandlingUtil.handleInvalidCaseDataException(
```
nitinprabhu: static import ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/ExceptionHandlingUtil.java
```diff
@@ -0,0 +1,26 @@
...
+import static org.springframework.http.HttpStatus.BAD_REQUEST;
+
+final class ExceptionHandlingUtil {
+    private ExceptionHandlingUtil() {
+        //
+    }
+
+    static CallbackException handleGenericException(Exception exception, String msg) {
+        // log happens individually to cover update/ccd cases
+        return new CallbackException(msg, exception);
```
nitinprabhu: we can inline this not sure we need this method. We can get rid of this ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/ExceptionHandlingUtil.java
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    static ProcessResult handleInvalidCaseDataException(InvalidCaseDataException exception, String msg) {
```
nitinprabhu: Also not sure why this is needed. Is it getting repeated at different places ?

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdaterTest.java
```diff
@@ -0,0 +1,409 @@
...
+        given(coreCaseDataApi.submitForCaseworker(
+            anyString(),
+            anyString(),
+            anyString(),
+            anyString(),
+            anyString(),
+            anyBoolean(),
+            any(CaseDataContent.class)
+        )).willReturn(CaseDetails.builder().id(1L).build());
+        given(exceptionRecordFinalizer.finalizeExceptionRecord(anyMap(), anyLong())).willReturn(originalFields);
```
nitinprabhu: can we extract this into a single method ? looks lot of repetition

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdaterTest.java
```diff
@@ -0,0 +1,409 @@
...
+        exceptionRecord = getExceptionRecord();
+
+        given(configItem.getService()).willReturn("Service");
+        given(existingCase.getId()).willReturn(1L);
+        given(configItem.getUpdateUrl()).willReturn("url");
+        given(authTokenGenerator.generate()).willReturn("token");
+    }
+
+    @Test
+    void should_update_case() throws Exception {
```
nitinprabhu: should update case when ?

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdaterTest.java
```diff
@@ -0,0 +1,409 @@
...
+        );
+
+        // then
+        assertThat(res.getErrors()).isEmpty();
+        assertThat(res.getWarnings()).isEmpty();
+        assertThat(res.getExceptionRecordData()).isEqualTo(originalFields);
+    }
+
+    @Test
+    void should_update_case_not_ignore_warnings() throws Exception {
```
nitinprabhu: I think in general it would be nice to follow the pattern which we follow in other test classes

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdNewCaseCreator.java
apereverzin: Oooh!

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
apereverzin: We discussed this recently. At the moment this is delegated to the client service.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/ExceptionHandlingUtil.java
apereverzin: This whole class is to fix a quality gate failure due to code duplication (as mentioned in the comment to the PR)

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdaterTest.java
apereverzin: I tried to guess which pattern you mean

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/ExceptionHandlingUtil.java
nitinprabhu: Could you please post link to sonar qube regarding the quality gate issue as I am not sure it looks like util class is used at only one place ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -0,0 +1,191 @@
...
+
+import static java.lang.String.format;
+import static java.util.Collections.emptyList;
+import static org.springframework.http.HttpStatus.BAD_REQUEST;
+
+@Service
+public class CcdCaseUpdater {
+    private static final Logger log = LoggerFactory.getLogger(CcdCaseUpdater.class);
+
+    private static final String EXCEPTION_RECORD_REFERENCE = "bulkScanCaseReference";
```
lgonczar: I guess this can go now

apereverzin: Oh, yes

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -0,0 +1,189 @@
...
+
+            return new ProcessResult(
+                exceptionRecordFinalizer.finalizeExceptionRecord(existingCase.getData(), existingCase.getId())
+            );
+        } catch (InvalidCaseDataException exception) {
+            if (BAD_REQUEST.equals(exception.getStatus())) {
+                throw new CallbackException("Failed to update case", exception);
+            } else {
+                return new ProcessResult(exception.getResponse().warnings, exception.getResponse().errors);
+            }
```
lgonczar: I realise this pattern is already present for case creation, but there are some problems with it and I think we should make a fix before using CaseUpdateClient.

Here's what we're expecting from update API:
- if the call was successful, return success
- if we sent unprocessable data (well-formed, but business-wise problematic), the update service should return 422 with errors that we should, in turn, return to CCD within a 200 response (so that the caseworkers can read those errors)
- if we sent malformed request, we should get 400 from update service and then return a 500 response to CCD, saying something went wrong, with no specific details

Here we catch `InvalidCaseDataException`, which currently represents a 400/422 response from the update service. This connection is unclear when reading this code. We have to dig deeper to find out what exactly we can expect in the exception. Also, while `BAD_REQUEST` and `UNPROCESSABLE_ENTITY` are the only possibilities at the moment, it's not guaranteed that nobody will at some point reuse it for another 4xx response, introducing a bug.

I think here's what we need to do:
- make `CaseUpdateClient` always throw an exception that strictly represents one response status
- catch those exceptions at the level where we call `CaseUpdateClient`, instead of the whole scope

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -0,0 +1,189 @@
...
+                existingCase.getId(),
+                exceptionRecord.id
+            );
+
+            return new ProcessResult(
+                exceptionRecordFinalizer.finalizeExceptionRecord(existingCase.getData(), existingCase.getId())
+            );
+        } catch (InvalidCaseDataException exception) {
+            if (BAD_REQUEST.equals(exception.getStatus())) {
+                throw new CallbackException("Failed to update case", exception);
```
lgonczar: The message for this case should be more like `Case Update API returned an error response with status {}.` Plus service name, exception record id and case ID

apereverzin: Will change

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -0,0 +1,189 @@
...
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+import uk.gov.hmcts.reform.ccd.client.model.Event;
+import uk.gov.hmcts.reform.ccd.client.model.StartEventResponse;
+
+import static java.lang.String.format;
+import static java.util.Collections.emptyList;
+import static org.springframework.http.HttpStatus.BAD_REQUEST;
+
+@Service
+public class CcdCaseUpdater {
```
lgonczar: Methods in this class are too long. Please extract smaller, subtask-specific ones, so that it reads better. We will need to do similar refactoring for case creation later, but `CcdCaseUpdater` is being introduced in current PR, so it should already be well split.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -0,0 +1,189 @@
...
+                return new ProcessResult(updateResponse.warnings, emptyList());
+            }
+
+            log.info(
+                "Successfully called case update endpoint of service {} to update case with case ID {} "
+                    + "based on exception record ref {}",
+                configItem.getService(),
+                existingCase.getId(),
+                exceptionRecord.id
+            );
```
lgonczar: This information should be logged earlier, before possible return.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -0,0 +1,189 @@
...
+            log.info(
+                "Successfully updated case for service {} with case ID {} based on exception record ref {}",
+                configItem.getService(),
+                existingCase.getId(),
+                exceptionRecord.id
+            );
+
+            return new ProcessResult(
+                exceptionRecordFinalizer.finalizeExceptionRecord(existingCase.getData(), existingCase.getId())
+            );
```
lgonczar: It would be ideal if this code was wrapped in an `else`. It reads nicest when we have some sort of a flow tree, rather than a single line with optional returns. We could also extract the selected bit into a separate method, called `updateCaseInCCD`, for example.

apereverzin: I have added 'else' clause. There is already 'updateCaseInCcd' method,, if we include log.info(...) calls into it we will need to provide service name as the value of a new parameter - but adding a new parameter just to log its value is not the best decision.

lgonczar: Wouldn't hurt much, in my opinion. Logging is an important part of what the app does, and this would be a private method.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
apereverzin: `if (!ignoreWarnings && !updateResponse.warnings.isEmpty())` - the method returns ProcessResult with warnings, it is not successful and should not be logged as successful. This is how it is done CcdNewCaseCreator and I agree with this approach. It just makes sense to log (inside `if (!ignoreWarnings && !updateResponse.warnings.isEmpty())`)  that `update` call returned warnings.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
apereverzin: ok

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
lgonczar: This is an operational log that help with tracing what happened. Response both with and without warnings is a success from Bulk Scan perspective (maybe not to the caseworker, but that's business level), i.e. everything worked as it should have - we successfully called the update endpoint. If we leave this log entry here, we won't have any trace of having called this endpoint.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -0,0 +1,238 @@
...
+
+    private final CaseUpdateClient caseUpdateClient;
+    private final ServiceResponseParser serviceResponseParser;
+    private final AuthTokenGenerator s2sTokenGenerator;
+    private final CoreCaseDataApi coreCaseDataApi;
+    private final ExceptionRecordFinalizer exceptionRecordFinalizer;
+
+    public CcdCaseUpdater(
+        CaseUpdateClient caseUpdateClient,
+        ServiceResponseParser serviceResponseParser, AuthTokenGenerator s2sTokenGenerator,
```
doncem: new line for third argument

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -0,0 +1,238 @@
...
+        CaseDetails existingCase,
+        String s2sToken
+    ) throws CaseClientServiceException {
+        return caseUpdateClient.updateCase(
+            configItem.getUpdateUrl(),
+            existingCase,
+            exceptionRecord,
+            s2sToken
+        );
+    }
```
doncem: don't need such a split - nothing else is happening and same arguments are passed

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
apereverzin: Will fix

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
apereverzin: Ok

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -0,0 +1,228 @@
...
+        ExceptionRecordFinalizer exceptionRecordFinalizer
+    ) {
+        this.caseUpdateClient = caseUpdateClient;
+        this.serviceResponseParser = serviceResponseParser;
+        this.s2sTokenGenerator = s2sTokenGenerator;
+        this.coreCaseDataApi = coreCaseDataApi;
+        this.exceptionRecordFinalizer = exceptionRecordFinalizer;
+    }
+
+    @SuppressWarnings({"squid:S2139", "unchecked"}) // squid for exception handle + logging
```
lgonczar: Why suppress `squid:S2139`? We shouldn't log the exception and then rethrow it, because that would result in two error log entries.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -0,0 +1,228 @@
...
+            log.info(
+                "Successfully called case update endpoint of service {} to update case with case Id {} "
+                    + "based on exception record ref {}",
+                configItem.getService(),
+                existingCase.getId(),
+                exceptionRecord.id
+            );
+
+            if (!ignoreWarnings && !updateResponse.warnings.isEmpty()) {
+                // do not log warnings
```
lgonczar: We can't log warnings, but we could log the fact that we returned warnings that wouldn't be ignored. Might be helpful.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -0,0 +1,228 @@
...
+                );
+
+                return new ProcessResult(
+                    exceptionRecordFinalizer.finalizeExceptionRecord(existingCase.getData(), existingCase.getId())
+                );
+            }
+        } catch (BadRequest exception) {
+            throw new CallbackException(
+                format(
+                    "Failed to call %s service Case Update API to update case with case Id %s "
```
lgonczar: This `catch` is for the try that wraps two calls, but we're assuming the exception concerns the first one. I think we should catch this exception closer to calling update endpoint (it doesn't have to be limited to `BadRequest` - can be the top-most exception for error response, because we're catching it to log a failure), and also handle `UnprocessableEntity` in that place instead of here.

lgonczar:
We need a discussion about what should happen during case update. In order to be safe, I think we should:
1. start case update event in CCD, which will also give us case details
2. call the update endpoint
3. submit the case update event and handle a 409 response from CCD
This approach guarantees no race conditions. There is one thing we haven't thought through here: we don't have event ID until we call the update endpoint. @nitinprabhu should we change the spec, so that we require an event of a specific name?

It looks like this PR assumes the following:
1. retrieve the case
2. call the update endpoint
3. start case update event in CCD
4. submit case update event in CCD
This flow allows for race conditions.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
apereverzin: I  have played a bit with different other options - unfortunately this try - catch layout looks to be the best out of other not ideal layouts.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
apereverzin: Will do

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    public ProcessResult updateCase(
+        ExceptionRecord exceptionRecord,
+        ServiceConfigItem configItem,
+        boolean ignoreWarnings,
+        String idamToken,
+        String userId,
+        String existingCaseId,
+        String eventId
```
lgonczar: We can remove eventId as the parameter and  hardcode it to `attachScannedDocsWithOcr`

apereverzin: Done

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -0,0 +1,239 @@
...
+                "Failed to update case for {} jurisdiction with case Id {} based on exception record with Id {}. "
+                    + "Service response: {}",
+                exceptionRecord.poBoxJurisdiction,
+                existingCase.getId(),
+                exceptionRecord.id,
+                exception.contentUTF8(),
+                exception
+            );
+
+            throw exception;
```
lgonczar: Why not throw an exception with the above message, but not have separate logging for it?

Also, we need to handle a 409 differently - that should result with a 200 response from us (see `ProcessResult` class), with an error saying that the case has been modified by someone else.

apereverzin: We are catching FeignException here specifically for logging because we can have access to its contentUTF8. I will remove catching Exception below because it is also caught on the level above.

Agree about 409, will do.

lgonczar: Why not throw an exception with that message instead?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -0,0 +1,239 @@
...
+        return Event
+            .builder()
+            .id(eventId)
+            .summary(format("Case updated, case Id %s", existingCaseId))
+            .description(
+                format(
+                    "Case with case Id %s updated based on exception record ref %s",
+                    existingCaseId,
+                    exceptionRecord.id
+                )
```
lgonczar: The event is going to be created in the context of "this" case - we'll see it in the history of the case. Therefore, we don't need to include the case ID. Exception record ID is fine in the message.

apereverzin: Will do

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
apereverzin: Will do

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
rkondratowicz: let's rely on automated code quality on this one

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
rkondratowicz: It's not just about the layout @apereverzin.
`coreCaseDataApi.startEventForCaseWorker(...)` from line 73 can also throw `BadRequest` and then the message in this catch will be incorrect.

apereverzin: Have fixed

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
lgonczar: There's only so much code quality tools can tell.

PR #794, opened 2019-12-10
Create Exception record when ccd returns 404 response for supplementary evidence

aliveni:
https://tools.hmcts.net/jira/browse/BPS-932

### Change description ###
Create an Exception record when attaching scanned documents to the case fails.
CCD returns 404 response in 2 scenarios:
- **Start event** - returns 404 if `attachScannedDocs` event is not configured for the case type.
- **Submit event** - returns 404 if the caseworker does not have access to the`attachScannedDocs` event.

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/SupplementaryEvidenceCreatorTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/SupplementaryEvidenceCreatorTest.java
@@ -1,26 +1,39 @@
 package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.events;
 
+import com.github.tomakehurst.wiremock.client.MappingBuilder;
 import com.github.tomakehurst.wiremock.client.WireMock;
 import com.microsoft.azure.servicebus.IMessageReceiver;
 import com.microsoft.azure.servicebus.Message;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.mock.mockito.SpyBean;
+import org.springframework.http.HttpStatus;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.config.Environment;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.config.IntegrationTest;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.domains.envelopes.EnvelopeEventProcessor;
 
 import java.util.concurrent.TimeUnit;
 
 import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
+import static com.github.tomakehurst.wiremock.client.WireMock.containing;
 import static com.github.tomakehurst.wiremock.client.WireMock.get;
 import static com.github.tomakehurst.wiremock.client.WireMock.givenThat;
+import static com.github.tomakehurst.wiremock.client.WireMock.post;
 import static com.github.tomakehurst.wiremock.client.WireMock.postRequestedFor;
 import static com.github.tomakehurst.wiremock.client.WireMock.urlPathEqualTo;
+import static org.apache.http.HttpHeaders.AUTHORIZATION;
 import static org.awaitility.Awaitility.await;
 import static org.mockito.BDDMockito.given;
 import static uk.gov.hmcts.reform.bulkscan.orchestrator.SampleData.fileContentAsString;
 import static uk.gov.hmcts.reform.bulkscan.orchestrator.config.Environment.CASE_EVENT_URL;
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.config.Environment.CASE_REF;
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.config.Environment.CASE_SEARCH_URL;
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.config.Environment.CASE_SUBMIT_URL;
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.config.Environment.CASE_TYPE_BULK_SCAN;
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.config.Environment.CASE_TYPE_EXCEPTION_RECORD;
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.config.Environment.EXCEPTION_RECORD_SEARCH_URL;
 import static uk.gov.hmcts.reform.bulkscan.orchestrator.config.Environment.GET_CASE_URL;
 
 @IntegrationTest
@@ -37,14 +50,30 @@ class SupplementaryEvidenceCreatorTest {
     @Autowired
     private EnvelopeEventProcessor envelopeEventProcessor;
 
+    private static final String CREATE_ExCEPTION_RECORD_SUBMIT_URL = Environment.CASE_SUBMIT_URL
+        .replace(CASE_TYPE_BULK_SCAN, CASE_TYPE_EXCEPTION_RECORD);
+
+    private static final String CREATE_EXCEPTION_START_EVENT_URL = Environment.CASE_EVENT_TRIGGER_START_URL
+        .replace(CASE_TYPE_BULK_SCAN, CASE_TYPE_EXCEPTION_RECORD);
+
+    private static final String SERVICE_AUTHORIZATION_HEADER = "ServiceAuthorization";
+
+    // see WireMock mapping json files
+    private static final String MOCKED_IDAM_TOKEN_SIG = "q6hDG0Z1Qbinwtl8TgeDrAVV0LlCTRtbQqBYoMjd03k";
+    private static final String MOCKED_S2S_TOKEN_SIG =
+        "X1-LdZAd5YgGFP16-dQrpqEICqRmcu1zL_zeCLyUqMjb5DVx7xoU-r8yXHfgd4tmmjGqbsBz_kLqgu8yruSbtg";
+
+    private static final String ELASTICSEARCH_EMPTY_RESPONSE = "{\"total\": 0,\"cases\": []}";
+
+    @BeforeEach
+    void before() throws Exception {
+        givenThat(get(GET_CASE_URL).willReturn(aResponse().withBody(MOCK_RESPONSE)));
+        given(messageReceiver.receive()).willReturn(MOCK_MESSAGE).willReturn(null);
+    }
 
     @DisplayName("Should call ccd to attach supplementary evidence for caseworker")
     @Test
     void should_call_ccd_to_attach_supplementary_evidence_for_caseworker() throws Exception {
-        // given
-        givenThat(get(GET_CASE_URL).willReturn(aResponse().withBody(MOCK_RESPONSE)));
-        given(messageReceiver.receive()).willReturn(MOCK_MESSAGE).willReturn(null);
-
         // when
         envelopeEventProcessor.processNextMessage();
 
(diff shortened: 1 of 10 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/SupplementaryEvidenceCreatorTest.java
```diff
@@ -37,26 +49,115 @@
...
+
+        // then
+        await()
+            .atMost(60, TimeUnit.SECONDS)
+            .pollInterval(2, TimeUnit.SECONDS)
+            .ignoreExceptions()
+            .until(() -> {
+                WireMock.verify(postRequestedFor(urlPathEqualTo(Environment.CASE_EVENT_URL)));
+
+                return true;
```
rkondratowicz: will `.until` try again if `verify` fails? Does `verify` throw an exception?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdApi.java
```diff
@@ -253,29 +254,51 @@ public StartEventResponse startEvent(
...
+                authenticator.getServiceToken(),
+                authenticator.getUserDetails().getId(),
+                jurisdiction,
+                caseTypeId,
+                caseRef,
+                eventTypeId
+            );
+        } catch (FeignException e) {
+            if (e.status() == 404) {
+                throw new UnableToAttachDocumentsException(
```
rkondratowicz: the method is called `startEventForExistingCase` and it can have nothing to do with attaching documents

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/SupplementaryEvidenceCreatorTest.java
aliveni: Yes it tries every 2 seconds for max 2 minutes.
`.atMost(60, TimeUnit.SECONDS)
            .pollInterval(2, TimeUnit.SECONDS)
`

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdApi.java
aliveni: This is called from https://github.com/hmcts/bulk-scan-orchestrator/blob/28d55bfd9c583dc96a71f0048c07359e60f073f6/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/AttachDocsToSupplementaryEvidence.java#L54

If the CCD event id (`attachScannedDocs`) is not configured for the case type, CCD will return 404 response.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/SupplementaryEvidenceCreatorTest.java
rkondratowicz: I mean, will the test not fail immediately due to `WireMock.verify(postRequestedFor(urlPathEqualTo(Environment.CASE_EVENT_URL)));` throwing an exception? does it handle code throwing exceptions or just predicate (true/false)

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdApi.java
rkondratowicz: Doesn't matter where it's called from **right now**.
The method name has nothing to do with attaching docs.
You can rename it.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/SupplementaryEvidenceCreatorTest.java
aliveni: It will fail if the verification fails

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/AttachDocsToSupplementaryEvidence.java
```diff
@@ -48,37 +48,48 @@ public void attach(Envelope envelope, CaseDetails existingCase) {
...
+                    envelope.id
+                );
+            } catch (UnableToAttachDocumentsException e) {
+                log.error(
+                    "Failed to attach documents from envelope to case. Case Type: {}, Case ID: {}, envelope ID: {}",
+                    existingCase.getCaseTypeId(),
+                    existingCase.getId(),
+                    envelope.id
+                );
+                return false;
```
rkondratowicz: is there a benefit of replacing throwing an exception with boolean response?
imho it would be clearer tho handle this exception where this method is called.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdApi.java
aliveni: renamed the method names for both start and submit events

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdApi.java
```diff
@@ -253,29 +254,51 @@ public StartEventResponse startEvent(
...
+                authenticator.getUserToken(),
+                authenticator.getServiceToken(),
+                authenticator.getUserDetails().getId(),
+                jurisdiction,
+                caseTypeId,
+                caseRef,
+                eventTypeId
+            );
+        } catch (FeignException e) {
+            if (e.status() == 404) {
```
lgonczar: We could specifically catch `FeignException.NotFound` instead, and `FeignException` in a subsequent catch. Up to you.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EnvelopeHandler.java
```diff
@@ -35,12 +35,15 @@ public void handleEnvelope(Envelope envelope) {
 
                 if (caseDetailsFound.isPresent()) {
                     CaseDetails existingCase = caseDetailsFound.get();
-                    evidenceAttacher.attach(envelope, existingCase);
-                    paymentsProcessor.createPayments(envelope, existingCase.getId(), false);
+                    boolean docsAttached = evidenceAttacher.attach(envelope, existingCase);
+                    if (docsAttached) {
+                        paymentsProcessor.createPayments(envelope, existingCase.getId(), false);
```
lgonczar: Should payment creation be unconditional, i.e. happen in both cases?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/AttachDocsToSupplementaryEvidence.java
lgonczar: There's some elegance in this solution. If we don't need to control the flow using exceptions, I think best not to. Using this method is quite straightforward as it is. As long as it returns a result, we always need to continue the processing of the message as if nothing bad happened. If it throws, that's a message processing error. Not that handling this exception in the calling class would be something incorrect, but I find this way a bit nicer - it leaves the calling code very clear.

It would be worth explaining in javadoc what is returned here.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EnvelopeHandler.java
nitinprabhu: If we don't attach scanned docs then we should not update payments as the caseworker would not be aware or may need documentation corresponding to payments(for e.g evidence for change of name or address)

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdApi.java
aliveni: updated in both start and submit events

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EnvelopeHandler.java
aliveni: After creating the exception record, the payment message is published with the exception record id. Is this ok?

lgonczar: Ok, there's some lack of clarity here. From the code `docsAttached` is `false` if we failed to attach them to the case, getting a 404 response. If there were no documents to be attached, the method will return `true`.

I guess we have too little information coming from this method, because:
- if nothing failed, but there were no docs attached, we shouldn't create payments
- if at least one doc was attached, we should create payments
- if we failed because a 404 from CCD, we should create an exception record, but also payments

Is this what we should expect?

aliveni: I missed the `no documents to attach` case.  I don't think we have to create an Exception record when there are no documents to attach.

nitinprabhu: Exception handling way would be better imo way Rafal suggested

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/SupplementaryEvidenceCreatorTest.java
rkondratowicz: so it will not try again if first try fails?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EnvelopeHandler.java
lgonczar: > If we don't attach scanned docs then I think it is better we don't update payments as the caseworker would not be aware or may need documentation corresponding to payments(for e.g evidence for change of name or address)

How about the case when we previously attached supplementary case from this envelope to the case and failed to publish payments? The envelope returned to the queue and is processed again now. This time, we won't attach any documents to the case, as it has already happened, but we still haven't published payments. Given payment creation is idempotent in PayHub, so publishing them twice wouldn't cause any harm.

nitinprabhu: If we have attached the docs or skipped attaching because we have already attached previously then we should process the payments. Only when we fail to attach due to access issue or configuration issue(404/403) we should create an exception record and publish payments.
Transaction T1 -> Attach scanned docs or create exception record
Transaction T2->  Process payments 

Also the way T2 would work is also impacted what you do in T1 as we need to publish different payment messages.

lgonczar: So as long as we're not having a general failure resulting in not completing the envelope, we should always publish payments. Is that correct?

nitinprabhu: yes correct. Also we will have to keep in mind idempotency scenario for exception record creation(probably we can handle it as a separate PR else it will make this PR complicated)

lgonczar: Actually, the code already works as expected.

aliveni: > yes correct. Also we will have to keep in mind idempotency scenario for exception record creation(probably we can handle it as a separate PR else it will make this PR complicated)

@nitinprabhu As discussed earlier, we are already checking if an Exception record was created for an envelope
https://github.com/hmcts/bulk-scan-orchestrator/blob/28d55bfd9c583dc96a71f0048c07359e60f073f6/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/CreateExceptionRecord.java#L48

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/SupplementaryEvidenceCreatorTest.java
aliveni: discussed in person

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EnvelopeHandler.java
```diff
@@ -35,12 +35,15 @@ public void handleEnvelope(Envelope envelope) {
...
+                    } else {
+                        createExceptionRecord(envelope);
```
rkondratowicz: imho worth logging here why this happened as this is different from regular path to ER creation

PR #795, opened 2019-12-10
Return 500 status for internal errors

rkondratowicz:
...  when attaching ER to a case

https://tools.hmcts.net/jira/browse/BPS-934

Surprisingly both exception handler and unit test for the controller didn't exist until now :/

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/ResponseExceptionHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/ResponseExceptionHandler.java
@@ -18,7 +18,8 @@ public class ResponseExceptionHandler extends ResponseEntityExceptionHandler {
 
     @ExceptionHandler(Exception.class)
     protected ResponseEntity<ErrorResponse> handleInternalException(Exception exception) {
-        log.error(exception.getMessage(), exception);
+        log.error("Unhandled exception. Returning 500 response to client", exception);
+
         return status(INTERNAL_SERVER_ERROR).body(new ErrorResponse(exception.getMessage()));
     }
 
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/ResponseExceptionHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/ResponseExceptionHandler.java
@@ -1,5 +1,6 @@
 package uk.gov.hmcts.reform.bulkscan.orchestrator.controllers;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.http.ResponseEntity;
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/ResponseExceptionHandler.java
```diff
@@ -0,0 +1,23 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler;
+
+import static org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR;
+import static org.springframework.http.ResponseEntity.status;
+
+@ControllerAdvice
+public class ResponseExceptionHandler extends ResponseEntityExceptionHandler {
```
lgonczar: 👍

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/ResponseExceptionHandler.java
```diff
@@ -0,0 +1,23 @@
...
+
+    private static final Logger log = LoggerFactory.getLogger(ResponseExceptionHandler.class);
+
+    @ExceptionHandler(Exception.class)
+    protected ResponseEntity<Void> handleInternalException(Exception exception) {
+        log.error(exception.getMessage(), exception);
```
lgonczar: Instead of `exception.getMessage()` we could have a message saying that an error occurred when serving a request.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/ResponseExceptionHandler.java
```diff
@@ -0,0 +1,34 @@
...
+
+@ControllerAdvice
+public class ResponseExceptionHandler extends ResponseEntityExceptionHandler {
+
+    private static final Logger log = LoggerFactory.getLogger(ResponseExceptionHandler.class);
+
+    @ExceptionHandler(Exception.class)
+    protected ResponseEntity<ErrorResponse> handleInternalException(Exception exception) {
+        log.error(exception.getMessage(), exception);
+        return status(INTERNAL_SERVER_ERROR).body(new ErrorResponse(exception.getMessage()));
```
rkondratowicz: Some existing tests (for other endpoints) assume 500 responses have body.

lgonczar: I think it's best to either return a generic failure message or not return anything at all (and adjust tests). The caller shouldn't be informed about the details of our internal errors.

lgonczar: As discussed, this PR doesn't change the behaviour, as far as the response body is concerned. This can be addressed at a different point.

PR #796, opened 2019-12-10
Create log entry after starting an event in CCD

lgonczar:
### Change description ###

Create log entry after starting an event in CCD, so that it's easier to trace what exactly went wrong in production.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
@@ -79,6 +79,16 @@ public class CcdCaseUpdater {
                 EVENT_ID_ATTACH_SCANNED_DOCS_WITH_OCR
             );
 
+            log.info(
+                "Started CCD event to update case. "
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -79,6 +79,15 @@ public ProcessResult updateCase(
                 EVENT_ID_ATTACH_SCANNED_DOCS_WITH_OCR
             );
 
+            log.info(
+                "Started event {} for updating case {} from exception record {}. Case type: {}. Case state: {}",
```
rkondratowicz: maybe just stick to the format used for the second part of the log. No need to build a sentence imho

lgonczar: What format string would you propose?

rkondratowicz: Started CCD event. EventId: {}. Case id: {}. Exception record id: {}. Case type: {}. Case state: {}

PR #801, opened 2019-12-12
Configure case type IDs for services (duplicate prevention)

lgonczar:
### Change description ###

Configure case type IDs for services, so that duplicate case prevention can work.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -57,8 +57,8 @@ idam:
 service-config:
   # Services and all their case type IDs, excluding exception records.
   # These mappings are used by Bulk Scan to determine which case type IDs to search across
-  # when looking for a case by its legacy ID. Those CTIDs that are not present here, will
-  # not be included in the search.
+  # when looking for a service case (by its legacy ID, ID of related exception record, etc).
+  # Those CTIDs that are not present here, will not be included in the search.
   services:
     - service: bulkscan # service name corresponds to the name of the container in Blob Storage
       jurisdiction: BULKSCAN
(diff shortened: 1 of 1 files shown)


src/main/resources/application.yaml
```diff
@@ -75,16 +75,24 @@ service-config:
...
+        - StandingSearch
+        - WillLodgement
       allow-creating-case-before-payments-are-processed: true
 
     - service: divorce
       jurisdiction: DIVORCE
-      case-type-ids: []
+      case-type-ids:
+        - DIVORCE
+        - DIVORCE_NOTICE_OF_ACTING
```
lgonczar: This one is to be confirmed by Divorce (they're going to come back to us), but for now it's better to stay on the safe side and include it.

lgonczar:
> could you please add change history ?

What do you mean?

src/main/resources/application.yaml
```diff
@@ -75,16 +75,24 @@ service-config:
     - service: probate
       jurisdiction: PROBATE
       transformation-url: ${TRANSFORMATION_URL_PROBATE}
-      case-type-ids: []
+      case-type-ids:
+        - GrantOfRepresentation
+        - Caveat
+        - StandingSearch
+        - WillLodgement
```
rkondratowicz: what are `case-type-ids` and what does an empty list mean

lgonczar: There is a comment on the whole configuration section. I've updated it. Not introducing anything in this PR, just concrete values.

PR #803, opened 2019-12-12
Calling CcdCaseUpdater

apereverzin:

### Change description ###
Calling CcdCaseUpdater from AttachCaseCallbackService.

This PR include controller class CcdCallbackController, service class AttachCaseCallbackService, a unit test for the controller, an integration test and a functional test (which just became enabled in this PR).

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/AttachExceptionRecordWithOcrTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/AttachExceptionRecordWithOcrTest.java
@@ -1,9 +1,11 @@
 package uk.gov.hmcts.reform.bulkscan.orchestrator.controllers;
 
+import com.github.tomakehurst.wiremock.client.ResponseDefinitionBuilder;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Sets;
 import io.restassured.path.json.JsonPath;
 import io.restassured.response.ValidatableResponse;
+import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 import org.springframework.boot.web.server.LocalServerPort;
 import org.springframework.http.HttpHeaders;
@@ -18,13 +20,16 @@ import java.util.Set;
 import static com.github.tomakehurst.wiremock.client.WireMock.containing;
 import static com.github.tomakehurst.wiremock.client.WireMock.get;
 import static com.github.tomakehurst.wiremock.client.WireMock.givenThat;
+import static com.github.tomakehurst.wiremock.client.WireMock.notFound;
 import static com.github.tomakehurst.wiremock.client.WireMock.okJson;
 import static com.github.tomakehurst.wiremock.client.WireMock.post;
+import static com.github.tomakehurst.wiremock.client.WireMock.serverError;
 import static com.google.common.io.Resources.getResource;
 import static com.google.common.io.Resources.toByteArray;
 import static io.restassured.RestAssured.given;
 import static java.util.stream.Collectors.toSet;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR;
 import static org.springframework.http.HttpStatus.OK;
 import static uk.gov.hmcts.reform.bulkscan.orchestrator.config.Environment.CASE_REF;
 
@@ -58,15 +63,12 @@ class AttachExceptionRecordWithOcrTest {
     @LocalServerPort
     int serverPort;
 
+    @DisplayName("Should successfully callback with correct information")
     @Test
-    void should_update_case_with_ocr_data() throws Exception {
+    void should_update_case_with_ocr_data() {
         setUpClientUpdate(getResponseBody("client-update-ok-no-warnings.json"));
-        setUpCcdUpdateCaseEvents(
-            getResponseBody("ccd-start-event-for-case-worker.json"),
-            getResponseBody("ccd-submit-event-for-case-worker.json")
-        );
-
-        byte[] requestBody = getRequestBody("valid-supplementary-evidence-with-ocr.json");
+        ccdStartEvent(okJson(getResponseBody("ccd-start-event-for-case-worker.json")));
+        ccdSubmitEvent(okJson(getResponseBody("ccd-submit-event-for-case-worker.json")));
 
         Map<String, Object> caseData = new HashMap<>();
         caseData.put("poBox","PO 12345");
@@ -74,10 +76,46 @@ class AttachExceptionRecordWithOcrTest {
         caseData.put("formType","B123");
         caseData.put("attachToCaseReference","1539007368674134");
 
+        byte[] requestBody = getRequestBody("valid-supplementary-evidence-with-ocr.json");
+
         ValidatableResponse response = postWithBody(requestBody)
             .statusCode(OK.value());
 
         verifySuccessResponse(response, caseData);
+    }
+
+    @DisplayName("Should fail with the correct error when submit api call fails")
+    @Test
+    public void should_fail_with_the_correct_error_when_submit_api_call_fails() {
+        setUpClientUpdate(getResponseBody("client-update-ok-no-warnings.json"));
+        ccdStartEvent(okJson(getResponseBody("ccd-start-event-for-case-worker.json")));
+        ccdSubmitEvent(serverError());
+
+        byte[] requestBody = getRequestBody("valid-supplementary-evidence-with-ocr.json");
+
+        postWithBody(requestBody).statusCode(INTERNAL_SERVER_ERROR.value());
+    }
+
+    @DisplayName("Should fail with the correct error when start event api call fails")
+    @Test
+    public void should_fail_with_the_correct_error_when_start_event_api_call_fails() {
+        setUpClientUpdate(getResponseBody("client-update-ok-no-warnings.json"));
+        ccdStartEvent(notFound());
+
+        byte[] requestBody = getRequestBody("valid-supplementary-evidence-with-ocr.json");
+
+        postWithBody(requestBody).statusCode(INTERNAL_SERVER_ERROR.value());
+    }
+
+    @DisplayName("Should fail correctly if ccd is down")
+    @Test
+    public void should_fail_correctly_if_ccd_is_down() {
+        setUpClientUpdate(getResponseBody("client-update-ok-no-warnings.json"));
+        ccdStartEvent(serverError());
+
+        byte[] requestBody = getRequestBody("valid-supplementary-evidence-with-ocr.json");
+
+        postWithBody(requestBody).statusCode(INTERNAL_SERVER_ERROR.value());
     }
 
     private void verifySuccessResponse(
@@ -135,7 +173,7 @@ class AttachExceptionRecordWithOcrTest {
         );
     }
 
-    private void setUpCcdUpdateCaseEvents(String startResponseBody, String submitResponseBody) {
+    private void ccdStartEvent(ResponseDefinitionBuilder response) {
         givenThat(
             get(
               "/caseworkers/" + USER_ID
@@ -146,9 +184,11 @@ class AttachExceptionRecordWithOcrTest {
                 + "/token"
             )
                 .withHeader("ServiceAuthorization", containing("Bearer"))
-                .willReturn(okJson(startResponseBody))
+                .willReturn(response)
         );
+    }
 
+    private void ccdSubmitEvent(ResponseDefinitionBuilder response) {
         givenThat(
             post(
                 // values from config + initial request body
@@ -158,7 +198,7 @@ class AttachExceptionRecordWithOcrTest {
                     + "/cases?ignore-warning=true"
             )
                 .withHeader("ServiceAuthorization", containing("Bearer"))
-                .willReturn(okJson(submitResponseBody))
+                .willReturn(response)
         );
     }
 
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/AttachExceptionRecordWithOcrTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/AttachExceptionRecordWithOcrTest.java
@@ -1,0 +1,200 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.controllers;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Sets;
+import io.restassured.path.json.JsonPath;
+import io.restassured.response.ValidatableResponse;
+import org.junit.jupiter.api.Test;
+import org.springframework.boot.web.server.LocalServerPort;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.MediaType;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.config.IntegrationTest;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.containing;
+import static com.github.tomakehurst.wiremock.client.WireMock.get;
+import static com.github.tomakehurst.wiremock.client.WireMock.givenThat;
+import static com.github.tomakehurst.wiremock.client.WireMock.okJson;
+import static com.github.tomakehurst.wiremock.client.WireMock.post;
+import static com.google.common.io.Resources.getResource;
+import static com.google.common.io.Resources.toByteArray;
+import static io.restassured.RestAssured.given;
+import static java.util.stream.Collectors.toSet;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.springframework.http.HttpStatus.OK;
+import static uk.gov.hmcts.reform.bulkscan.orchestrator.config.Environment.CASE_REF;
+
+@IntegrationTest
+class AttachExceptionRecordWithOcrTest {
+
+    // event id used for service specific case creation
+    private static final String EVENT_ID = "attachScannedDocsWithOcr";
+
+    private static final String IDAM_TOKEN = "idam-token";
+    private static final String USER_ID = "640";
+
+    private static final String SERVICE_AUTHORIZATION_HEADER = "ServiceAuthorization";
+    private static final String RESPONSE_FIELD_ERRORS = "errors";
+    private static final String RESPONSE_FIELD_WARNINGS = "warnings";
+    private static final String RESPONSE_FIELD_DATA = "data";
+    private static final String EVENT_ID_ATTACH_TO_CASE = "attachToExistingCase";
+    private static final String CLASSIFICATION_EXCEPTION = "EXCEPTION";
+    private static final String CALLBACK_ATTACH_CASE_PATH = "/callback/attach_case";
+    private static final String ATTACH_TO_CASE_REFERENCE_FIELD_NAME = "attachToCaseReference";
+
+    private static final String CASE_REFERENCE_TYPE_CCD = "ccdCaseReference";
+    private static final long EXCEPTION_RECORD_ID = 26409983479785245L;
+    private static final String EXCEPTION_RECORD_FILENAME = "record.pdf";
+    private static final String EXCEPTION_RECORD_DOCUMENT_NUMBER = "654321";
+    private static final Map<String, Object> EXCEPTION_RECORD_DOC = document(
+        EXCEPTION_RECORD_FILENAME,
+        EXCEPTION_RECORD_DOCUMENT_NUMBER
+    );
+
+    @LocalServerPort
+    int serverPort;
+
+    @Test
+    void should_update_case_with_ocr_data() throws Exception {
+        setUpClientUpdate(getResponseBody("client-update-ok-no-warnings.json"));
+        setUpCcdUpdateCaseEvents(
+            getResponseBody("ccd-start-event-for-case-worker.json"),
+            getResponseBody("ccd-submit-event-for-case-worker.json")
+        );
+
+        byte[] requestBody = getRequestBody("valid-supplementary-evidence-with-ocr.json");
+
+        Map<String, Object> caseData = new HashMap<>();
+        caseData.put("poBox","PO 12345");
+        caseData.put("journeyClassification","SUPPLEMENTARY_EVIDENCE_WITH_OCR");
+        caseData.put("formType","B123");
+        caseData.put("attachToCaseReference","1539007368674134");
+
+        ValidatableResponse response = postWithBody(requestBody)
+            .statusCode(OK.value());
+
+        verifySuccessResponse(response, caseData);
+    }
+
+    private void verifySuccessResponse(
+        ValidatableResponse response,
+        Map<String, Object> requestData
+    ) {
+        JsonPath responseJson = response.extract().jsonPath();
+
+        assertThat(responseJson.getList(RESPONSE_FIELD_ERRORS)).isNullOrEmpty();
+        assertThat(responseJson.getList(RESPONSE_FIELD_WARNINGS)).isNullOrEmpty();
+
+        Map<String, Object> responseData = responseJson.getMap(RESPONSE_FIELD_DATA);
+        assertThat(responseData).isNotNull();
+        assertThat(responseData.get(ATTACH_TO_CASE_REFERENCE_FIELD_NAME)).isEqualTo(CASE_REF);
+
+        assertMapsAreEqualIgnoringFields(
+            responseData,
+            requestData,
+            ATTACH_TO_CASE_REFERENCE_FIELD_NAME,
+            "deliveryDate",
+            "openingDate",
+            "scannedDocuments",
+            "scanOCRData"
+        );
+    }
+
+    private void assertMapsAreEqualIgnoringFields(
+        Map<String, Object> actual,
+        Map<String, Object> expected,
+        String... fieldsToIgnore
+    ) {
+        Set<String> ignoredFieldSet = Sets.newHashSet(fieldsToIgnore);
+
+        Set<Map.Entry<String, Object>> actualWithoutIgnoredFields = getFilteredFieldSet(actual, ignoredFieldSet);
+        Set<Map.Entry<String, Object>> expectedWithoutIgnoredFields = getFilteredFieldSet(expected, ignoredFieldSet);
+
+        assertThat(actualWithoutIgnoredFields).hasSameElementsAs(expectedWithoutIgnoredFields);
+    }
+
+    private Set<Map.Entry<String, Object>> getFilteredFieldSet(
+        Map<String, Object> fieldMap,
+        Set<String> fieldsToExclude
+    ) {
+        return fieldMap
+            .entrySet()
+            .stream()
+            .filter(e -> !fieldsToExclude.contains(e.getKey()))
+            .collect(toSet());
+    }
+
+    private void setUpClientUpdate(String responseBody) {
+        givenThat(post("/update-case")
+            .withHeader("ServiceAuthorization", containing("Bearer"))
+            .willReturn(okJson(responseBody))
+        );
+    }
+
+    private void setUpCcdUpdateCaseEvents(String startResponseBody, String submitResponseBody) {
+        givenThat(
+            get(
+              "/caseworkers/" + USER_ID
+                + "/jurisdictions/BULKSCAN"
+                + "/case-types/BULKSCAN_ExceptionRecord"
+                + "/cases/1539007368674134"
+                + "/event-triggers/" + EVENT_ID
+                + "/token"
+            )
+                .withHeader("ServiceAuthorization", containing("Bearer"))
+                .willReturn(okJson(startResponseBody))
+        );
+
+        givenThat(
+            post(
+                // values from config + initial request body
+                "/caseworkers/" + USER_ID
+                    + "/jurisdictions/BULKSCAN"
+                    + "/case-types/BULKSCAN_ExceptionRecord"
+                    + "/cases?ignore-warning=true"
+            )
+                .withHeader("ServiceAuthorization", containing("Bearer"))
+                .willReturn(okJson(submitResponseBody))
+        );
+    }
+
+    private byte[] getFileContents(String ccdCallbackSubFolders, String filename) {
+        try {
+            return toByteArray(getResource("ccd/callback/attach" + ccdCallbackSubFolders + filename));
+        } catch (IOException exception) {
+            throw new RuntimeException(exception);
+        }
+    }
+
+    private byte[] getRequestBody(String filename) {
+        return getFileContents("/request/", filename);
+    }
+
+    private String getResponseBody(String filename) {
+        return new String(getFileContents("/response/", filename));
+    }
+
+    private ValidatableResponse postWithBody(byte[] body) {
+        return given()
+            .header(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE)
+            .header(HttpHeaders.AUTHORIZATION, IDAM_TOKEN)
+            .header(CcdCallbackController.USER_ID, USER_ID)
+            .body(body)
+            .post("http://localhost:" + serverPort + "/callback/attach_case")
+            .then();
+    }
+
+    private static Map<String, Object> document(String filename, String documentNumber) {
+        return ImmutableMap.of(
+            "value", ImmutableMap.of(
+                "fileName", filename,
+                "controlNumber", documentNumber,
+                "someNumber", 3
+            )
+        );
+    }
+}
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/AttachExceptionRecordWithOcrTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/AttachExceptionRecordWithOcrTest.java
@@ -210,7 +210,8 @@ class AttachExceptionRecordWithOcrTest {
                 "/caseworkers/" + USER_ID
                     + "/jurisdictions/BULKSCAN"
                     + "/case-types/BULKSCAN_ExceptionRecord"
-                    + "/cases?ignore-warning=true"
+                    + "/cases/1539007368674134"
+                    + "/events?ignore-warning=true"
             )
                 .withHeader("ServiceAuthorization", containing("Bearer"))
                 .willReturn(response)
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/AttachExceptionRecordWithOcrTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/AttachExceptionRecordWithOcrTest.java
@@ -43,23 +43,10 @@ class AttachExceptionRecordWithOcrTest {
     private static final String IDAM_TOKEN = "idam-token";
     private static final String USER_ID = "640";
 
-    private static final String SERVICE_AUTHORIZATION_HEADER = "ServiceAuthorization";
     private static final String RESPONSE_FIELD_ERRORS = "errors";
     private static final String RESPONSE_FIELD_WARNINGS = "warnings";
     private static final String RESPONSE_FIELD_DATA = "data";
-    private static final String EVENT_ID_ATTACH_TO_CASE = "attachToExistingCase";
-    private static final String CLASSIFICATION_EXCEPTION = "EXCEPTION";
-    private static final String CALLBACK_ATTACH_CASE_PATH = "/callback/attach_case";
     private static final String ATTACH_TO_CASE_REFERENCE_FIELD_NAME = "attachToCaseReference";
-
-    private static final String CASE_REFERENCE_TYPE_CCD = "ccdCaseReference";
-    private static final long EXCEPTION_RECORD_ID = 26409983479785245L;
-    private static final String EXCEPTION_RECORD_FILENAME = "record.pdf";
-    private static final String EXCEPTION_RECORD_DOCUMENT_NUMBER = "654321";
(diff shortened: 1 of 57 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -283,6 +284,42 @@ static boolean hasOcr(CaseDetails theCase) {
...
+            .orElse(invalid("Missing journeyClassification"));
+    }
+
+    private static Validation<String, Classification> validateClassificationForAttachToCase(
+        Classification classification,
+        CaseDetails theCase
+    ) {
+        if (!EXCEPTION.equals(classification)
+            && !SUPPLEMENTARY_EVIDENCE.equals(classification)
+            && !SUPPLEMENTARY_EVIDENCE_WITH_OCR.equals(classification)) {
```
rkondratowicz: maybe `if (!asList(EXCEPTION, EXCEPTION, SUPPLEMENTARY_EVIDENCE_WITH_OCR).contains(classification))`?
or even
```java
var validClassifications = asList(EXCEPTION, EXCEPTION, SUPPLEMENTARY_EVIDENCE_WITH_OCR);

...

if (!validClassifications.contains(classification) {
    ...
}
```

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -257,48 +292,65 @@ private void attachCaseByCcdId(
...
         log.info("Attaching exception record '{}' to a case by CCD ID '{}'", exceptionRecordId, targetCaseCcdRef);
 
-        CaseDetails theCase = ccdApi.getCase(targetCaseCcdRef, exceptionRecordJurisdiction);
-        List<Map<String, Object>> targetCaseDocuments = getScannedDocuments(theCase);
-
-        verifyExceptionRecordIsNotAttachedToCase(exceptionRecordJurisdiction, exceptionRecordId);
+        Validation<String, Classification> classificationValidation =
+            hasJourneyClassificationForAttachToCase(exceptionRecordData);
+        if (classificationValidation.isValid()) {
+            if (SUPPLEMENTARY_EVIDENCE_WITH_OCR == classificationValidation.get()) {
```
lgonczar: This comparison can't be done like this. You need to use `equals` instead.

rkondratowicz: `if (SUPPLEMENTARY_EVIDENCE_WITH_OCR == classificationValidation.get())` seems much more natural

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -419,4 +471,14 @@ public AttachToCaseEventData(
...
+
+    private ServiceConfigItem getServiceConfig(CaseDetails caseDetails) {
+        return hasServiceNameInCaseTypeId(caseDetails).flatMap(service -> Try
+            .of(() -> serviceConfigProvider.getConfig(service))
+            .toValidation()
+            .mapError(Throwable::getMessage)
+        )
+            .filter(item -> !com.google.common.base.Strings.isNullOrEmpty(item.getUpdateUrl()))
+            .getOrElseThrow(() -> new CallbackException("Update URL is not configured")).get();
+    }
```
rkondratowicz: maybe it would be clearer to simply check whether case contains service name and **then** get url for it,
instead of doing flatmaps, converting trys to validations etc

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -283,6 +284,42 @@ static boolean hasOcr(CaseDetails theCase) {
         return valid(classification);
     }
 
+    public static Validation<String, Classification> hasJourneyClassificationForAttachToCase(CaseDetails theCase) {
```
lgonczar: Do we do anything with validation errors from this method?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -283,6 +284,42 @@ static boolean hasOcr(CaseDetails theCase) {
...
+
+        return classificationOption
+            .map(classification -> Try.of(() -> Classification.valueOf(classification)))
+            .map(Try::toValidation)
+            .map(validation -> validation
+                .mapError(throwable -> "Invalid journeyClassification. Error: " + throwable.getMessage())
+                .flatMap(classification -> validateClassificationForAttachToCase(classification, theCase))
+            )
+            .orElse(invalid("Missing journeyClassification"));
+    }
```
lgonczar: Really hard to follow

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
rkondratowicz: it's an enum, you can use `==`

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
apereverzin: Will fix

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
apereverzin: Will do

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/AttachExceptionRecordWithOcrTest.java
```diff
@@ -0,0 +1,240 @@
...
+    void should_update_case_with_ocr_data() {
+        setUpClientUpdate(getResponseBody("client-update-ok-no-warnings.json"));
+        ccdStartEvent(okJson(getResponseBody("ccd-start-event-for-case-worker.json")));
+        ccdSubmitEvent(okJson(getResponseBody("ccd-submit-event-for-case-worker.json")));
+
+        Map<String, Object> caseData = new HashMap<>();
+        caseData.put("poBox","PO 12345");
+        caseData.put("journeyClassification","SUPPLEMENTARY_EVIDENCE_WITH_OCR");
+        caseData.put("formType","B123");
+        caseData.put("attachToCaseReference","1539007368674134");
```
rkondratowicz: missing spaces after `,`

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CcdCallbackController.java
```diff
@@ -41,16 +40,22 @@ public CcdCallbackController(
...
-                .getOrElseGet(errors -> errorResponse(errors));
+                .process(
+                    callback.getCaseDetails(),
+                    idamToken,
+                    userId,
+                    callback.getEventId(),
+                    callback.isIgnoreWarnings()
+                )
+                .map(this::okResponse)
+                .getOrElseGet(this::errorResponse);
```
lgonczar: I think in this case lambdas provide very useful context.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
rkondratowicz: looks like this will throw exception with message saying `"Update URL is not configured"` when service name does not have case type id

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -44,22 +52,36 @@
...
+
     private final PaymentsProcessor paymentsProcessor;
 
-    public AttachCaseCallbackService(CcdApi ccdApi, PaymentsProcessor paymentsProcessor) {
+    public AttachCaseCallbackService(ServiceConfigProvider serviceConfigProvider,
+                                     CcdApi ccdApi,
+                                     ExceptionRecordValidator exceptionRecordValidator,
+                                     CcdCaseUpdater ccdCaseUpdater,
+                                     PaymentsProcessor paymentsProcessor
+    ) {
```
lgonczar: We should, at some point (not this PR), think about how to reduce the number of dependencies in this class.

We have two high-level clients of CCD - `CcdApi` and `CcdCaseUpdater`. This seems unnecessary and the distinction between them is unclear, as both can update a CCD case.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -69,10 +91,11 @@ public AttachCaseCallbackService(CcdApi ccdApi, PaymentsProcessor paymentsProces
      *         or the list of errors, in case of errors
      */
     public Either<List<String>, Map<String, Object>> process(
-        CaseDetails exceptionRecord,
+        CaseDetails exceptionRecordData,
```
lgonczar: If we're making a change from `exceptionRecord`, how about `exceptionRecordDetails`? `CaseDetails` contains `data`, so this might be slightly confusing.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -257,48 +292,72 @@ private void attachCaseByCcdId(
...
+                    "Successfully attach Exception Record with ID {} to case with id {}",
+                    theCase.getId(),
+                    exceptionRecordData.getId()
+                );
 
+                paymentsProcessor.updatePayments(exceptionRecordData, theCase.getId());
+            }
+        } else {
+            throw new CallbackException(classificationValidation.getError());
+        }
```
doncem: How does the following look like for us?

```java
Classification classification = hasJourneyClassificationForAttachToCase(exceptionRecordData)
    .getOrElseThrow(() -> new CallbackException(classificationValidation.getError()));

switch (classification) {
    case EXCEPTION:
    case SUPPLEMENTARY_EVIDENCE:
        updateSupplementaryEvidence();
        break;
    case SUPPLEMENTARY_EVIDENCE_WITH_OCR:
        updateSupplementaryEvidenceWithOcr();
        break;
    case NEW_APPLICATION:
    default:
        throw new exception about invalid ocr being used here
}
```

- avoids unnecessary if/else combination together with validation tool being used here
- avoids nested if/elses
- splits functionality into methods. which should be done disregarding the suggestion

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
doncem: we indeed can simplify this method. first of all we need to amend the signature of already existing one so to re-use `service` retrieved prior all these steps:

```java
attachCaseByCcdId(
    event.exceptionRecordJurisdiction,
    targetCaseCcdId,
    event.exceptionRecordDocuments,
    event.exceptionRecordId,
    event.idamToken,
    event.userId,
    exceptionRecordData,
    ignoreWarnings
);
```

`event` already has `service` retrieved from case type id. perhaps we could just pass entire `event` to that method? then `getServiceConfig` gets really simple:

```java
ServiceConfigItem configItem = serviceConfigProvider.getConfig(service);

if (Strings.isNullOrEmpty(item.getUpdateUrl())) {
    throw new CallbackException();
}

return configItem;
```

NOTE. We can replace currently wrong class imported of `Strings` and just use guava's one

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -283,6 +288,40 @@ static boolean hasOcr(CaseDetails theCase) {
...
+        if (SUPPLEMENTARY_EVIDENCE_WITH_OCR.equals(classification) && !hasOcr(theCase)) {
+            return invalid(format(
+                "The current journey classification %s is not allowed without OCR",
+                classification
+            ));
+        }
+
+        return valid(classification);
+    }
+
```
doncem: Do we really need to add **any** changes to this file? it is enough to just use `hasJourneyClassification` and everything else is/will be/should be done in service (check previous suggestion). Validation on classification has been already done at the top level before executing any activity on callback. First mandatory fail fast check ;)

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
apereverzin: Will rename

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
apereverzin: Looks good

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
lgonczar: Ordering aside, can't be `==`. It must be `equals` or `Objects.equals`, but `==` is unsafe.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
apereverzin: Will do

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
lgonczar: My bad, I falsely assumed it's a constant with a String value

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -283,6 +288,40 @@ static boolean hasOcr(CaseDetails theCase) {
...
+        Optional<String> classificationOption = getJourneyClassification(theCase);
+
+        return classificationOption
+            .map(classification -> Try.of(() -> Classification.valueOf(classification)))
+            .map(Try::toValidation)
+            .map(validation -> validation
+                .mapError(throwable -> "Invalid journeyClassification. Error: " + throwable.getMessage())
+                .flatMap(classification -> validateClassificationForAttachToCase(classification, theCase))
+            )
+            .orElse(invalid("Missing journeyClassification"));
```
lgonczar: This will be an error to the caseworker, so it would be best to replace `journeyClassification` with the name that is displayed to them.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -283,6 +288,40 @@ static boolean hasOcr(CaseDetails theCase) {
...
     }
 
+    public static Validation<String, Classification> hasJourneyClassificationForAttachToCase(CaseDetails theCase) {
+        Optional<String> classificationOption = getJourneyClassification(theCase);
+
+        return classificationOption
+            .map(classification -> Try.of(() -> Classification.valueOf(classification)))
+            .map(Try::toValidation)
+            .map(validation -> validation
+                .mapError(throwable -> "Invalid journeyClassification. Error: " + throwable.getMessage())
```
lgonczar: If I understand correctly, this message will be returned to the caseworker. If so, it needs to be reduced to something like this: 
```
Invalid Journey Classification: {actual value}
```

No exception record message should be included

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
apereverzin: Will do

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/AttachExceptionRecordWithOcrTest.java
apereverzin: Oh, my bad

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -257,48 +292,72 @@ private void attachCaseByCcdId(
...
-            targetCaseDocuments,
-            ids -> throwDuplicateError(targetCaseCcdRef, ids)
-        );
-
-        List<Map<String, Object>> newCaseDocuments = attachExceptionRecordReference(
-            exceptionRecordDocuments,
-            exceptionRecordId
-        );
+        Validation<String, Classification> classificationValidation =
+            hasJourneyClassificationForAttachToCase(exceptionRecordData);
```
lgonczar: We've already got a validation stage in this service. We need to include classification validation there and hot here, so that it's all in one place and the caseworker receives all errors at once (which is the expected behaviour). It's also not ideal to have validation spread across the processing code.

Same with exception record validation.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -257,48 +292,72 @@ private void attachCaseByCcdId(
...
+                    ccdCaseUpdater.updateCase(
+                        exceptionRecordValidation.get(),
+                        serviceConfigItem,
+                        ignoreWarnings,
+                        idamToken,
+                        userId,
+                        targetCaseCcdRef
+                    );
+                } else {
+                    throw new CallbackException(exceptionRecordValidation.getError().mkString());
```
lgonczar: If exception record is invalid, the caseworker should see the related errors. This exception will result in a 500 response returned to ccd an error being logged in Bulk Scanning and the caseworker seeing a generic error message.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
lgonczar: This validation has already been done, so we shouldn't repeat it. Instead, just retrieve the service config, without any exception handling.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -283,6 +288,40 @@ static boolean hasOcr(CaseDetails theCase) {
...
+                classification
+            ));
+        }
+
+        if (SUPPLEMENTARY_EVIDENCE_WITH_OCR.equals(classification) && !hasOcr(theCase)) {
+            return invalid(format(
+                "The current journey classification %s is not allowed without OCR",
+                classification
+            ));
+        }
```
lgonczar: I'd move OCR presence validation outside classification, because it's misleading as a part of classification validation.

Also, worth adding `data` to the end of the message, because while we internally think about `OCR` as `OCR data`, it may not be obvious to the caseworker. OCR stands for Optical Character Recognition.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CallbackValidations.java
```diff
@@ -191,14 +159,16 @@ private CallbackValidations() {
...
                                     "The 'attach to case' event is not supported for supplementary evidence with OCR "
                                         + "but not containing OCR data"
+                                )
                             );
                         case CLASSIFICATION_EXCEPTION:
                             return !hasOcr(theCase)
                                 ? Validation.<String, Void>valid(null)
                                 : Validation.<String, Void>invalid(
-                                    "The 'attach to case' event is not supported for exception records with OCR"
+                                format("The 'attach to case' event is not supported for exception records with OCR")
```
lgonczar: Why wrapping this in `format`?

apereverzin: Will fix

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/caseupdate/model/response/CaseUpdateDetails.java
```diff
@@ -4,7 +4,10 @@
 
 public class CaseUpdateDetails {
 
+    @JsonProperty("event_id")
     public final String eventId;
+
+    @JsonProperty("case_data")
     public final Object caseData;
```
rkondratowicz: this is response model, those annotation are not used

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/caseupdate/CaseUpdateClient.java
```diff
@@ -54,6 +55,17 @@ public SuccessfulUpdateResponse updateCase(
...
+            log.error(
+                "Failed to update Case for case type {} and id {}, response body {}, url {}",
+                existingCase.getCaseTypeId(),
+                existingCase.getId(),
+                ex.getResponseBodyAsString(),
+                url,
+                ex
+            );
+
+            throw ex;
```
lgonczar: I guess this code was the result of some troubleshooting. Should we treat 500 response as a special case like this?

lgonczar: Same question about this code - do we want it to stay here?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -419,4 +503,14 @@ public AttachToCaseEventData(
         merged.putAll(modifiedFields);
         return merged;
     }
+
+    private ServiceConfigItem getServiceConfig(CaseDetails caseDetails) {
```
lgonczar: Why do we keep this validation if case type id has already been validated? It's better to include the result of that validation in `AttachToCaseEventData` and then just retrieve service config item based on it.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -320,7 +404,7 @@ private void verifyExceptionRecordIsNotAttachedToCase(
 
         Object attachToCaseRef = fetchedExceptionRecord.getData().get(ATTACH_TO_CASE_REFERENCE);
 
-        if (attachToCaseRef != null && Strings.isNotEmpty(attachToCaseRef.toString())) {
+        if (attachToCaseRef != null && !Strings.isNullOrEmpty(attachToCaseRef.toString())) {
```
lgonczar: I don't think this change will help - if `attachToCaseRef` is null, an exception will still be thrown.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -162,143 +186,203 @@ public AttachCaseCallbackService(CcdApi ccdApi, PaymentsProcessor paymentsProces
...
+            ccdCaseUpdater.updateCase(
+                exceptionRecordValidation.get(),
+                serviceConfigItem,
+                ignoreWarnings,
+                callBackEvent.idamToken,
+                callBackEvent.userId,
+                targetCaseCcdRef
+            );
+        } else {
+            throw new CallbackException(exceptionRecordValidation.getError().mkString());
```
lgonczar: This is a validation error - it needs to be returned by the caseworker within a 200 response, inside `errors` array. This exception will, in turn, result in a 500 response, which the caseworker will see as a generic error. To make things slightly worse, we'll log all these errors, as they make the exception message.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -162,143 +186,203 @@ public AttachCaseCallbackService(CcdApi ccdApi, PaymentsProcessor paymentsProces
...
+        AttachToCaseEventData callBackEvent,
+        String targetCaseCcdRef,
+        CaseDetails exceptionRecordDetails,
+        boolean ignoreWarnings
+    ) {
+        ServiceConfigItem serviceConfigItem = getServiceConfig(exceptionRecordDetails);
+        Validation<Seq<String>, ExceptionRecord> exceptionRecordValidation =
+            exceptionRecordValidator.getValidation(exceptionRecordDetails);
+        if (exceptionRecordValidation.isValid()) {
+            ccdCaseUpdater.updateCase(
```
lgonczar: Warnings from the update endpoint (within a 200 response) must be returned to the citizen (also within a 200 endpoint). In this case, no action (like actual case update in CCD) should be taken.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -162,143 +186,203 @@ public AttachCaseCallbackService(CcdApi ccdApi, PaymentsProcessor paymentsProces
...
 
+    private void updateSupplementaryEvidenceWithOcr(
+        AttachToCaseEventData callBackEvent,
+        String targetCaseCcdRef,
+        CaseDetails exceptionRecordDetails,
+        boolean ignoreWarnings
+    ) {
+        ServiceConfigItem serviceConfigItem = getServiceConfig(exceptionRecordDetails);
+        Validation<Seq<String>, ExceptionRecord> exceptionRecordValidation =
+            exceptionRecordValidator.getValidation(exceptionRecordDetails);
```
lgonczar: This validation should happen together with the overall validation of the request. We can include this optional exception record in AttachToCaseEventData

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -192,11 +218,19 @@ private void updateCaseInCcd(
...
+                caseDataContentStr = "Exception";
+            }
             log.error(
-                "Failed to update case for {} jurisdiction with case Id {} based on exception record with Id {}. {}",
+                "Failed to update case for {} jurisdiction with case Id {} based on exception record with Id {}. "
+                    + "caseDataContent: {}, {}",
                 exceptionRecord.poBoxJurisdiction,
                 existingCase.getId(),
                 exceptionRecord.id,
+                caseDataContentStr,
```
aliveni: can we log the whole case data?

lgonczar: This is actually not case data content, as this is an error response, but I'm not sure we should log it, anyway. Otherwise, we might end up having some sensitive information accidentally logged, as a result of service team's mistake. Thoughts?

lgonczar: I guess this served troubleshooting purpose.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -192,11 +218,19 @@ private void updateCaseInCcd(
             );
         } catch (FeignException exception) {
```
lgonczar: We should also explicitly catch `FeignException.UnprocessableEntity` - in this case we should return a 200 response with errors and warnings from the called service. The caseworker needs to see those.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -148,6 +149,26 @@ public ProcessResult updateCase(
...
+
+            throw new CallbackException(
+                format(
+                    "Failed to update case for %s service with case Id %s based on exception record %s",
+                    configItem.getService(),
+                    existingCaseId,
+                    exceptionRecord.id
+                ),
+                exception
+            );
```
lgonczar: Why not just throw exception with the proper message? It will get logged.

lgonczar: Maybe these are some remains of troubleshooting?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -162,143 +180,219 @@ public AttachCaseCallbackService(CcdApi ccdApi, PaymentsProcessor paymentsProces
...
+
+            targetCaseCcdId = callBackEvent.targetCaseRef;
         }
 
-        return ImmutableMap.of(ATTACH_TO_CASE_REFERENCE, targetCaseCcdId);
+        log.info(
+            "Successfully attached exception record {} to case {}",
+            callBackEvent.exceptionRecordId,
+            callBackEvent.targetCaseRef
+        );
```
lgonczar: Are we sure this happened?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/AttachExceptionRecordWithOcrTest.java
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    private void setUpCcdSubmitEvent(ResponseDefinitionBuilder response) {
+        givenThat(
+            post(
+                // values from config + initial request body
+                "/caseworkers/" + USER_ID
+                    + "/jurisdictions/BULKSCAN"
+                    + "/case-types/BULKSCAN_ExceptionRecord"
+                    + "/cases/1539007368674134"
```
rkondratowicz: please make `1539007368674134` a parameter

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -190,6 +224,19 @@ private void updateCaseInCcd(
...
+            String msg = format("Service returned 422 Unprocessable Entity response when trying to update "
+                + "case for %s jurisdiction with case Id %s based on exception record with Id %s. "
+                + "Service response: %s",
+                exceptionRecord.poBoxJurisdiction,
+                existingCase.getId(),
+                exceptionRecord.id,
+                exception.contentUTF8());
+            log.error(msg, exception);
+
+            return Either.left(singletonList(msg));
```
lgonczar: Sorry, this bit is my fault - I thought I was making a comment on catching exceptions resulting from update API's response, not CCD's response. This bit should be removed.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -113,18 +139,61 @@ public AttachCaseCallbackService(CcdApi ccdApi, PaymentsProcessor paymentsProces
...
+                idValidation.get(),
+                scannedRecordValidation.get(),
+                idamTokenValidation.get(),
+                userIdValidation.get(),
+                classificationValidation.get(),
+                exceptionRecordValidation.get()
+            ));
+        }
+
+        if (exceptionRecordValidation.isInvalid()) {
```
lgonczar: Worth starting this line with `else`

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -68,35 +92,37 @@ public AttachCaseCallbackService(CcdApi ccdApi, PaymentsProcessor paymentsProces
...
             .map(attachCaseResult ->
-                attachCaseResult.map(modifiedFields -> mergeCaseFields(exceptionRecord.getData(), modifiedFields))
+                attachCaseResult.map(modifiedFields ->
+                    mergeCaseFields(exceptionRecordDetails.getData(), modifiedFields))
             )
-            .getOrElseGet(errors -> Either.left(errors.toJavaList()));
+            .getOrElseGet(errors -> Either.left(ErrorsAndWarnings.withErrors(errors.toJavaList())));
     }
 
     private Validation<Seq<String>, AttachToCaseEventData> getValidation(
```
lgonczar: This bit has grown so big that it would make sense to extract it to a separate class. Doesn't have to be done in this PR.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -162,143 +226,211 @@ public AttachCaseCallbackService(CcdApi ccdApi, PaymentsProcessor paymentsProces
...
-                event.exceptionRecordJurisdiction,
-                event.targetCaseRef,
-                event.exceptionRecordDocuments,
-                event.exceptionRecordId,
-                event.idamToken,
-                event.userId,
-                exceptionRecord
+            Either<ErrorsAndWarnings, Boolean> attachResult = attachCaseByCcdId(
+                callBackEvent,
+                callBackEvent.targetCaseRef,
```
lgonczar: If we're passing the whole event (rightly), we can get rid of this parameter.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -162,143 +226,211 @@ public AttachCaseCallbackService(CcdApi ccdApi, PaymentsProcessor paymentsProces
...
+            exceptionRecordDetails.getId()
         );
 
-        paymentsProcessor.updatePayments(exceptionRecord, theCase.getId());
+        paymentsProcessor.updatePayments(exceptionRecordDetails, theCase.getId());
 
+        return Either.right(true);
+    }
+
+    private Either<ErrorsAndWarnings, Boolean> updateSupplementaryEvidenceWithOcr(
```
lgonczar: If Right, the result will always be `true`, so maybe `Optional` would be better?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -162,143 +226,211 @@ public AttachCaseCallbackService(CcdApi ccdApi, PaymentsProcessor paymentsProces
...
+        ProcessResult processResult = ccdCaseUpdater.updateCase(
+            callBackEvent.exceptionRecord,
+            serviceConfigItem,
+            ignoreWarnings,
+            callBackEvent.idamToken,
+            callBackEvent.userId,
+            targetCaseCcdRef
+        );
+
+        if (!processResult.getWarnings().isEmpty()) {
```
lgonczar: What if `processResult` has errors?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -113,18 +140,59 @@ public AttachCaseCallbackService(CcdApi ccdApi, PaymentsProcessor paymentsProces
...
+        Validation<String, String> idamTokenValidation = hasIdamToken(requesterIdamToken);
+        Validation<String, String> userIdValidation = hasUserId(requesterUserId);
+        Validation<String, Classification> classificationValidation =
+            hasJourneyClassificationForAttachToCase(exceptionRecord);
+
+        final Validation<Seq<String>, ExceptionRecord> exceptionRecordValidation;
+        if (classificationValidation.isValid() && classificationValidation.get() == SUPPLEMENTARY_EVIDENCE_WITH_OCR) {
+            exceptionRecordValidation = exceptionRecordValidator.getValidation(exceptionRecord);
+        } else {
+            exceptionRecordValidation = Validation.valid(null);
```
rkondratowicz: please explain

PR #806, opened 2019-12-13
DO NOT MERGE - Simplify updating case a bit

rkondratowicz:
Proposing simplification to an open PR

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
@@ -4,7 +4,6 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 import io.vavr.collection.Seq;
 import io.vavr.control.Either;
-import io.vavr.control.Try;
 import io.vavr.control.Validation;
 import org.apache.logging.log4j.util.Strings;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -301,16 +300,24 @@ private void attachCaseByCcdId(
...
-                ccdCaseUpdater.updateCase(
-                    exceptionRecord,
-                    serviceConfigItem,
-                    ignoreWarnings,
-                    idamToken,
-                    userId,
-                    targetCaseCcdRef
-                );
+                Validation<String, String> serviceNameValidationResult =
+                    hasServiceNameInCaseTypeId(exceptionRecordData);
```
lgonczar: There's no point validating service name after it's already been validated at an earlier stage of request processing.

lgonczar: But still, this is an improvement to the original version

apereverzin: Leszek, are you not off today?

lgonczar: This is clearer, done using less code and is some step towards removing validation (and handling its failure) from the place in the code that shouldn't be validating things which were already validated.

lgonczar: This is clearer, done using less code and is some step towards removing validation (and handling its failure) from the place in the code that shouldn't be validating things which were already validated.

doncem: Removing or not - extra if/else is not necessary. will certainly make even more clearance if used `.getOrElseThrow` right here. Just a third party suggestion here, not a blocker comment

lgonczar: There will be nothing to invoke `getOrElseThrow` on when we remove this duplicated validation

PR #807, opened 2019-12-16
Return 400 when case in callback is empty

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-934


lgonczar:
If we go with this solution, we need to log this as an error in order to be alerted.

@nitinprabhu has a slightly different view on the response that we should return

nitinprabhu:
> If we go with this solution, we need to log this as an error in order to be alerted.
> 
> @nitinprabhu has a slightly different view on the response that we should return

If we return 400 then it absolutely makes sense from REST perspective but CCD won't handle it in a nice way and would show generic message stating callback failed.

If we could display a valid message to the caseworker it would be helpful for them so that they know what action needs to be taken(Please contact service desk...).

We should log an error so that we are alerted and let CCD know that contract is broken.

PR #811, opened 2019-12-18
Supplementary evidence cleanup

rkondratowicz:
minor refactoring to (hopefully) make it more clear what's happening

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/AttachDocsToSupplementaryEvidence.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/AttachDocsToSupplementaryEvidence.java
@@ -42,6 +42,7 @@ class AttachDocsToSupplementaryEvidence {
     public boolean attach(Envelope envelope, CaseDetails existingCase) {
         if (mapper.getDocsToAdd(getDocuments(existingCase), envelope.documents).isEmpty()) {
             log.warn("Envelope {} has no new documents. CCD Case {} not updated", envelope.id, existingCase.getId());
+            return true;
         } else {
             log.info(
                 "Attaching supplementary evidence. Envelope: {}. Case: {}. Case type: {}",
@@ -83,6 +84,7 @@ class AttachDocsToSupplementaryEvidence {
                     existingCase.getId(),
                     envelope.id
                 );
+                return true;
             } catch (UnableToAttachDocumentsException e) {
                 log.error(
                     "Failed to attach documents from envelope to case. Case Type: {}, Case ID: {}, envelope ID: {}",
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EnvelopeHandler.java
```diff
@@ -44,13 +44,15 @@ public void handleEnvelope(Envelope envelope) {
                         paymentsProcessor.createPayments(envelope, existingCase.getId(), false);
                     } else {
                         log.info(
-                            "Creating exception record as supplementary evidence failed for envelope {} case {}",
+                            "Adding supplementary evidence from envelope {} to case {} failed. "
+                                + "Creating exception record instead.",
```
rkondratowicz: took me a while to realise 'Creating exception record as supplementary evidence failed' meant 'because', not 'creating X as something Y' failed.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/AttachDocsToSupplementaryEvidence.java
```diff
@@ -93,7 +95,6 @@ public boolean attach(Envelope envelope, CaseDetails existingCase) {
                 return false;
             }
         }
-        return true;
```
lgonczar: 👍 No novels

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EnvelopeHandler.java
lgonczar: Message change seems right. Maybe we should log it as an exception? This is an error situation, that we should be alerted about, investigate and most likely product team will have to fix their definition as a result.

nitinprabhu: Yes that's correct we need to log it as an exception otherwise it will too late for us to figure out something has gone wrong. On another note we tend to log everything as error just so that we can get error email. Ideally we should create different alert config in app insights and not just make it log as an error.

lgonczar: Worth exploring that possibility

aliveni: it's being logged as an error in https://github.com/hmcts/bulk-scan-orchestrator/blob/39dccc83d766d859814da2c2b5f65061a3d2bcd6/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/AttachDocsToSupplementaryEvidence.java#L89

PR #813, opened 2019-12-19
Fix exception record finalisation

lgonczar:

### Change description ###

Fix exception record finalisation (updating exception record data after successful callback processing). A different case reference field should be set for case creation and case update.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCallbackType.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCallbackType.java
@@ -1,0 +1,6 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd;
+
+public enum CcdCallbackType {
+    CASE_CREATION,
+    ATTACHING_SUPPLEMENTARY_EVIDENCE
+}
--- a/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/ExceptionRecordFinalizerTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/ExceptionRecordFinalizerTest.java
@@ -1,5 +1,6 @@
 package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd;
 
+import com.google.common.collect.ImmutableMap;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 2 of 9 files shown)


src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/ExceptionRecordFinalizerTest.java
```diff
@@ -9,16 +10,18 @@
...
 
 class ExceptionRecordFinalizerTest {
 
     private static final long CASE_ID = 100L;
     private static final String FIELD_1 = "field1";
 
+    public static final String CASE_REFERENCE = "caseReference";
+    public static final String DISPLAY_WARNINGS = "displayWarnings";
+    public static final String OCR_DATA_VALIDATION_WARNINGS = "ocrDataValidationWarnings";
+    public static final String ATTACH_TO_CASE_REFERENCE = "attachToCaseReference";
```
lgonczar: We can't reuse constants from the main code when verifying map entries. Otherwise, we'd be checking if "the code sets the entries that it sets".

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCallbackType.java
```diff
@@ -0,0 +1,6 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd;
+
+public enum CcdCallbackType {
```
nitinprabhu: may be name it as `CcdEventType`

lgonczar: I don't know, event type might be misleading. This enum represents the kind of action that's happening, rather than a specific CCD event.

nitinprabhu: There is nothing called as callback type :) what you are referring here are purely ccd events attach supplementary evidence/ case creation.

lgonczar: This enum is here to provide information about the kind of action we're serving, not which event was triggered. If you're not happy with `CcdCallbackType`, we can come up with something else, but I don't want to create the impression that these are event IDs.

PR #814, opened 2019-12-20
Change surname conf

mdayican:

https://tools.hmcts.net/jira/browse/BPS-965

### Change description ###

changing fomtype ocr surname mapping to support multiple field names

```
[ ] Yes
[x] No
```


mdayican:
pipeline fails

PR #834, opened 2020-01-15
Idempotent supplementary evidence with OCR updates

mdayican:

https://tools.hmcts.net/jira/browse/BPS-951

### Change description ###

comparing case document control_number numbers with exception

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
@@ -453,7 +453,7 @@ public class AttachCaseCallbackService {
                 } else {
                     log.warn(
                         "Attempt to attach an already attached exception record to same case, "
-                            + "Document control number(s)  are not matching, will update the case "
+                            + "Document control number(s) are not matching, will update the case "
                             + "Exception record ID: {}, attach to case: {}",
                         callBackEvent.exceptionRecordId,
                         targetCaseCcdRef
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -420,30 +421,55 @@ private void updateSupplementaryEvidence(
         CaseDetails exceptionRecordDetails,
         boolean ignoreWarnings
     ) {
-        Optional<String> attachedToCase = getCaseExceptionRecordIsAttachedTo(
-            callBackEvent.exceptionRecordJurisdiction,
-            callBackEvent.exceptionRecordId
+        CaseDetails fetchedExceptionRecord = ccdApi.getCase(
+            callBackEvent.exceptionRecordId.toString(),
+            callBackEvent.exceptionRecordJurisdiction
```
nitinprabhu: not sure this is correct. You are fetching exception record but that is already passed in the callback ?

mdayican: trying to get existing case

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -427,13 +428,36 @@ private void updateSupplementaryEvidence(
 
         if (attachedToCase.isPresent()) {
```
mdayican: of this is present it means it is updated and status already changed.
if it is present we should trow an exception.

mdayican:
will create new one.

PR #831, opened 2020-01-07
Idempotent update

apereverzin:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/BPS-951

### Change description ###
Handling already attached exception records - making no calls in this case.
Integration tests updated.

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #836, opened 2020-01-16
Idempotent supplementary evidence with ocr updates

mdayican:

https://tools.hmcts.net/jira/browse/BPS-951

### Change description ###

make Idempotent supplementary evidence with ocr updates.

revert some code back which was done in this PR:
https://github.com/hmcts/bulk-scan-orchestrator/pull/831

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
@@ -22,11 +22,14 @@ import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
 import uk.gov.hmcts.reform.ccd.client.model.Event;
 import uk.gov.hmcts.reform.ccd.client.model.StartEventResponse;
 
+import java.util.List;
 import java.util.Optional;
 
 import static java.lang.String.format;
 import static java.util.Collections.emptyList;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -195,6 +208,22 @@ public ProcessResult updateCase(
         }
     }
 
+    private boolean isExceptionAlreadyAttached(CaseDetails existingCase,
+                                               ExceptionRecord exceptionRecord) {
```
rkondratowicz: formatting

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -93,6 +96,16 @@ public ProcessResult updateCase(
 
             final CaseDetails existingCase = startEvent.getCaseDetails();
 
+            if (isExceptionAlreadyAttached(existingCase, exceptionRecord)) {
+                log.warn(
+                    "Exception already attached to case. "
```
lgonczar: Maybe it would be worth mentioning that we're skipping update because all documents from this exception record are already in the case. That's to distinguish between this scenario and one where we check by `attachedToCaseReference` field.

PR #838, opened 2020-01-17
Pass case id in case update

mdayican:

https://tools.hmcts.net/jira/browse/BPS-968

### Change description ###

Pass case id in when updating case

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/caseupdate/model/request/ExistingCaseDetails.java
```diff
@@ -5,16 +5,22 @@
 import java.util.Map;
 
 public class ExistingCaseDetails {
+
+    @JsonProperty("id")
+    public final String id;
```
rkondratowicz: will sample-app accept this extra field? Perhaps we should add it on the other side first?

mdayican: sample app changes:
https://github.com/hmcts/bulk-scan-ccd-event-handler-sample-app/pull/113

PR #842, opened 2020-01-22
Handling FeignException.Conflict

apereverzin:

### Change description ###
Handling FeignException.Conflict thrown by CCD instead of Conflict exception.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
@@ -279,11 +279,27 @@ public class CcdCaseUpdater {
             return Optional.empty();
         } catch (FeignException.UnprocessableEntity exception) {
             String msg = format(
-                "Service returned 422 Unprocessable Entity response "
+                "CCD returned 422 Unprocessable Entity response "
                     + "when trying to update case for %s jurisdiction "
                     + "with case Id %s "
                     + "based on exception record with Id %s. "
-                    + "Service response: %s",
+                    + "CCD response: %s",
+                exceptionRecord.poBoxJurisdiction,
+                existingCase.getId(),
+                exceptionRecord.id,
+                exception.contentUTF8()
+            );
+            log.error(msg, exception);
+
+            return Optional.of(msg);
+        } catch (FeignException.Conflict exception) {
+            String msg = format(
+                "CCD returned 409 Conflict response "
+                    + "when trying to update case for %s jurisdiction "
+                    + "with case Id %s "
+                    + "based on exception record with Id %s "
+                    + "because it has been updated in the meantime. "
+                    + "CCD response: %s",
                 exceptionRecord.poBoxJurisdiction,
                 existingCase.getId(),
                 exceptionRecord.id,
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -279,11 +279,27 @@ private String getErrorMessage(String service, String existingCaseId, String exc
...
+            log.error(msg, exception);
+
+            return Optional.of(msg);
+        } catch (FeignException.Conflict exception) {
+            String msg = format(
+                "CCD returned 409 Conflict response "
+                    + "when trying to update case for %s jurisdiction "
+                    + "with case Id %s "
+                    + "based on exception record with Id %s "
+                    + "because it has been updated in the meantime. "
```
rkondratowicz: this is just our assumption. Perhaps this can be removed and building this message could be them moved to a shared function.

lgonczar: This message will go to the caseworker, so the message returned by this method should be simple, e.g.
```
This action couldn't be completed, because this exception record was being updated by someone else.
```
Ideally, message should be agreed with Ade.

lgonczar: It's ok to log an error like this, though.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -279,11 +279,27 @@ private String getErrorMessage(String service, String existingCaseId, String exc
...
-                    + "Service response: %s",
+                    + "CCD response: %s",
+                exceptionRecord.poBoxJurisdiction,
+                existingCase.getId(),
+                exceptionRecord.id,
+                exception.contentUTF8()
+            );
+            log.error(msg, exception);
+
+            return Optional.of(msg);
```
lgonczar: We can't return the error to the caseworker in this case. Instead, the processing should fail, resulting in the orchestrator returning a 500 response to CCD.

lgonczar:
We don't seem to handle 422 responses from the Update API. We need some cleaning before the 409 change.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -159,7 +158,7 @@ public ProcessResult updateCase(
         } catch (UnprocessableEntity exception) {
             ClientServiceErrorResponse errorResponse = serviceResponseParser.parseResponseBody(exception);
             return new ProcessResult(errorResponse.warnings, errorResponse.errors);
-        } catch (Conflict exception) {
+        } catch (FeignException.Conflict exception) {
```
lgonczar: I've looked in the code and it turns out we don't need this `catch` here at all, because 409 is handled in `updateCaseInCcd`. If here we get any other 409, that should be treated as any error.

PR #843, opened 2020-01-23
Add config for divorce to create a new case before payments are processed

aliveni:

```
[ ] Yes
[x] No
```


src/main/resources/application.yaml
```diff
@@ -102,6 +102,7 @@ service-config:
         - DIVORCE
         - DIVORCE_NOTICE_OF_ACTING
       transformation-url: ${TRANSFORMATION_URL_DIVORCE}
+      allow-creating-case-before-payments-are-processed: true
```
nitinprabhu: should this be enabled per environment ?

aliveni: I don't think so. For probate, we just have it in applciation.yaml file

PR #847, opened 2020-01-27
Fix wiremock usage in integration tests

doncem:
### Change description ###

For a very long time we could not get out of boundary of spring wiremock contract dependency further then 2.0.*. Hopefully this one sorts things out.

Thing to consider: perhaps just leaving `AutoConfigureWireMock` in `IntegrationTest` would be sufficient for a fix/change introduced here. Would reduce lines of change. Started of by separating this feature and ended up the way it is at the moment of description draft

```
[ ] Yes
[x] No
```


src/integrationTest/resources/application.properties
```diff
@@ -0,0 +1,30 @@
+scheduling.task.delete-envelopes-dlq-messages.enabled=false
+scheduling.task.consume-envelopes-queue.enabled=false
+
+hystrix.command.default.metrics.rollingStats.timeInMilliseconds=500
+hystrix.command.default.metrics.rollingStats.numBuckets=5
+hystrix.command.check-jurisdiction-log-in.circuitBreaker.sleepWindowInMilliseconds=100000
```
nitinprabhu: do we need this for int tests ?

doncem: left the differences compared to main app config. didn't pay too much attention to other configs while doing this pr

config/owasp/suppressions.xml
```diff
@@ -56,79 +56,23 @@
...
-    <cve>CVE-2019-11358</cve>
-    <!--jackson databind-->
-    <cve>CVE-2019-14379</cve>
-    <cve>CVE-2019-14439</cve>
-    <cve>CVE-2019-14540</cve>
-    <cve>CVE-2019-16335</cve>
-    <cve>CVE-2019-16942</cve>
-    <cve>CVE-2019-16943</cve>
-    <cve>CVE-2019-17531</cve>
-  </suppress>
```
rkondratowicz: 🎖

PR #864, opened 2020-02-18
Improve logging for attaching supplementary evidence

rkondratowicz:
Some logs included zip file name and some did not (just envelope ID), making it harder to check the history of given file.

Updated the logs so that we can see all the steps just by knowing the file name.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/AttachDocsToSupplementaryEvidence.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/AttachDocsToSupplementaryEvidence.java
@@ -48,9 +48,9 @@ class AttachDocsToSupplementaryEvidence {
             existingCase.getState()
         );
         if (mapper.getDocsToAdd(getDocuments(existingCase), envelope.documents).isEmpty()) {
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/AttachDocsToSupplementaryEvidence.java
```diff
@@ -40,15 +40,17 @@ public AttachDocsToSupplementaryEvidence(
...
-            log.warn("Envelope {} has no new documents. CCD Case {} not updated", envelope.id, existingCase.getId());
+            log.warn("Envelope has no new documents. CCD Case not updated. " + loggingContext);
         } else {
-            log.info(
-                "Attaching supplementary evidence. Envelope: {}. Case: {}. Case type: {}",
-                envelope.id,
-                existingCase.getId(),
-                existingCase.getCaseTypeId()
-            );
+            log.info("Attaching supplementary evidence. " + loggingContext);
```
doncem: shall we use `log.level("Fact. {}", context);` instead?

rkondratowicz: kk

PR #866, opened 2020-02-19
Add secrets for FPLA IDAM credentails

aliveni:
https://tools.hmcts.net/jira/browse/BPS-1050

### Change description ###
Added publiclaw username and password secrets, which will be used to create Exception Record (and other CCD operations) for FPLA.

Flux PR - https://github.com/hmcts/cnp-flux-config/pull/2169

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/bulk-scan-orchestrator/values.yaml
+++ b/charts/bulk-scan-orchestrator/values.yaml
@@ -17,7 +17,6 @@ java:
     DELETE_ENVELOPES_DLQ_MESSAGES_ENABLED: "true"
     DELETE_ENVELOPES_DLQ_MESSAGES_CRON: "0 * * * * *"
     DELETE_ENVELOPES_DLQ_MESSAGES_TTL: "10s"
-    SUPPORTED_JURISDICTIONS: ["SSCS", "BULKSCAN", "DIVORCE", "PROBATE", "FINREM", "CMC", "PUBLICLAW"]
   keyVaults:
     bulk-scan:
       resourceGroup: bulk-scan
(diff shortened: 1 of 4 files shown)


charts/bulk-scan-orchestrator/values.yaml
```diff
@@ -31,6 +31,8 @@ java:
         - idam-users-div-password
         - idam-users-probate-username
         - idam-users-probate-password
+        - idam-users-publiclaw-username
+        - idam-users-publiclaw-password
```
doncem: those only exist in aat. will service be ok to launch without secrets being present in other environments?

nitinprabhu: prod pod will fail if secrets are not there so we should add them I think in all environments.
But I agree we should just enable them in AAT https://github.com/hmcts/bulk-scan-orchestrator/blob/master/infrastructure/aat.tfvars#L4

nitinprabhu: ok this is now deprecated so secrets has to exist in prod.

aliveni: I added supported jurisdictions config to be able to disable publiclaw in prod. But, we still need to add idam credentials to prod key vault

aliveni: Updated  prod config to exclude `publilaw` https://github.com/hmcts/cnp-flux-config/pull/2169/files#diff-a81b4fc624fb7995c6f17e36a59a0540R28

aliveni: The idam secrets are now added to prod vault

doncem: i presume perftest can be done together with the ticket responsible deploying orchestrator to perftest

PR #867, opened 2020-02-20
process envelopes with supported jurisdictions only

aliveni:
https://tools.hmcts.net/jira/browse/BPS-1050

### Change description ###
- Added config for supported jurisdictions. This will be moved to AKS config in https://github.com/hmcts/bulk-scan-orchestrator/pull/866 
- Process envelopes with the configured jurisdictions

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EnvelopeHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EnvelopeHandler.java
@@ -39,7 +39,7 @@ public class EnvelopeHandler {
 
     public void handleEnvelope(Envelope envelope) {
         // check if envelope jurisdiction is configured
-        isJurisdictionConfigured(envelope);
+        checkJurisdictionConfigured(envelope);
 
         switch (envelope.classification) {
             case SUPPLEMENTARY_EVIDENCE:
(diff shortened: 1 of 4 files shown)


src/main/resources/application.yaml
```diff
@@ -144,3 +144,4 @@ task:
   check-jurisdiction-log-in:
     check-validity-duration: PT5M # only ensure this often that no log-in attempt is rejected by IDAM
 
+supported-jurisdictions: SSCS,BULKSCAN,DIVORCE,PROBATE,FINREM,CMC
```
apereverzin: All these values are already present in application.yaml in service-config.services.service. Adding this new property makes sense only if we can configure a service but not to support it, otherwise it is just adding maintenance overhead.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EnvelopeHandler.java
```diff
@@ -66,6 +74,20 @@ public void handleEnvelope(Envelope envelope) {
         }
     }
 
+    private void isJurisdictionConfigured(Envelope envelope) {
```
apereverzin: 'is' prefix suggests boolean return type. 'checkIfJurisdictionIsConfigured' looks better in this case.

src/main/resources/application.yaml
aliveni: I agree.. but then, we will need a property for each service in AKS. Alternatively, we can just have 1 property for FPLA and remove it later (when it is live)

aliveni:
Closing this PR. This config will be moved to Bulk scan processor

PR #868, opened 2020-02-20
Idempotency bugfix for case creation callback

doncem:
### Change description ###

Payment should be called regardless case was found or newly created. Doing a bit of a re-work around the specific area and then moving the payment trigger there

Service Now: INC5029932

```
[x] Yes // payments will be called when case is found -> idempotent action is to proceed with payments
[x] No // it's kind of "expected behaviour" from us
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdNewCaseCreator.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdNewCaseCreator.java
@@ -1,7 +1,6 @@
 package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd;
 
 import feign.FeignException;
-import io.vavr.control.Either;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
@@ -17,7 +16,7 @@ import uk.gov.hmcts.reform.bulkscan.orchestrator.client.transformation.model.res
 import uk.gov.hmcts.reform.bulkscan.orchestrator.client.transformation.model.response.SuccessfulTransformationResponse;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.config.ServiceConfigItem;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.callback.CallbackException;
-import uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.callback.ProcessResult;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.callback.CreateCaseResult;
 import uk.gov.hmcts.reform.ccd.client.CoreCaseDataApi;
 import uk.gov.hmcts.reform.ccd.client.model.CaseDataContent;
 import uk.gov.hmcts.reform.ccd.client.model.Event;
@@ -53,7 +52,7 @@ public class CcdNewCaseCreator {
     }
 
     @SuppressWarnings("squid:S2139") // squid for exception handle + logging
-    public Either<ProcessResult, Long> createNewCase(
+    public CreateCaseResult createNewCase(
         ExceptionRecord exceptionRecord,
         ServiceConfigItem configItem,
         boolean ignoreWarnings,
@@ -81,7 +80,7 @@ public class CcdNewCaseCreator {
                     configItem.getService(),
                     exceptionRecord.id
                 );
-                return Either.left(new ProcessResult(transformationResponse.warnings, emptyList()));
+                return new CreateCaseResult(transformationResponse.warnings, emptyList());
             }
 
             log.info(
@@ -105,7 +104,7 @@ public class CcdNewCaseCreator {
                 exceptionRecord.id
             );
 
-            return Either.right(newCaseId);
+            return new CreateCaseResult(newCaseId);
         } catch (BadRequest exception) {
             throw new CallbackException(
                 format("Failed to transform exception record with Id %s", exceptionRecord.id),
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdNewCaseCreator.java
```diff
@@ -39,33 +38,27 @@
...
         this.s2sTokenGenerator = s2sTokenGenerator;
-        this.paymentsProcessor = paymentsProcessor;
         this.coreCaseDataApi = coreCaseDataApi;
-        this.exceptionRecordFinalizer = exceptionRecordFinalizer;
     }
 
-    @SuppressWarnings({"squid:S2139", "unchecked"}) // squid for exception handle + logging
-    public ProcessResult createNewCase(
+    @SuppressWarnings("squid:S2139") // squid for exception handle + logging
+    public Either<ProcessResult, Long> createNewCase(
```
lgonczar: This result type is a bit unclear. Maybe we can adjust ProcessResult to contain what we need? Or, if it's used in more places, we can have a version just for case creation.

doncem: At the moment I don't see anything unclear here only missing docblock explaining it. i'll give it a go fresh tomorrow or will await for more input. and perhaps make assumptions in case there's a significant delay on _bugfix_

lgonczar: `ProcessResult` class represents success or failure. Before the proposed changes, this method returned a result that could be returned straight to CCD. Now it returns either `ProcessResult` which only represents failure, or a bit of information that will be used for creating ProcessResult representing success. It's not very clear and needs a bit of analysis to figure out what's going on. How about replacing `ProcessResult` (just in `CcdNewCaseCreator`) with a different class, say `CaseCreationResult`, which would contain errors, warnings and new case ID, and make the calling code always prepare `ProcessResult` out of it?

I like the fact that you moved the use of `ExceptionRecordFinalizer` to the calling method - now we consistently prepare exception record data there.

doncem: appreciate such response. will proceed with changes sooner than later ;)

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -163,32 +167,57 @@ private ProcessResult tryCreateNewCase(
...
+                        return new ProcessResult(
+                            exceptionRecordFinalizer.finalizeExceptionRecord(
+                                exceptionRecordData.getData(),
+                                id,
+                                CcdCallbackType.CASE_CREATION
+                            )
+                        );
+                    });
+
+                return processedResult.isLeft() ? processedResult.getLeft() : processedResult.get();
```
lgonczar: The logic here is quite simple, but the code doesn't read very easily. There are transformations on `Either<ProcessResult, ProcessResult>`, where the left `ProcessResult` represents a different situation than the right one and at some point the right one gets transformed into a different `ProcessResult`. Worth making this code more explicit.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -163,32 +167,57 @@ private ProcessResult tryCreateNewCase(
...
+            paymentsProcessor.updatePayments(exceptionRecordData, caseId);
+        } catch (PaymentsPublishingException exception) {
+            log.error(
+                "Failed to send update to payment processor for {} exception record {}",
+                serviceName,
+                exceptionRecordData.getId(),
+                exception
+            );
+
+            throw new CallbackException("Payment references cannot be processed. Please try again later", exception);
```
lgonczar: The caseworker won't see this message - we'll just log the error and return a 500 response. This bit of code has been moved from the other class and it doesn't change the existing behaviour, so not necessarily something for this PR.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -163,32 +167,67 @@ private ProcessResult tryCreateNewCase(
...
-                );
-            } else if (ids.size() == 1) {
-                final Map<String, Object> finalizedExceptionRecordData =
-                    exceptionRecordFinalizer.finalizeExceptionRecord(
-                        exceptionRecordData.getData(),
-                        ids.get(0),
-                        CcdCallbackType.CASE_CREATION
-                    );
-                return new ProcessResult(finalizedExceptionRecordData);
-            } else {
```
nitinprabhu: not sure why is this being removed ? If a case exists then will it not create a new case ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackService.java
```diff
@@ -189,6 +188,46 @@ private ProcessResult tryCreateNewCase(
...
+            log.error(
+                "Failed to send update to payment processor for {} exception record {}",
+                serviceName,
+                exceptionRecordData.getId(),
+                exception
+            );
+
+            return new ProcessResult(
+                emptyList(),
+                singletonList("Payment references cannot be processed. Please try again later")
```
nitinprabhu: worth confirming this message with Ade as this will be displayed on UI.

doncem: it was confirmed long since. same message is used for attaching docs callback

PR #869, opened 2020-02-21
Revert 866 add publiclaw secrets

nitinprabhu:
### Change description ###

- `QueueProcessingReadinessChecker` tries to login for all jurisdictions and if login fails then it pauses the processing of queue through. Circuit breaker.

- As a result we are also getting heartbeat message failure alerts.

- There are currently 196 messages in envelopes queue in prod which are not processed due to envelopes queue being paused.

- Idam team has still not created the user in prod hence reverting as messages in the queue needs to be processed asap.

```
[ ] Yes
[ ] No
```

### DIFF
--- a/charts/bulk-scan-orchestrator/Chart.yaml
+++ b/charts/bulk-scan-orchestrator/Chart.yaml
@@ -1,7 +1,7 @@
 name: bulk-scan-orchestrator
 apiVersion: v1
 home: https://github.com/hmcts/bulk-scan-orchestrator
-version: 0.2.5
+version: 0.2.4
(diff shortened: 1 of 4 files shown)


charts/bulk-scan-orchestrator/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 name: bulk-scan-orchestrator
 apiVersion: v1
 home: https://github.com/hmcts/bulk-scan-orchestrator
-version: 0.2.5
+version: 0.2.6
```
nitinprabhu: https://github.com/hmcts/cnp-flux-config/pull/2204

PR #875, opened 2020-02-24
Fix log entry to recognise case ID

doncem:

[Report](https://tools.hmcts.net/jira/browse/BPS-960)

### Change description ###

While was adding reason into the existing query, also noticed there are quite a sum of legacy cases which are not included in the report. Although it is only ad-hoc but ideally would like to have such link in automatic way. The report linked ^ will automatically work for legacy cases too

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/envelopes/EnvelopeEventProcessor.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/envelopes/EnvelopeEventProcessor.java
@@ -185,13 +185,14 @@ public class EnvelopeEventProcessor {
 
     private void logMessageParsed(IMessage message, Envelope envelope) {
         log.info(
-            "Parsed message. ID: {}, Envelope ID: {}, File name: {}, Jurisdiction: {}, Classification: {}, Case: {}",
+            "Parsed message. ID: {}, Envelope ID: {}, File name: {}, Jurisdiction: {}, Classification: {}, {}: {}",
             message.getMessageId(),
             envelope.id,
             envelope.zipFileName,
             envelope.jurisdiction,
             envelope.classification,
-            envelope.caseRef
+            envelope.caseRef == null ? "Legacy Case" : "Case",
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/envelopes/EnvelopeEventProcessor.java
```diff
@@ -191,7 +191,7 @@ private void logMessageParsed(IMessage message, Envelope envelope) {
             envelope.zipFileName,
             envelope.jurisdiction,
             envelope.classification,
-            envelope.caseRef
+            envelope.caseRef == null ? envelope.legacyCaseRef : envelope.caseRef
```
rkondratowicz: and how do we then know if it legacy or not in logs?

nitinprabhu: is this query resolved @doncem ?

doncem: this query is successfully ignored as very low priority plus initial solution just automagically solves the App Insights report for me/us

rkondratowicz: consider:

```java
envelope.caseRef != null
    ? "Case: " + envelope.caseRef
    : "Legacy case: " + envelope.legacyCaseRef
```
instead of 
```java
envelope.caseRef == null ? "Legacy Case" : "Case",
envelope.caseRef == null ? envelope.legacyCaseRef : envelope.caseRef
```

doncem: when logging, passing parameters is recommended (and even used to be reported by sonar on failure doing that) instead of mixing up with string concatenation

PR #880, opened 2020-02-27
Increase timeout for case creation assertion in the functional test

aliveni:
### Change description ###
Create case idempotency functional test is frequently failing due to the short timeout.
Increased timeout to 30 seconds.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/CreateCaseTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/CreateCaseTest.java
@@ -95,7 +95,6 @@ class CreateCaseTest {
 
         await("Case is ingested")
             .atMost(30, TimeUnit.SECONDS)
-            .pollDelay(2, TimeUnit.SECONDS)
             .pollInterval(1, TimeUnit.SECONDS)
             .until(() -> caseIngested(bulkScanCaseReference));
 
--- a/src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/CreateCaseTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/CreateCaseTest.java
@@ -94,7 +94,7 @@ class CreateCaseTest {
         assertThat(bulkScanCaseReference.equals(String.valueOf(exceptionRecord.getId())));
 
         await("Case is ingested")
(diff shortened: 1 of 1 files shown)


src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/CreateCaseTest.java
```diff
@@ -94,8 +94,9 @@ public void should_idempotently_create_case_from_valid_exception_record() throws
         assertThat(bulkScanCaseReference.equals(String.valueOf(exceptionRecord.getId())));
 
         await("Case is ingested")
-            .atMost(2, TimeUnit.SECONDS)
-            .pollDelay(1, TimeUnit.SECONDS)
+            .atMost(30, TimeUnit.SECONDS)
+            .pollDelay(2, TimeUnit.SECONDS)
+            .pollInterval(1, TimeUnit.SECONDS)
```
nitinprabhu: do we need this if we have poll delay ?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/CreateCaseTest.java
```diff
@@ -94,8 +94,9 @@ public void should_idempotently_create_case_from_valid_exception_record() throws
         assertThat(bulkScanCaseReference.equals(String.valueOf(exceptionRecord.getId())));
 
         await("Case is ingested")
-            .atMost(2, TimeUnit.SECONDS)
-            .pollDelay(1, TimeUnit.SECONDS)
+            .atMost(30, TimeUnit.SECONDS)
```
nitinprabhu: 30 secs may be too much ? I would say let's keep it 10 secs for now ?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/CreateCaseTest.java
aliveni: I removed it. The pollInterval should be enough

aliveni:
> Do we know the reason what causes the delay?
Creating a new case and searching for the case by `bulkscanCaseReference` seems to be taking 2 seconds (It's probably missed by milliseconds). By increasing the timeout, we can avoid rebuilding the whole pipeline.
As discussed, I decreased the timeout from 30 seconds to 10 seconds

PR #885, opened 2020-02-28
Introduce ER document connectives for duplicates

doncem:

### Change description ###

In order to send payments message disregarding documents where attached or not, need to rework the duplicate check first. Introducing helper which will contain other logical set comparison to determine whether there are duplicates amond missing or duplicates entirely which means everything is attached and just need to publish payment message

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/util/ExceptionRecordAttachDocumentConnectives.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/util/ExceptionRecordAttachDocumentConnectives.java
@@ -1,0 +1,25 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd.util;
+
+import java.util.Set;
+
+public class ExceptionRecordAttachDocumentConnectives {
+
+    /**
+     * Exception Record documents already present in target case.
+     */
+    private Set<String> duplicates;
+
+    public ExceptionRecordAttachDocumentConnectives(
+        Set<String> duplicates
+    ) {
+        this.duplicates = duplicates;
+    }
+
+    public boolean hasDuplicates() {
+        return !duplicates.isEmpty();
+    }
+
+    public Set<String> getDuplicates() {
+        return duplicates;
+    }
+}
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -388,13 +388,22 @@ private void updateSupplementaryEvidence(
...
-        //This is done so exception record does not change state if there is a document error
-        checkForDuplicatesOrElse(
+        ExceptionRecordAttachDocumentConnectives erDocumentConnectives = calculateDocumentConnectives(
             callBackEvent.exceptionRecordDocuments,
-            targetCaseDocuments,
-            ids -> throwDuplicateError(targetCaseCcdRef, ids)
+            targetCaseDocuments
         );
 
+        if (erDocumentConnectives.hasDuplicates()) {
```
nitinprabhu: question : If documents were already attached should we not proceed with payments ?
Duplicate document scenario would only be possible when callback timed out and caseworker has retried ?

doncem: i'll handle this scenario in future prs ;). this is refactor so that document connectives can be expanded and recognise such scenario

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/util/ExceptionRecordAttachDocumentConnectives.java
```diff
@@ -0,0 +1,25 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.model.ccd.util;
+
+import java.util.Set;
+
+public class ExceptionRecordAttachDocumentConnectives {
```
apereverzin: Disagree with having such thin wrapper around Set, it is enough to have Set object with a relevant name - e.g. documentConnectivesDuplicates

doncem: it will contain other value in the following PR - missing documents. the logical connection between duplicates and missing will determine whether there are duplicates or everything got attached and we face the idempotency as expected.

this is doing small refactoring approach so that further implementation doesn't make reviewers cross-eyed

p.s. it is really very difficult to come up with decent name. can always be renamed in the future once bright idea comes in. should not be a blocker here tbh

doncem:
will recreate as pipeline upgrade not compatible

PR #891, opened 2020-03-09
Idempotency for attaching supplementary evidence callback

doncem:

### Change description ###

2 "Update" commits are preparing the change which is included in "Add" commit at the end

```
[x] Yes # changing the current behaviour
[x] No # but we suppose to support it anyway
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/util/ExceptionRecordAttachDocumentConnectives.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/util/ExceptionRecordAttachDocumentConnectives.java
@@ -9,14 +9,25 @@ public class ExceptionRecordAttachDocumentConnectives {
      */
     private Set<String> duplicates;
 
+    /**
+     * Exception Record documents which are still missing in target case.
+     */
+    private Set<String> missing;
+
     public ExceptionRecordAttachDocumentConnectives(
-        Set<String> duplicates
+        Set<String> duplicates,
+        Set<String> missing
     ) {
         this.duplicates = duplicates;
+        this.missing = missing;
     }
 
     public boolean hasDuplicates() {
-        return !duplicates.isEmpty();
+        return !duplicates.isEmpty() && !missing.isEmpty();
+    }
+
+    public boolean hasMissing() {
+        return duplicates.isEmpty() && !missing.isEmpty();
     }
 
     public Set<String> getDuplicates() {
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/util/ExceptionRecordAttachDocumentConnectives.java
```diff
@@ -9,14 +9,25 @@
...
+        Set<String> duplicates,
+        Set<String> missing
     ) {
         this.duplicates = duplicates;
+        this.missing = missing;
     }
 
     public boolean hasDuplicates() {
-        return !duplicates.isEmpty();
+        return !duplicates.isEmpty() && !missing.isEmpty();
```
apereverzin: Not sure my understanding of implied logic is correct but the logical expression does not look matching the name of the method

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/util/ExceptionRecordAttachDocumentConnectives.java
```diff
@@ -9,14 +9,25 @@
...
+    }
+
+    public boolean hasMissing() {
+        return duplicates.isEmpty() && !missing.isEmpty();
```
apereverzin: Similar here - the expression says that there are no duplicates and there are missing documents but the name of the method says that there are missing documents and nothing about duplicates

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -404,30 +405,32 @@ private void updateSupplementaryEvidence(
             );
         }
 
-        List<Map<String, Object>> newCaseDocuments = attachExceptionRecordReference(
-            callBackEvent.exceptionRecordDocuments,
-            callBackEvent.exceptionRecordId
-        );
+        if (erDocumentConnectives.hasMissing()) {
+            List<Map<String, Object>> newCaseDocuments = attachExceptionRecordReference(
```
apereverzin: In a future PR makes sense to wrap this block into a method call something like 'attachNewDocuments'

doncem: lot's of refactoring is screaming around these services and whatnot. best to leave such activity as a separate action once changes in particular areas have cooled down and investigation + refactoring can be done

PR #892, opened 2020-03-10
Payment queue retry config

mdayican:

https://tools.hmcts.net/jira/browse/BPS-1075

### Change description ###

normally servicebus api provides default retry mechanism for transient exceptions.
servicebus exceptions are transient and we got timeout exception which is transient too 
but api did not retried as I understand. again from my investigations retries are bounded with 
operational timeout.  I tried to active retry mechanism in my local but I could not managed to run.
if you think we can active retry mechanism please go head, I am open for ideas.

for now, if we can reduce the timeout and retry manually and we can handle the most cases.

in the api:
default timeout =30 seconds

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/payments/PaymentsPublisher.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/payments/PaymentsPublisher.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.microsoft.azure.servicebus.IMessage;
 import com.microsoft.azure.servicebus.Message;
 import com.microsoft.azure.servicebus.QueueClient;
+import com.microsoft.azure.servicebus.primitives.ServiceBusException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Qualifier;
@@ -44,7 +45,7 @@ public class PaymentsPublisher implements IPaymentsPublisher {
             );
             message.setLabel(cmd.getLabel());
 
-            queueClient.send(message);
+            doSend(message, true);
 
             LOG.info(
                 "Sent message to payments queue. ID: {}, Label: {}, Content: {}",
@@ -57,6 +58,19 @@ public class PaymentsPublisher implements IPaymentsPublisher {
                 "An error occurred when trying to publish message to payments queue.",
                 ex
             );
+        }
+    }
+
+    private void doSend(IMessage message, boolean retry) throws ServiceBusException, InterruptedException {
+        try {
+            queueClient.send(message);
+        } catch (ServiceBusException ex) {
+            if (retry && ex.getIsTransient()) {
+                doSend(message, false);
+            } else {
+                throw ex;
+            }
+
         }
     }
 }
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/payments/PaymentsPublisher.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/payments/PaymentsPublisher.java
@@ -67,7 +67,7 @@ public class PaymentsPublisher implements IPaymentsPublisher {
         } catch (Exception ex) {
             if (retry) {
                 LOG.error(
-                        "Sending to payment queue got error, Message ID: {}, {}. Will retry....",
+                        "Sending to payment queue got error, Message ID: {}. Will retry....",
                         message.getMessageId(),
                         ex
                 );
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/payments/PaymentsPublisher.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/payments/PaymentsPublisher.java
@@ -66,6 +66,7 @@ public class PaymentsPublisher implements IPaymentsPublisher {
             queueClient.send(message);
         } catch (Exception ex) {
             if (retry) {
+                LOG.error("Send to payment queue got error, will retry....", ex);
                 doSend(message, false);
             } else {
                 throw ex;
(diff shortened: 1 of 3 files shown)


apereverzin:
Are log entries related to queue timeouts in prod available? It looks to me that we should use a customized retry policy; having those log entries will give us some insight why default retry policy (which is already in place) did not prevent queue connection errors in prod. One other subject to investigate: what happens if a retry policy is in place but the pod crashes after the first failed attempt to send a message to the queue - will it still be possible to retry?

mdayican:
> Are log entries related to queue timeouts in prod available? It looks to me that we should use a customized retry policy; having those log entries will give us some insight why default retry policy (which is already in place) did not prevent queue connection errors in prod. One other subject to investigate: what happens if a retry policy is in place but the pod crashes after the first failed attempt to send a message to the queue - will it still be possible to retry?


it seems Retry is limited with  the duration of operationtimout. as it gets timeoutException there is no time left to retry.
if pods are crashed we can not do anything.

apereverzin:
In this case if we increase timeout - e.g. to several minutes - will it help? I do not think that immediate retry (as suggested in this PR) is a solution.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/payments/PaymentsPublisher.java
```diff
@@ -59,4 +60,17 @@ public void send(PaymentCommand cmd) {
...
+            queueClient.send(message);
+        } catch (Exception ex) {
+            if (retry) {
+                doSend(message, false);
+            } else {
+                throw ex;
+            }
+
+        }
+    }
```
doncem: from library i see specific mentioning on retry and suppose to use special method:

```java
} catch (ServiceBusException exception) {
    if (exception.getIsTransient() && retry) {
        doSend;
    }
}
```

mdayican: first I thought we should retry for transient exception as library  but I changed my mind as we reduced the timeout  we can try for all exceptions.

doncem: i see. perhaps worth doing at most 10s as callback is for 30s. all operations with idam and ccd will consume most of that time. hopefully retry will fit in

apereverzin: I suggest not to catch any exception from the queue client but investigate how retry policy works and if necessary apply a customized policy.

mdayican: I believe if it works it should work less than a second but I am not sure about the side effects.
we can make operation timeout 10s.

lgonczar: As Andrey said, retry policy would be a nice thing to try first. It would be good if that policy was applied to all publishing, including `processed-envelopes` queue.

mdayican: queueclient has default retry policy already running.

ConnectionStringBuilder

`public RetryPolicy getRetryPolicy()
	{
		return (this.retryPolicy == null ? RetryPolicy.getDefault() : this.retryPolicy);
	}`

nitinprabhu: can we just adjust the timeout  as a first before retrying it manually ?

nitinprabhu: and see how the system behaves in prod.

mdayican: increase or decrease ?

apereverzin: It is necessary to work out how the default retry policy (which is in place and applies some kind of exponentially increasing delays) handles the timeout value. Having these prod failures means that this policy or its params are not suitable for all prod situations. It is worth looking through their code and even running some tests (with mocks) to ensure its behaviour and then change params or the policy itself (some change is definitely necessary).

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/payments/PaymentsPublisher.java
```diff
@@ -59,4 +60,17 @@ public void send(PaymentCommand cmd) {
             );
         }
     }
+
+    private void doSend(IMessage message, boolean retry) throws ServiceBusException, InterruptedException {
+        try {
+            queueClient.send(message);
+        } catch (Exception ex) {
+            if (retry) {
+                doSend(message, false);
```
nitinprabhu: could you please add some logs so that we know the behaviour ?

mdayican: added

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/payments/PaymentsPublisher.java
```diff
@@ -59,4 +60,18 @@ public void send(PaymentCommand cmd) {
             );
         }
     }
+
+    private void doSend(IMessage message, boolean retry) throws ServiceBusException, InterruptedException {
+        try {
+            queueClient.send(message);
+        } catch (Exception ex) {
+            if (retry) {
+                LOG.error("Send to payment queue got error, will retry....", ex);
```
nitinprabhu: Could you please include message id ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/payments/PaymentsPublisher.java
```diff
@@ -66,7 +66,12 @@ private void doSend(IMessage message, boolean retry) throws ServiceBusException,
             queueClient.send(message);
         } catch (Exception ex) {
             if (retry) {
-                LOG.error("Send to payment queue got error, will retry....", ex);
+                LOG.error(
+                        "Sending to payment queue got error, Message ID: {}, {}. Will retry....",
```
doncem: second `{}` is not needed

PR #893, opened 2020-03-10
Add more info to the "duplicate documents" scenario

doncem:
### Change description ###

Gathering more information on edge case which suppose to be removed/resolved. At the moment, logically there should not be a fault but for some reason - it is experienced. Without extra information I cannot confirm case worker is right. And it was meant to be added anyway. Once our QA confirms the workflow further development can be conducted.

NOTE. This is not changing current and previous functionality (before idempotency) per say. Just including extra information to new flow for adoption of idempotency

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/util/ExceptionRecordAttachDocumentConnectives.java
```diff
@@ -22,7 +22,7 @@ public ExceptionRecordAttachDocumentConnectives(
         this.toBeAttachedToTargetCase = toBeAttachedToTargetCase;
     }
 
-    public boolean hasDuplicates() {
+    public boolean hasDuplicatesAndMissing() {
```
apereverzin: Good!

PR #897, opened 2020-03-13
Update connective to check missing docs

doncem:

### Change description ###

Not using `Sets.difference` and simply verifying missing factor with size of current documents in exception record against intersection which was there before.

Another option why I did not experience previous message is service was not deployed to aat environment. Nevertheless, this approach simplifies decision making

Will have to make sure it has been deployed to verify outcome. If still fails - go uncertain path and just delete the exception throw :(

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/util/ExceptionRecordAttachDocumentConnectives.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/util/ExceptionRecordAttachDocumentConnectives.java
@@ -10,31 +10,27 @@ public class ExceptionRecordAttachDocumentConnectives {
     private Set<String> existingInTargetCase;
 
     /**
-     * Exception Record documents which are still missing in target case.
+     * Number of Exception Record documents.
      */
-    private Set<String> toBeAttachedToTargetCase;
+    private int exceptionRecordDocSize;
 
     public ExceptionRecordAttachDocumentConnectives(
         Set<String> existingInTargetCase,
-        Set<String> toBeAttachedToTargetCase
+        int exceptionRecordDocSize
     ) {
         this.existingInTargetCase = existingInTargetCase;
-        this.toBeAttachedToTargetCase = toBeAttachedToTargetCase;
+        this.exceptionRecordDocSize = exceptionRecordDocSize;
     }
 
     public boolean hasDuplicatesAndMissing() {
-        return !existingInTargetCase.isEmpty() && !toBeAttachedToTargetCase.isEmpty();
+        return !existingInTargetCase.isEmpty() && exceptionRecordDocSize != existingInTargetCase.size();
     }
 
     public boolean hasMissing() {
-        return existingInTargetCase.isEmpty() && !toBeAttachedToTargetCase.isEmpty();
+        return existingInTargetCase.isEmpty() && exceptionRecordDocSize > 0;
     }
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/util/ExceptionRecordAttachDocumentConnectives.java
```diff
@@ -10,31 +10,27 @@
...
+        int exceptionRecordDocSize
     ) {
         this.existingInTargetCase = existingInTargetCase;
-        this.toBeAttachedToTargetCase = toBeAttachedToTargetCase;
+        this.exceptionRecordDocSize = exceptionRecordDocSize;
     }
 
     public boolean hasDuplicatesAndMissing() {
-        return !existingInTargetCase.isEmpty() && !toBeAttachedToTargetCase.isEmpty();
+        return !existingInTargetCase.isEmpty() && exceptionRecordDocSize != existingInTargetCase.size();
```
apereverzin: Not sure I understand the logic behind this: for me equal size does not mean the sets are equal.

doncem: please re-investigate the condition and what parameters are used. there's nothing wrong with logic here

apereverzin: yes, indeed

nitinprabhu: could you specify a situation when this method will return true ?

nitinprabhu: and even when it returns true message doesn't seems right it says Documents already attached ....

doncem: the scenario is what method says. this is logical scenario. not the real life one as i think you mentioned either all docs go in or none. so there should not be a case for partial attachments happening. nevertheless, case worker **can** add/remove documents before submitting a form. so theoretically it's possible to get here.

regarding error message - it can be updated. perhaps even in this PR. want me to do that?

nitinprabhu: If the caseworker does it intentionally we should allow them to update the case ? With the current implementation it will block it. may be an edge case but imagine a caseworker removed a doc mistakenly and submitted the event and then cancels the browser then this would block it ?
This should be worth checking with business what they expect. But I don't think blocking attaching of documents is fine

doncem: that's what i initially said - this should be business decision ;). right now all these chain of PRs are not removing the exception throw just working around the idempotency integration. for correctness - can do the improved exception/message now if you feel like ;)

doncem:
See reference

PR #909, opened 2020-03-31
Create a service for handling reclassification callbacks

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-1108

### Change description ###

Create a service for handling reclassification callbacks. This service will be called by a new callback-handling method in CcdCallbackController.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/ReclassifyCallbackService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/ReclassifyCallbackService.java
@@ -32,7 +32,7 @@ public class ReclassifyCallbackService {
         String jurisdiction = exceptionRecordDetails.getJurisdiction();
 
         log.info(
-            "Reclassifying exception record {}. Jurisdiction: {}. User: {}",
+            "Reclassifying exception record. Exception record ID: {}. Jurisdiction: {}. User ID: {}",
             exceptionRecordId,
             jurisdiction,
             userId
@@ -42,7 +42,8 @@ public class ReclassifyCallbackService {
 
         if (errors.isEmpty()) {
             log.info(
-                "Returning successful reclassification result for exception record. ID: {}. Jurisdiction: {}. User: {}",
+                "Returning successful reclassification result for exception record. "
+                    + "Exception record ID: {}. Jurisdiction: {}. User ID: {}",
                 exceptionRecordId,
                 jurisdiction,
                 userId
@@ -53,9 +54,11 @@ public class ReclassifyCallbackService {
             // Logging validation failure as error, because the caseworker should have no way of causing it.
             // If it happens, the team needs to be alerted.
             log.error(
-                "Validation failed for exception record reclassification. ID: {}. Jurisdiction: {}. Errors: {}",
+                "Validation failed for exception record reclassification. "
+                + "Exception record ID: {}. Jurisdiction: {}. User ID: {}. Errors: {}",
                 exceptionRecordId,
                 jurisdiction,
+                userId,
                 String.join(", ", errors)
             );
 
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/ReclassifyCallbackService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/ReclassifyCallbackService.java
@@ -1,0 +1,116 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd;
+
+import com.google.common.collect.Maps;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Service;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/ReclassifyCallbackService.java
```diff
@@ -0,0 +1,116 @@
...
+
+    private static final String REQUIRED_INITIAL_STATE = "ScannedRecordReceived";
+    private static final String REQUIRED_INITIAL_CLASSIFICATION = NEW_APPLICATION.name();
+
+    public ProcessResult reclassifyExceptionRecord(CaseDetails exceptionRecordDetails, String userId) {
+        Long exceptionRecordId = exceptionRecordDetails.getId();
+        String jurisdiction = exceptionRecordDetails.getJurisdiction();
+
+        log.info(
+            "Reclassifying exception record {}. Jurisdiction: {}. User: {}",
```
nitinprabhu: User -> UserId ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/ReclassifyCallbackService.java
```diff
@@ -0,0 +1,116 @@
...
+                jurisdiction,
+                userId
+            );
+
+            return new ProcessResult(updateClassification(exceptionRecordDetails.getData()));
+        } else {
+            // Logging validation failure as error, because the caseworker should have no way of causing it.
+            // If it happens, the team needs to be alerted.
+            log.error(
+                "Validation failed for exception record reclassification. ID: {}. Jurisdiction: {}. Errors: {}",
```
nitinprabhu: Id -> ExceptionRecordId ?

lgonczar:
Changed log messages

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/ReclassifyCallbackService.java
```diff
@@ -0,0 +1,119 @@
...
+        List<Optional<String>> validationErrors = asList(
+            validateState(exceptionRecordDetails),
+            validateClassification(exceptionRecordDetails)
+        );
+
+        return validationErrors
+            .stream()
+            .filter(Optional::isPresent)
+            .map(Optional::get)
+            .collect(toList());
```
doncem: Spring has a nice wrapper called `Optionals`, it could transform this piece of code to something as following:

```java
return Optionals
    .toStream(
        alidateState(exceptionRecordDetails),
        alidateClassification(exceptionRecordDetails)
    )
    .collect(toList());
```

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/ReclassifyCallbackService.java
```diff
@@ -0,0 +1,119 @@
...
+                    "This exception record's journey classification is '%s'. "
+                        + "Reclassification is only possible from '%s' classification.",
+                    classification,
+                    REQUIRED_INITIAL_CLASSIFICATION
+                )
+            );
+        } else {
+            return Optional.empty();
+        }
+    }
```
doncem: Kind of beats the purpose of optional response in conjunction with if/else conditions. would it look ugly if we try something functionally here?

```java
return Optional.ofNullable(...)
    .map(cast to string if necessary)
    .filter(v -> !v.equals(...))
    .map(format validation fail message);
```

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/ReclassifyCallbackService.java
lgonczar: Nice one, will do

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/ReclassifyCallbackService.java
lgonczar: I agree that once you've got an Optional, especially in Java 11, using if-else may make little sense, but this code creates an Optional instead of dealing with an existing one. While it's possible to replace pretty much every kind of logic with a chain of transformations, I think it would be a bit artificial here, given how simple this case is.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/ReclassifyCallbackService.java
lgonczar: ... although we would need to import a new library for this.

doncem: ah. possibly it's badly implemented as i think it comes from data. tested on send letter project - had that window open. ignore then. unless orchestrator has smth handy from its' dependencies

PR #912, opened 2020-04-01
Introduce validating rest template

doncem:

### Change description ###

First out of 2 PRs: `ValidatableRestTemplate` ready to be used in custom clients. Which is transformation client as the ticket refers.

The following PR is self-explanatory given this is introduced into codebase

```
[ ] Yes
[x] No
```


doncem:
will squash into client itself

PR #913, opened 2020-04-02
Validate user ID in reclassification endpoint

lgonczar:
### Change description ###

Validate user ID in reclassification endpoint in order to prevent attacks through its value (user ID is logged). This should fix currently broken build (due to sonarcloud).

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CcdCallbackController.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CcdCallbackController.java
@@ -89,7 +89,9 @@ public class CcdCallbackController {
         @RequestBody CcdCallbackRequest callbackRequest,
         @RequestHeader(value = USER_ID, required = false) String userId
     ) {
-        if (callbackRequest.getCaseDetails() == null) {
+        if (userId != null && userId.matches("^.*[\\n|\\r|\\t]+.*$")) {
+            throw new InvalidRequestException("User ID contains invalid characters");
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CcdCallbackController.java
```diff
@@ -89,7 +89,9 @@ public CallbackResponse reclassifyExceptionRecord(
         @RequestBody CcdCallbackRequest callbackRequest,
         @RequestHeader(value = USER_ID, required = false) String userId
     ) {
-        if (callbackRequest.getCaseDetails() == null) {
+        if (userId != null && userId.matches("^.*[\\n|\\r|\\t]+.*$")) {
```
rkondratowicz, line 92: is this the only controller that should perform this validation?

lgonczar: Ideally, there should be more, but this one definitely needs that, as this user ID is logged. It's a quick fix. We may add this validation to other endpoints later.

mdayican: do we use any validation api ?

lgonczar: we use vavr + custom validator for validating exception record details

PR #915, opened 2020-04-02
Integrate `ValidatableRestTemplate` to `TransformationClient`

doncem:

### Change description ###

Accompanying integration PR alongside the #912. It is also based on top of that PR as it is a hard dependency.

```
[ ] Yes
[x] No
```


doncem:
feedback arrived split second after raised this one

PR #916, opened 2020-04-02
Add validation to transformation response

doncem:

```
[x] Yes - different behaviour - more restrictions
[x] No - should've happened anyway
```


src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CreateCaseCallbackTest.java
```diff
@@ -134,6 +136,29 @@ void should_create_case_if_classification_exception_with_documents_and_ocr_data(
...
+        "bad-case-data-field-is-null.json",
+        "bad-case-details-field-is-null.json",
+        "bad-case-type-id-field-is-null.json",
+        "bad-event-id-field-is-null.json",
+        "bad-no-case-data-field.json",
+        "bad-no-case-details-field.json",
+        "bad-no-case-type-id-field.json",
+        "bad-no-event-id-field.json"
+    })
+    void should_(String transformationResponseBody) {
```
rkondratowicz, line 150: something missing in test name

doncem: yep. development mode. 1 moment

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdNewCaseCreator.java
```diff
@@ -114,6 +115,18 @@ public CreateCaseResult createNewCase(
...
+                "Invalid response received from transformation endpoint. "
+                    + "Service: %s, exception record: %s, violations: %s",
+                configItem.getService(),
+                exceptionRecord.id,
+                exception.getMessage()
+            );
+
+            log.error(message, exception);
+
+            throw new CallbackException(message, exception);
```
lgonczar, line 129: Throwing the exception is enough. Logging the error additionally will result in two error entries saying the same thing.

lgonczar:
We also need to make sure exception record data is a map, because we're adding an entry to that map. The solution should support future validation of the structure of that map, as we'll look into scanned documents and adding exception record reference to them.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdNewCaseCreator.java
doncem: do we also want to remove same behaviour in all current catch blocks?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdNewCaseCreator.java
```diff
@@ -206,9 +217,8 @@ private long createNewCaseInCcd(
         }
     }
 
-    @SuppressWarnings("unchecked")
-    private Map<String, Object> caseDataWithExceptionRecordId(Object caseData, String exceptionRecordId) {
-        Map<String, Object> data = new HashMap((Map<String, Object>) caseData);
+    private Map<String, Object> caseDataWithExceptionRecordId(Map<String, Object> caseData, String exceptionRecordId) {
+        Map<String, Object> data = new HashMap<>(caseData);
```
nitinprabhu, line 221: do we need to convert this again to new map ? Can we not use caseData as it is already a map ?

doncem: iirc, this was so to not use original object as it conflicted with end result. i'm sure this can be tracked backwards... and here it is: #785

PR #921, opened 2020-04-06
Configure case update URL for Probate

lgonczar:
### Change description ###

Configure case update URL for Probate.

```
[ ] Yes
[x] No
```


lgonczar:
> preview chart needs update to use AAT ?

I was thinking that it shouldn't. But then, it still has to be configured. Will add.

infrastructure/demo.tfvars
```diff
@@ -3,5 +3,6 @@ idam_client_redirect_uri = "https://bulk-scan-orchestrator-demo.service.core-com
 supported_jurisdictions = ["SSCS", "BULKSCAN", "DIVORCE", "PROBATE", "CMC"]
 
 transformation_url_probate = "http://probate-back-office-demo.service.core-compute-demo.internal"
+update_url_probate = "http://probate-back-office-demo.service.core-compute-demo.internal"
```
mdayican, line 6: this needs to be in flux

lgonczar: yes, but for now also here

PR #937, opened 2020-04-23
Use service case type id in the startEvent call for the attachScannedDocsWithOcr event

aliveni:
https://tools.hmcts.net/jira/browse/BPS-1180

### Change description ###
Retrieve the service case before starting the "attachScannedDocsWithOcr" event so that the service case type id can be fetched and used in the startEvent call.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
@@ -447,6 +447,8 @@ public class AttachCaseCallbackService {
             callBackEvent.exceptionRecordId
         );
 
+        CaseDetails targetCase = ccdApi.getCase(targetCaseCcdRef, callBackEvent.exceptionRecordJurisdiction);
+
         ServiceConfigItem serviceConfigItem = getServiceConfig(exceptionRecordDetails);
         ProcessResult processResult = ccdCaseUpdater.updateCase(
             callBackEvent.exceptionRecord,
@@ -454,7 +456,8 @@ public class AttachCaseCallbackService {
             ignoreWarnings,
             callBackEvent.idamToken,
             callBackEvent.userId,
-            targetCaseCcdRef
+            targetCaseCcdRef,
+            targetCase.getCaseTypeId()
         );
 
         if (!processResult.getErrors().isEmpty() || !processResult.getWarnings().isEmpty()) {
--- a/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdaterTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdaterTest.java
@@ -261,7 +261,7 @@ class CcdCaseUpdaterTest {
             .willReturn(noWarningsUpdateResponse);
         initMockData();
         prepareMockForSubmissionEventForCaseWorker().willThrow(
-            new FeignException.BadRequest("Msg", mock(Request.class), "Body".getBytes())
+                new FeignException.BadRequest("Msg", mock(Request.class), "Body".getBytes())
         );
 
         // when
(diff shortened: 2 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -447,14 +447,17 @@ private void updateSupplementaryEvidence(
             callBackEvent.exceptionRecordId
         );
 
+        CaseDetails targetCase = ccdApi.getCase(targetCaseCcdRef, callBackEvent.exceptionRecordJurisdiction);
```
nitinprabhu, line 450: if this call fails will the caseworker shown an error screen. If this call fails they should be shown with a message to retry.

nitinprabhu: Also test case is missing when this call throws an exception

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdaterTest.java
```diff
@@ -466,7 +478,7 @@ void updateCase_should_handle_feign_exception_from_start_event() {
             anyString(),
             anyString()
         ))
-            .willThrow(new FeignException.MethodNotAllowed("Msg", mock(Request.class),  "Body".getBytes()));
+            .willThrow(new FeignException.MethodNotAllowed("Msg", mock(Request.class), "Body".getBytes()));
```
nitinprabhu, line 481: not needed change ?

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdaterTest.java
```diff
@@ -534,7 +548,7 @@ void updateCase_should_handle_exception_when_start_event_returns_not_found_respo
             anyString()
         ))
             .willThrow(
-                    new FeignException.NotFound("case not found",  mock(Request.class), "Body".getBytes())
+                new FeignException.NotFound("case not found", mock(Request.class), "Body".getBytes())
```
nitinprabhu, line 551: same here formatting change not needed ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -447,14 +447,17 @@ private void updateSupplementaryEvidence(
...
         ServiceConfigItem serviceConfigItem = getServiceConfig(exceptionRecordDetails);
         ProcessResult processResult = ccdCaseUpdater.updateCase(
             callBackEvent.exceptionRecord,
             serviceConfigItem,
             ignoreWarnings,
             callBackEvent.idamToken,
             callBackEvent.userId,
-            targetCaseCcdRef
+            targetCaseCcdRef,
+            targetCase.getCaseTypeId()
```
doncem, line 460: isn't this available from `callBackEvent` (same as `targetCaseCcdRef`)?

aliveni: `targetCaseCcdRef` is for the "search case reference type"  which can be either CCD case reference or external case reference

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
aliveni: I'll update the pr

aliveni: When the ccd call fails, the error will be displayed to the caseworker. see https://github.com/hmcts/bulk-scan-orchestrator/blob/a0b8a51ac54e9a57863b7ca7c5c3483596a90bee/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java#L235-L248

I've updated the integration tests for invalid case reference and case not found scenarios.

PR #942, opened 2020-04-27
add remove access token from cache functionality

mdayican:
### JIRA link (if applicable) ###

https://tools.hmcts.net/jira/browse/BPS-1122

### Change description ###

part 2:https://github.com/hmcts/bulk-scan-orchestrator/pull/940
this is part of part2,
add remove accesstoken from cache functionality.

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #943, opened 2020-04-27
add userdetail cache

mdayican:

https://tools.hmcts.net/jira/browse/BPS-1122

### Change description ###

Part 3: add userdetails cache
to remove a user detail from cache whenever token invalidated or expired removalListener used.
maxsize of userdetails 200 hardcoded next pr I will make it configurable.
caffeine manage the eviction if  max size set:
 * As the cache size grows close to the maximum, the cache evicts entries that are less likely to
   * be used again. For example, the cache may evict an entry because it hasn't been used recently
   * or very often.
but in our case it should not happen as it should remove from user cache whenever token removed from cache.

Part2 last pr: https://github.com/hmcts/bulk-scan-orchestrator/pull/942

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/cache/IdamCachedClient.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/cache/IdamCachedClient.java
@@ -57,6 +57,10 @@ public class IdamCachedClient {
         accessTokenCache.invalidate(jurisdiction);
     }
 
+    public void cleanUpAccessTokenCache() {
+        accessTokenCache.cleanUp();
+    }
+
     private CachedIdamToken retrieveToken(String jurisdiction) {
         log.info("Retrieve access token for jurisdiction: {} from IDAM", jurisdiction);
         Credential user = users.getUser(jurisdiction);
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/cache/IdamCachedClient.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/cache/IdamCachedClient.java
@@ -5,11 +5,13 @@ import com.auth0.jwt.interfaces.Claim;
 import com.auth0.jwt.interfaces.DecodedJWT;
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.RemovalCause;
 import org.springframework.stereotype.Service;
 import uk.gov.hmcts.reform.authorisation.exceptions.InvalidTokenException;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.services.idam.Credential;
 import uk.gov.hmcts.reform.bulkscan.orchestrator.services.idam.JurisdictionToUserMapping;
 import uk.gov.hmcts.reform.idam.client.IdamClient;
+import uk.gov.hmcts.reform.idam.client.models.UserDetails;
 
 @Service
 public class IdamCachedClient {
@@ -18,6 +20,7 @@ public class IdamCachedClient {
     public static final String EXPIRES_IN = "expires_in";
 
     private static Cache<String, CachedIdamToken> accessTokenCache;
+    private static Cache<String, UserDetails> userDetailsCache;
 
     private final IdamClient idamClient;
     private final JurisdictionToUserMapping users;
@@ -31,6 +34,11 @@ public class IdamCachedClient {
         this.users = users;
         this.accessTokenCache = Caffeine.newBuilder()
             .expireAfter(accessTokenCacheExpiry)
+            .removalListener(this::onCachedIdamTokenRemoval)
+            .build();
+
+        this.userDetailsCache =  Caffeine.newBuilder()
+            .maximumSize(200)
             .build();
     }
 
@@ -68,6 +76,22 @@ public class IdamCachedClient {
         }
 
         return expires.asLong();
+    }
+
+    private void onCachedIdamTokenRemoval(
+        String jurisdiction,
+        CachedIdamToken cachedIdamToken,
+        RemovalCause cause
+    ) {
+        userDetailsCache.invalidate(cachedIdamToken.accessToken);
+    }
+
+    public UserDetails getUserDetails(String accessToken) {
+        return this.userDetailsCache.get(accessToken, this::retrieveUserDetails);
+    }
+
+    private UserDetails retrieveUserDetails(String accessToken) {
+        return idamClient.getUserDetails(accessToken);
     }
 
 }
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/cache/IdamCachedClient.java
```diff
@@ -70,4 +78,20 @@ private long stripExpiryFromBearerToken(String tokenWithBearer) {
...
+    ) {
+        userDetailsCache.invalidate(cachedIdamToken.accessToken);
+    }
+
+    public UserDetails getUserDetails(String accessToken) {
+        return this.userDetailsCache.get(accessToken, this::retrieveUserDetails);
+    }
+
+    private UserDetails retrieveUserDetails(String accessToken) {
+        return idamClient.getUserDetails(accessToken);
```
lgonczar, line 94: Worth logging this bit.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/cache/IdamCachedClient.java
```diff
@@ -70,4 +78,20 @@ private long stripExpiryFromBearerToken(String tokenWithBearer) {
         return expires.asLong();
     }
 
+    private void onCachedIdamTokenRemoval(
+        String jurisdiction,
+        CachedIdamToken cachedIdamToken,
+        RemovalCause cause
+    ) {
+        userDetailsCache.invalidate(cachedIdamToken.accessToken);
```
lgonczar, line 86: It would be good to log this.

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/cache/IdamCachedClientTest.java
```diff
@@ -196,4 +206,56 @@ void should_create_new_token_when_token_removed_from_cache() {
...
+
+    @Test
+    public void should_get_userDetail_when_no_error() {
+        given(idamClient.getUserDetails(JWT)).willReturn(USER_DETAILS);
+
+        UserDetails userDetails1 =
+            idamCachedClient.getUserDetails(JWT);
+
+        assertThat(userDetails1).usingRecursiveComparison().isEqualTo(USER_DETAILS);
+        verify(idamClient).getUserDetails(any());
```
lgonczar, line 219: Just in case, best to verify agains the expected user name.

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/cache/IdamCachedClientTest.java
```diff
@@ -196,4 +206,56 @@ void should_create_new_token_when_token_removed_from_cache() {
...
+    @Test
+    public void should_invalidate_userDetail_when_cached_token_removed_from_cache() {
+        String jurisdiction1 = "divorce";
+
+        given(users.getUser(jurisdiction1)).willReturn(new Credential(USERNAME, PASSWORD));
+        given(idamClient.authenticateUser(USERNAME, PASSWORD)).willReturn(JWT, JWT2);
+
+        UserDetails expectedUserDetails1 =  new UserDetails("12","q@a.com","","",null);
+
+        given(idamClient.getUserDetails(JWT)).willReturn(expectedUserDetails1, USER_DETAILS);
```
lgonczar, line 245: Suggestion: consider creating another var, called `expectedUserDetails2` or `expectedUserDetailsAfter...`, even if the object behind it is USER_DETAILS. It will make this test easier to follow.

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/cache/IdamCachedClientTest.java
```diff
@@ -196,4 +206,56 @@ void should_create_new_token_when_token_removed_from_cache() {
...
+
+        String token1 =
+            idamCachedClient.getAccessToken(jurisdiction1);
+        UserDetails userDetails = idamCachedClient
```
lgonczar, line 249: How about `userDetailsBefore` or `userDetailsBeforeInvalidating`? Then `userDetails2` would be `userDetailsAfter...`

lgonczar:
I've added tiny suggestions in tests, to consider, but logging would really be useful.

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/cache/IdamCachedClientTest.java
mdayican: I just want to very getUserDetails method called only once with any value

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/cache/IdamCachedClient.java
mdayican: I will add log with next pr with configurable cache maxsize

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/cache/IdamCachedClient.java
```diff
@@ -31,20 +37,28 @@ public IdamCachedClient(
...
         return cachedIdamToken.accessToken;
     }
 
     public void removeAccessTokenFromCache(String jurisdiction) {
+        log.info("Remove access token from cache for jurisdiction: {} ", jurisdiction);
         accessTokenCache.invalidate(jurisdiction);
     }
 
     private CachedIdamToken retrieveToken(String jurisdiction) {
+        log.info("Retrieve access token for jurisdiction: {} ", jurisdiction);
```
lgonczar, line 61: This log entry is very similar to `Get access token ...`. How about adding `... from IdAM` at the end of the message to make it easier to distinguish between these two situations?

Similar things with getting user details.

mdayican: chnaged

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/cache/IdamCachedClient.java
```diff
@@ -31,20 +37,32 @@ public IdamCachedClient(
         this.users = users;
         this.accessTokenCache = Caffeine.newBuilder()
             .expireAfter(accessTokenCacheExpiry)
+            .removalListener(this::onCachedIdamTokenRemoval)
+            .build();
+
+        this.userDetailsCache =  Caffeine.newBuilder()
+            .maximumSize(200)
```
nitinprabhu, line 44: is there a reason this value is high. We would have 5 keys /values at any given time ?

mdayican: it will be configurable. but 5 is not enough as we better give some buffer to maintain the cache

PR #944, opened 2020-04-27
Fix cache invalidate test

mdayican:
### Change description ###
Fix cache invalidate test.
cache invalite test sometime fails as  it seems there is sync. issue with threads.
hope cachecleanup triggers maintaining the cache.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/cache/IdamCachedClientTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/cache/IdamCachedClientTest.java
@@ -233,28 +233,4 @@ class IdamCachedClientTest {
         verify(idamClient).getUserDetails(any());
     }
 
-    @Test
-    public void should_invalidate_userDetail_when_cached_token_removed_from_cache() {
-        String jurisdiction1 = "divorce";
-
-        given(users.getUser(jurisdiction1)).willReturn(new Credential(USERNAME, PASSWORD));
-        given(idamClient.authenticateUser(USERNAME, PASSWORD)).willReturn(JWT, JWT2);
-
-        UserDetails expectedUserDetails1 =  new UserDetails("12","q@a.com","","",null);
-        UserDetails expectedUserDetails2 = USER_DETAILS;
-
-        given(idamClient.getUserDetails(JWT)).willReturn(expectedUserDetails1, expectedUserDetails2);
-
-        String token1 = idamCachedClient.getAccessToken(jurisdiction1);
-        UserDetails userDetailsBefore = idamCachedClient.getUserDetails(token1);
-        assertThat(userDetailsBefore).isEqualTo(expectedUserDetails1);
-
-        idamCachedClient.removeAccessTokenFromCache(jurisdiction1);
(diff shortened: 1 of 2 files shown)


mdayican:
> Why not revert the commit https://build-beta.platform.hmcts.net/job/HMCTS_BSP/job/bulk-scan-orchestrator/job/master/49/changes and then fix the test and functionality together ?

as we do not use it, we do not need to revert whole thing. this is just to leave the master in peace.

nitinprabhu:
> > Why not revert the commit https://build-beta.platform.hmcts.net/job/HMCTS_BSP/job/bulk-scan-orchestrator/job/master/49/changes and then fix the test and functionality together ?
> 
> as we do not use it, we do not need to revert whole thing. this is just to leave the master in peace.

Removing test will make master green but we will end up something in master which is not working consistently. So I would suggest even if it is not getting used let's remove it and merge with tests once it's fixed :)

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/cache/IdamCachedClientTest.java
```diff
@@ -250,7 +251,7 @@ public void should_invalidate_userDetail_when_cached_token_removed_from_cache()
         assertThat(userDetailsBefore).isEqualTo(expectedUserDetails1);
 
         idamCachedClient.removeAccessTokenFromCache(jurisdiction1);
-        idamCachedClient.cleanUpAccessTokenCache();
+        idamCachedClient.cleanUpCaches();
```
lgonczar, line 254: I realise it was like that before, but does this test verify the behaviour of `removeAccessTokenFromCache` if we clean the whole cache afterwards?

mdayican: it is not cleaning the cache from javadoc of method:
  /**
   * Performs any pending maintenance operations needed by the cache. Exactly which activities are
   * performed -- if any -- is implementation-dependent.
   */

mdayican: first i tried 3 sec sleep, it passed test but it was luck.

lgonczar: Not sure we need `cleanUpCaches` to exist at all

lgonczar: aaa, I see

mdayican:
> Suggestion for a future PR: ideally, get rid of the public cleanup method, so that the cache is easier to use. If not, change its name, because cleanup often means remove everything from.

tidyUpCache?
maintainCache?

lgonczar:
> tidyUpCache?
> maintainCache?
> I can fix now, as I am doing some commits to trigger builds to see it always pass.

`forceCacheUpdate()`?

mdayican:
open 
https://github.com/hmcts/bulk-scan-orchestrator/pull/945

PR #945, opened 2020-04-27
use cache write instead of removal listener

mdayican:
### Change description ###

sometimes invalidate test  is failing as removal listener is working asynchronously.
replace listener with cachewriter which is synchronously runs on removal.

```
[ ] Yes
[x] No
```


mdayican:
> Are there tests to verify this listener is invoked and cache works as expected ?

if the test pass it verifies it invoked userdetails invalidate as user details cache do size based eviction

nitinprabhu:
> > Are there tests to verify this listener is invoked and cache works as expected ?
> 
> if the test pass it verifies it invoked userdetails invalidate as user details cache do size based eviction

ok let's make sure we add such details in description or comment so that it is easier for reviewer to follow :)

PR #956, opened 2020-05-01
Remove validation on event_id in update case response

aliveni:
https://tools.hmcts.net/jira/browse/BPS-1180

### Change description ###
Removed validation on event_id in case update details

```
[ ] Yes
[x] No
```


aliveni:
> just wanted to make sure for both supplementary_evidence and supplementary_evidence_with_ocr update we use barcoded event ids ?
> 
> PR looks good once you confirm above

we have validation on for supplementary_evidence_with_ocr response only. For `supplementary_evidence` classification, we are not validating the response.

doncem:
the update endpoint was suppose to be used only for `..with_ocr` classification

PR #960, opened 2020-05-07
Update remaining time when value in cache is updated

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/idam/cache/IdamCacheExpiry.java
```diff
@@ -35,7 +35,7 @@ public long expireAfterUpdate(
         long currentTime,
         @NonNegative long currentDuration
     ) {
-        return currentDuration;
+        return TimeUnit.SECONDS.toNanos(cachedIdamCredential.expiresIn - refreshTokenBeforeExpiry);
```
mdayican, line 38: as we do not refresh the token if this works it will change the timer for existing token which is wrong.
for example remaining time is 300 for current token if this works it will again make it 28800.

rkondratowicz: this is for replacing the token

mdayican: where do you replace the token ?

lgonczar: I guess we should either have correct implementation of this method or always throw exception.

mdayican: I think returning `currentDuration` is not wrong. if some thing is triggering this method is not us.we are not refreshing the token. if you think returning currentDuration is wrong , throwing exception might a better option.

PR #968, opened 2020-05-13
Add validator for attaching scanned docs to target case

doncem:

[Handle missing and duplicate documents for attaching exception record with supplementary evidence](https://tools.hmcts.net/jira/browse/BPS-1095)

### Change description ###

More detailed validation for scanned documents. Only throw exception when documents are truly attached without expected exception record reference. Otherwise - all good

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/callback/AttachScannedDocumentsValidator.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/callback/AttachScannedDocumentsValidator.java
@@ -9,7 +9,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
-import java.util.function.Function;
+import java.util.function.Predicate;
 
 import static java.util.stream.Collectors.toSet;
 import static org.slf4j.LoggerFactory.getLogger;
@@ -106,13 +106,13 @@ public final class AttachScannedDocumentsValidator {
     }
 
     private static Set<String> findDocumentsIntersection(
-        Function<Map<String, Object>, Boolean> filterOnTargetCaseDocuments,
+        Predicate<Map<String, Object>> filterOnTargetCaseDocuments,
         List<Map<String, Object>> exceptionRecordDocuments,
         List<Map<String, Object>> targetCaseDocuments
     ) {
         Set<String> dcnsOfCaseDocumentsFromOtherSources = targetCaseDocuments
             .stream()
-            .filter(filterOnTargetCaseDocuments::apply)
+            .filter(filterOnTargetCaseDocuments)
             .map(Documents::getDocumentId)
             .collect(toSet());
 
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/callback/AttachScannedDocumentsValidator.java
```diff
@@ -0,0 +1,126 @@
...
+
+import static java.util.stream.Collectors.toSet;
+import static org.slf4j.LoggerFactory.getLogger;
+
+public final class AttachScannedDocumentsValidator {
+
+    private static final Logger log = getLogger(AttachScannedDocumentsValidator.class);
+
+    private AttachScannedDocumentsValidator() {
+        // utility class constructor
```
apereverzin, line 22: It should be a normal component with non-static method

doncem: after brief discussion i don't mind at all if its component or not. will convert to component ;)

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/callback/AttachScannedDocumentsValidatorTest.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static java.util.Collections.singletonList;
+import static java.util.Collections.singletonMap;
+import static org.assertj.core.api.Assertions.assertThatCode;
+
+class AttachScannedDocumentsValidatorTest {
+
+    private static final AttachScannedDocumentsValidator VALIDATOR = new AttachScannedDocumentsValidator();
```
apereverzin, line 16: After the class under test is a component setUp() method with Before annotation can be used

doncem: I would use it if some mocked dependencies are needed which is not the case here so no need for extra lines of code just to initialise class for every test. It will not hold any state either

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/callback/AttachScannedDocumentsValidator.java
```diff
@@ -0,0 +1,128 @@
...
+
+import static java.util.stream.Collectors.toSet;
+import static org.slf4j.LoggerFactory.getLogger;
+
+@Component
+public class AttachScannedDocumentsValidator {
+
+    private static final Logger log = getLogger(AttachScannedDocumentsValidator.class);
+
+    public AttachScannedDocumentsValidator() {
```
mdayican, line 23: why do we need this?

PR #973, opened 2020-05-14
Add config for SSCS transformation url

aliveni:
https://tools.hmcts.net/jira/browse/BPS-1241

### Change description ###
Flux config for env var - https://github.com/hmcts/cnp-flux-config/pull/3550

```
[ ] Yes
[x] No
```


aliveni:
> same comment as in processor. perhaps need default chart value as false otherwise client will use this url

If the callback URL is `sscs-bulk-scan` url , orchestrator won't be in picture and will not try to transform.

aliveni:
> same comment as in processor. perhaps need default chart value as false otherwise client will use this url

Added default value. Flux or https://github.com/hmcts/cnp-flux-config/pull/3555

PR #1002, opened 2020-06-12
override env. var

mdayican:
### Change description ###

seems it does not override spring property with the env var.
override env. var

```
[ ] Yes
[x] No
```


mdayican:
> It's not just this url. Issue is with all env vars

issue solved but still we can merge it ?

PR #1012, opened 2020-06-17
Fix build - Make queue name configurable per environment

nitinprabhu:
### Change description ###

- Queue name cannot be hardcoded anymore as we use different queue for staging.

```
[ ] Yes
[x ] No
```


nitinprabhu:
> hmm.. kind of beats the purpose of having another "staging" resource present. could just re-use existing with these amended `staging` suffixed queues

You would still need some env var to figure if it is staging pod and then append `-staging` which means config and code change.

doncem:
i'm not doubting anything here ;). just observing as a third person

PR #1014, opened 2020-06-18
[IGNORE] Test Github workflow

nitinprabhu:


doncem:
who enabled github action for prs? it is only meant for master builds. we suppose to only require for successful jenkins build. atm there is no action present which would be useful for PR

PR #1015, opened 2020-06-18
Use searchCaseReference as the target case reference when attaching an exception record

aliveni:
https://tools.hmcts.net/jira/browse/BPS-1275

### Change description ###
- When attaching an Exception Record to an existing case, 
   - Validate `searchCaseReference` value in the callback request as the target case reference.
   - Validate  `attachToCaseReference` value when `searchCaseReference`  value doesn't exist (so 
   that the current functionality will not be broken)

- Added a functional test to verify if `searchCaseReference` takes priority when both  `searchCaseReference` and `attachToCaseReference` have values in the callback request.

**Next steps:**
- update Exception record definitions to include searchCaseReference field
- Update callback service to attach the Exception Record to `searchCaseReference` and 
remove validation on `attachToCaseReference`

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
@@ -206,7 +206,7 @@ class AttachExceptionRecordToExistingCaseTest {
         // set searchCaseReference to the callback request data
         // and searchCaseReferenceType and attachToCaseReference doesn't exist
         invokeCallbackEndpointWithSearchCaseRefAndAttachToCaseRef(
-            String.valueOf(targetCase.getId()),
+            Long.toString(targetCase.getId()),
             exceptionRecord
         ).jsonPath()
             .getList("errors")
--- a/src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
@@ -194,6 +194,39 @@ class AttachExceptionRecordToExistingCaseTest {
     }
 
     @Test
+    public void should_attach_exception_record_to_case_by_searchCaseReference_and_attachToCaseReference()
+        throws Exception {
+        //given
+        // set searchCaseReference as targetCase id
+        CaseDetails targetCase = ccdCaseCreator.createCase(emptyList(), Instant.now());
+
+        // set attachToCaseReference value with attachToCaseDetails id, which should be ignored
+        CaseDetails attachToCaseDetails = ccdCaseCreator.createCase(emptyList(), Instant.now());
+
+        CaseDetails exceptionRecord =
+            createExceptionRecord("envelopes/supplementary-evidence-envelope.json");
+
+        // when
+        // set searchCaseReference and attachToCaseReference to callback request
+        // and searchCaseReferenceType doesn't exist
+        invokeCallbackEndpointWithSearchCaseRefAndAttachToCaseRef(
+            targetCase,
+            attachToCaseDetails,
+            exceptionRecord
+        ).jsonPath()
+            .getList("errors")
+            .isEmpty();
+
+        //then
+        await("Exception record is attached to the case with searchCaseReference")
+            .atMost(60, TimeUnit.SECONDS)
+            .pollDelay(2, TimeUnit.SECONDS)
+            .until(() -> isExceptionRecordAttachedToTheCase(targetCase, 1));
+
+        verifyExistingCaseIsUpdatedWithExceptionRecordData(targetCase, exceptionRecord, 1);
+    }
+
+    @Test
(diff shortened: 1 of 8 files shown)


src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -193,6 +193,39 @@ public void should_attach_exception_record_to_case_by_attach_to_case_reference()
         verifyExceptionRecordAttachesToCase(null);
     }
 
+    @Test
+    public void should_attach_exception_record_to_case_by_searchCaseReference_and_attachToCaseReference()
```
nitinprabhu, line 197: how are both `searchCaseReference` and `attachToCaseReference` going to be set ? Is it by CCD but I am not sure how both fields will be populated

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -193,6 +193,39 @@ public void should_attach_exception_record_to_case_by_attach_to_case_reference()
...
+        throws Exception {
+        //given
+        // set searchCaseReference as targetCase id
+        CaseDetails targetCase = ccdCaseCreator.createCase(emptyList(), Instant.now());
+
+        // set attachToCaseReference value with attachToCaseDetails id, which should be ignored
+        CaseDetails attachToCaseDetails = ccdCaseCreator.createCase(emptyList(), Instant.now());
```
nitinprabhu, line 204: not sure comments and code are in sync. Where are we setting the `searchCaseReference` value

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
aliveni: yeah, it's not realistic. I'll update the test to send `searchCaseReference` value, `searchCaseReferenceType` as null and without `attachToCaseReference` field. Is that ok?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
aliveni: I'll update the test

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -312,6 +340,37 @@ private Response invokeCallbackEndpoint(
             .post("/callback/attach_case?ignore-warning=true");
     }
 
+    private Response invokeCallbackEndpointWithSearchCaseRefAndAttachToCaseRef(
+        CaseDetails searchCaseRefCaseDetails,
```
nitinprabhu, line 344: do you need to pass whole object here ? just case id is used so may be just pass that ?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -312,6 +340,37 @@ private Response invokeCallbackEndpoint(
...
+            .given()
+            .relaxedHTTPSValidation()
+            .baseUri(testUrl)
+            .header(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE)
+            .header(SyntheticHeaders.SYNTHETIC_TEST_SOURCE, "Bulk Scan Orchestrator Functional test")
+            .header(AUTHORIZATION, ccdAuthenticator.getUserToken())
+            .header(CcdCallbackController.USER_ID, ccdAuthenticator.getUserDetails().getId())
+            .body(callbackRequest)
+            .when()
+            .post("/callback/attach_case?ignore-warning=true");
```
nitinprabhu, line 371: warning can be false here ? No warnings list is populated so ignore warning = true would not be the right scenario

src/functionalTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/AttachExceptionRecordToExistingCaseTest.java
```diff
@@ -193,6 +193,34 @@ public void should_attach_exception_record_to_case_by_attach_to_case_reference()
         verifyExceptionRecordAttachesToCase(null);
     }
 
+    @Test
+    public void should_attach_exception_record_to_case_by_search_case_reference()
```
nitinprabhu, line 197: Also do we have a test case where both searchCaseReference and attachToCaseReference are set in call back ?

nitinprabhu: Realised this is not needed

PR #1018, opened 2020-06-26
Case update finalizer

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-1074

### Change description ###
Refactoring CcdCaseUpdater - extracted CaseUpdateFinalizer, updated CcdCaseUpdaterTest and added CaseUpdateFinalizerTest.

```
[ ] Yes
[x] No
```


lgonczar:
I disagree with this change. It adds sixth dependency, which submits an event to CCD, while the original class still starts the event directly. So now both talk to CCD at low level, the root class starting the event and the dependency completing it.

doncem:
> I disagree with this change.

i think team is not on a ball with this. ticket has been moved out and more brain melting investigation is due to come

apereverzin:
Sorry @doncem I should have close this PR

apereverzin:
Needs more thinking

PR #1020, opened 2020-07-01
Add FeignExceptionLogger utility class

doncem:

[Log the content of CCD error responses in non-production environments](https://tools.hmcts.net/jira/browse/BPS-697)

### Change description ###

Add debug logs in all relevant places where ccd api calls are made. Messages are up for discussion - just thought it should be super simple as we only want response body. The rest is pretty much logged thoroughly already

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/logging/FeignExceptionLogger.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/logging/FeignExceptionLogger.java
@@ -13,7 +13,7 @@ public final class FeignExceptionLogger {
         logger.debug(
             "{}. CCD response: {}",
             introMessage,
-            exception.responseBody().map(b -> new String(b.array())).orElse(exception.getMessage())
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/logging/FeignExceptionLogger.java
```diff
@@ -0,0 +1,19 @@
...
+import feign.FeignException;
+import org.slf4j.Logger;
+
+public final class FeignExceptionLogger {
+
+    private FeignExceptionLogger() {
+        // empty utility class construct
+    }
+
+    public static void debugCcdException(Logger logger, FeignException exception, String introMessage) {
```
apereverzin, line 12: Test class would make sense

doncem:
No don't need @mdayican, as soon as logging library is prepped by release drafter will make a 5.1.6 release and all it will mean is just extra env var in flux config - the rest will be automatically handled

PR #1024, opened 2020-07-03
Pact Testing: Transformation consumer

doncem:

[Implement consumer pact test for transformation in orchestrator](https://tools.hmcts.net/jira/browse/BPS-1281)

### Change description ###

Grunt work PR: test suite for transformation client contract

Next PR: enable in pipeline + publish

P.S. provider name together with hmcts/bulk-scan-processor#1397 will group well with bulk scan project

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/contractTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/transformation/SampleAppConsumerTest.java
+++ b/src/contractTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/transformation/SampleAppConsumerTest.java
@@ -26,7 +26,7 @@ public class SampleAppConsumerTest {
 
     private static final String TEST_S2S_TOKEN = "pact-test-s2s-token";
 
-    @Pact(provider = "sample_app_exception_transformation", consumer = "bulk_scan_orchestrator")
+    @Pact(provider = "bulk_scan_sample_app", consumer = "bulk_scan_orchestrator")
     public RequestResponsePact validTransformationPact(PactDslWithProvider builder) throws Exception {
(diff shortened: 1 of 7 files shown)


src/contractTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/transformation/SampleAppConsumerTest.java
```diff
@@ -0,0 +1,164 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.springframework.http.HttpStatus.OK;
+import static org.springframework.http.HttpStatus.UNPROCESSABLE_ENTITY;
+
+@ExtendWith(PactConsumerTestExt.class)
+public class SampleAppConsumerTest {
+
+    private static final String TEST_S2S_TOKEN = "pact-test-s2s-token";
+
+    @Pact(provider = "sample_app_ocr_validation", consumer = "bulk_scan_orchestrator")
```
mdayican, line 29: may be we need to change name of the provider.

doncem: dang. missed that ;). ta

mdayican: we can make it sth like sample_app and use same provider for all as sample app will not use any state, just we need to mock s2s and call the real system?

PR #1029, opened 2020-07-07
add log for ccd error to see whole

mdayican:
### Change description ###

currently we are getting 422 error and we can not see the whole stack trace.
add log for ccd error to see all.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdApi.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdApi.java
@@ -387,6 +387,7 @@ public class CcdApi {
             );
         } catch (FeignException e) {
             debugCcdException(log, e, "Failed to call 'submitEventForCaseWorker'");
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdApi.java
```diff
@@ -387,6 +387,7 @@ public CaseDetails submitEventForAttachScannedDocs(
             );
         } catch (FeignException e) {
             debugCcdException(log, e, "Failed to call 'submitEventForCaseWorker'");
+            log.error("CCD submitEventForCaseWorker for caseRef: {} ", caseRef, e);
```
apereverzin, line 390: This exception is rethrown, is it not logged by an error handler?

nitinprabhu: would this log expose sensitive information ?

PR #1030, opened 2020-07-07
Review CCD Api usage: AUTO event for SUPPLEMENTARY_EVIDENCE

doncem:
### JIRA link (if applicable) ###

[Orchestrator: refactor CcdCaseUpdater and its surroundings](https://tools.hmcts.net/jira/browse/BPS-1074)

### Change description ###

What has been done so far: review all possible scenarios orchestrator has involvement with ccd feign client which is key element in this service.

Goal: if possible, reduce calls, reduce split of those calls, identify what can be unified and reduced.

Most of those calls are fairly "simple":

- start event
- log some success
- build data to submit event
- submit event

There are some customised exception handling so cannot completely unify everything but solution may appear clearer once all such behavioural usage is reviewed/re-jigged.

Here is the first one, simple enough to apply: same actions were applied for both start and submit events. Just unifying them in one go making event class clear enough what it needs doing

Summary:

```markdown
BEFORE:

- ccdApi.authenticateJurisdiction(envelope.jurisdiction)
- ccdApi.startEventForAttachScannedDocs:
  - feignCcdApi.startEventForCaseWorker
  - exception:
    - 404 -> throw new UnableToAttachDocumentsException
    - otherwise: removeFromIdamCacheIfAuthProblem(e.status(), jurisdiction) + throw new CcdCallException
- ccdApi.submitEventForAttachScannedDocs:
  - feignCcdApi.submitEventForCaseWorker
  - exception:
    - 404 -> throw new UnableToAttachDocumentsException
    - otherwise: removeFromIdamCacheIfAuthProblem(e.status(), jurisdiction) + throw new CcdCallException
- catch (UnableToAttachDocumentsException e)? return false

AFTER:

- ccdApi.authenticateJurisdiction(envelope.jurisdiction)
- ccdApi.attachScannedDocs:
  - feignCcdApi.startEventForCaseWorker
  - feignCcdApi.submitEventForCaseWorker
  - exception:
    - 404 -> throw new UnableToAttachDocumentsException
    - otherwise: removeFromIdamCacheIfAuthProblem(e.status(), jurisdiction) + throw new CcdCallException
- catch (UnableToAttachDocumentsException e)? return false
```

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #1031, opened 2020-07-07
Unify ccdApi method calls for new exception

doncem:

### Change description ###

Second simple enough PR which follows up #1030. It is tightly dependent and cannot be merged separately - apparently attach docs tests were using exception record related methods and they were not referenced in former class :shrug: 

Marking as draft until dependent PR is merged.

Summary

```markdown
BEFORE:

- ccdApi.authenticateJurisdiction(envelope.jurisdiction)
- ccdApi.startEvent:
  - feignCcdApi.startForCaseworker
  - exception? removeFromIdamCacheIfAuthProblem(ex.status(), jurisdiction) + throw back
- ccdApi.submitEvent:
  - feignCcdApi.submitForCaseworker
  - exception? removeFromIdamCacheIfAuthProblem(ex.status(), jurisdiction) + throw back

AFTER:

- ccdApi.authenticateJurisdiction(envelope.jurisdiction)
- ccdApi.createExceptionRecord:
  - feignCcdApi.startForCaseworker
  - feignCcdApi.submitForCaseworker
  - exception? removeFromIdamCacheIfAuthProblem(ex.status(), jurisdiction) + throw back
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdApi.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdApi.java
@@ -16,6 +16,7 @@ import uk.gov.hmcts.reform.ccd.client.model.StartEventResponse;
 
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
 import javax.annotation.Nonnull;
 
 import static java.lang.String.format;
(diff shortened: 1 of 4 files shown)


doncem:
> as you said name can be generic like `startAndSubmitEvent`

once everything is unified in one place - it is easier to generalise. there are some bitty pieces which are specific for each scenario

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdApi.java
```diff
@@ -282,23 +282,40 @@ public CcdAuthenticator authenticateJurisdiction(String jurisdiction) {
...
+    // ideally should be more generic. atm just used in creating new exception record as auto event
+    public CaseDetails createExceptionRecord(
         CcdAuthenticator authenticator,
         String jurisdiction,
         String caseTypeId,
-        String eventTypeId
+        String eventTypeId,
+        Function<StartEventResponse, CaseDataContent> caseDataContentBuilder,
+        String logContext
     ) {
```
rkondratowicz, line 293: If it's `createExceptionRecord` then `eventTypeId` parameter should not be allowed.

So either rename the method or remove the param.

doncem: `eventTypeId` is also used in `caseDataContentBuilder`. not sure which way it would be better to get rid of it right now :thinking:

doncem: i suggest to leave it out as it is same thing for another auto action. generalising ccd api calls was the next step (as we discussed the other day). or if not next then just separate from all this refactor

PR #1033, opened 2020-07-08
Remove validation on attachToCaseReference field

aliveni:
https://tools.hmcts.net/jira/browse/BPS-1275

### Change description ###
When attaching the Exception Record to the case, the Callback request contains the target case reference value set to `searchCaseReference` field. So, `attachToCaseReference` field call back request validation is no longer required. 

We still validation to check if Exception Record is already attached (using `attachToCasereference` field).

- Removed validation on `attachToCaseReference` field. 
- Updated tests to set target case reference to `searchCaseReference` field 
- Removed no longer valid tests

```
[ ] Yes
[x] No
```


aliveni:
The PR was deployed in demo and tested for all services

aliveni:
Unable to test `supplementary_evidence_with_ocr` classification for probate due to the lack of test data. (in demo)

aliveni:
> This PR is blocked because we are unable to test `supplementary_evidence_with_ocr` classification for probate due to the lack of test data. (in demo)

Completed testing. Tested Probate case for the missing scenario

PR #1037, opened 2020-07-08
Replace direct usage of feign client to built-in ccd api

doncem:

### Change description ###

For new case creation callback the dependencies were farely split in the first place. But in order to unify all ccd interactions in one place it would be nice all elements do play their part. During this implementation this #1036 occurred. Raised as a helper PR. Now the leftovers

```
[ ] Yes
[x] No
```


apereverzin:
Absolutely agree with this PR. It would also be nice to get rid of direct use of CoreCaseDataApi in CcdCaseCreator, CcdCaseUpdator and CaseSearcher.

doncem:
> Absolutely agree with this PR. It would also be nice to get rid of direct use of CoreCaseDataApi in CcdCaseCreator, CcdCaseUpdator and CaseSearcher.

if i find inspiration during/after all this refactor - i'll do it. it was meant to happen but was always delayed

PR #1039, opened 2020-07-10
Remove case details awareness in chain of method calls

doncem:

### Change description ###

Through the day I tried to pick a place where to do some refactoring but everything is so much tangled up. Ended up in technological bottleneck.

At first I started shifting around related stuff (private class and more). But reverted the whole thing as got really confused. Moving out private class though serves as a bonus here.

The rest is looking around what can be re-used, reduced in jumping around for logical sequential tracking

Goal, which might not yet be completed, - get rid of nesting around case details as we already parse/validate it significantly as an initial member of method chains. This is mostly visible within the last 2 commits - move call to payments higher in the method chain as those looked like last things to execute without causing any side-affects. ~didn't move too far - would like to see how this change goes and then take another stab in all this~

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
@@ -239,7 +239,11 @@ public class AttachCaseCallbackService {
                 callBackEvent.targetCaseRef
             );
 
-            return attachToCase(callBackEvent, exceptionRecordDetails, ignoreWarnings);
+            return attachToCase(callBackEvent, ignoreWarnings)
+                .peek(map -> paymentsProcessor.updatePayments(
+                    exceptionRecordDetails,
+                    (String) map.get(ATTACH_TO_CASE_REFERENCE)
+                ));
         } catch (AlreadyAttachedToCaseException
             | DuplicateDocsException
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -592,13 +529,9 @@ public AttachToCaseEventData(
         return merged;
     }
 
-    private ServiceConfigItem getServiceConfig(CaseDetails caseDetails) {
-        return hasServiceNameInCaseTypeId(caseDetails).flatMap(service -> Try
-            .of(() -> serviceConfigProvider.getConfig(service))
-            .toValidation()
-            .mapError(Throwable::getMessage)
-        )
+    private ServiceConfigItem getServiceConfig(String service) {
```
aliveni, line 532: can we rename this method? this is also verifying the update URL

doncem: `getValidServiceConfig` sounds good enough?

aliveni: let's leave this change for future PRs because the case update related code could be moved from this class

PR #1045, opened 2020-07-14
Review legacy id prefilter

doncem:

### Change description ###

`attachCaseByCcdId` is executed in 2 places: normal process and legacy id fiddler. The latter only needs to verify about it's singleton existence and call same thing with same parameters. Moving that call one method up 

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/AttachCaseCallbackService.java
```diff
@@ -279,65 +280,62 @@ public AttachCaseCallbackService(
...
+                            targetCaseRef,
+                            exceptionRecordId
+                        );
+
+                        return Long.toString(targetCaseCcdId);
+                    }
+                )
+                .orElseThrow(() -> new CaseNotFoundException(
+                    String.format("No case found for legacy case reference %s", targetCaseRef)
+                ));
```
rkondratowicz, line 338: imho if-elseif-else was clearer

doncem: i'm cool. let's revert that then ;)

PR #1046, opened 2020-07-14
Review GetTargetCase usage in AttachToCase callback service

doncem:

### Change description ###

Moving case getter one method up. Will make future tracking easier and perhaps can be summarised by the looks into the switch statement. We only need to know about the target case further down the line

```
[ ] Yes
[x] No
```


doncem:
One of the options:

```java
Optional.of(callBackEvent.classification)
    .filter(this::isWithinAllowed) // list contains allowed classifications
    .map(voidClassification -> ccdApi.getCase(...))
    .map(update method call)
    .orElseThrow(() -> new CallbackException(...));
```

single call instead of wherever in the class: at 300 or 400 line :shrug: - it doesn't depend on classification. it just needs to be fetched for further actions to do attachments

PR #1035, opened 2020-07-08
Review CCD Api usage: case updater exception catches for submit

doncem:
### JIRA link (if applicable) ###

[Orchestrator: refactor CcdCaseUpdater and its surroundings](https://tools.hmcts.net/jira/browse/BPS-1074)

#### BEFORE:

2 sets of try/catch blocks related to ccd/feign client: start event and submit event. latter one has tiny behavioural amendment

#### AFTER:

1 set of try/catch block. squashed needed amendments into individual catch statements

#### Goal

Ability to combine ccd client calls into one. In some way like #1030 and #1031 

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #1048, opened 2020-07-15
Show error as error

rkondratowicz:
Currently an error in updating a case in CCD is presented to a user as a warning.
Changing to error.

(whether those details should be shown to user or not is another question...)

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
@@ -149,7 +149,7 @@ public class CcdCaseUpdater {
                 );
 
                 if (errorMsg.isPresent()) {
-                    return new ProcessResult(singletonList(errorMsg.get()), emptyList());
+                    return new ProcessResult(emptyList(), singletonList(errorMsg.get()));
                 } else {
                     var updatedExceptionRecordData = exceptionRecordFinalizer.finalizeExceptionRecord(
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -149,7 +149,7 @@ public ProcessResult updateCase(
                 );
 
                 if (errorMsg.isPresent()) {
-                    return new ProcessResult(singletonList(errorMsg.get()), emptyList());
+                    return new ProcessResult(emptyList(), singletonList(errorMsg.get()));
```
lgonczar, line 152: I think this whole bit is wrong - updateCaseInCcd should throw an exception instead of returning this optional. We're supposed to fail for a generic reason and not pass details of what went wrong in CCD to the caseworker.

nitinprabhu: I think you are right there is no point of showing a 422 error to caseworker and also this 422 won't be fixed on retrying so yes we should throw exception which would result into generic error but also we should log it so that we would be alerted.

doncem: for starters #1035 does throw exception on method call `updateCaseInCcd` :wink: . Of course, will need further review after that but if it can be unified/simplified on that bit - it should go on and do that first thing

PR #1050, opened 2020-07-15
Fix attach-to-case names

rkondratowicz:
Rename classes/variables from `attachCase` to `attachToCase` (as we're attaching exception recored to a case).

Those now match things like `AttachToCaseEventData` model and `attachToCase` method in callback controller.


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/CcdCallbackController.java
```diff
@@ -24,19 +24,19 @@
...
     public CcdCallbackController(
-        AttachCaseCallbackService attachCaseCallbackService,
+        AttachToCaseCallbackService attachToCaseCallbackService,
         CreateCaseCallbackService createCaseCallbackService,
         ReclassifyCallbackService reclassifyCallbackService
     ) {
-        this.attachCaseCallbackService = attachCaseCallbackService;
+        this.attachToCaseCallbackService = attachToCaseCallbackService;
         this.createCaseCallbackService = createCaseCallbackService;
         this.reclassifyCallbackService = reclassifyCallbackService;
```
doncem, line 41: it's all ruined now :(

PR #1063, opened 2020-07-21
Handling UnprocessableEntity as server error

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-1329

### Change description ###
Feign.UnprocessableEntity is handled as Server error.

It appears that Conflict is already handled as defined in the task.

Also reduced number of arguments in `updateCaseInCcd` method to pass the quality gateway.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
@@ -208,6 +208,12 @@ public class CcdCaseUpdater {
                 "Invalid case-update response. " + errorMessage,
                 exc
             );
+        } catch (CcdCallException exception) {
+            log.error(exception.getMessage(), exception.getCause());
+            throw new CallbackException(
+                getErrorMessage(configItem.getService(), existingCaseId, exceptionRecord.id),
+                exception
+            );
         } catch (Exception exception) {
             throw new CallbackException(
                 getErrorMessage(configItem.getService(), existingCaseId, exceptionRecord.id),
@@ -244,8 +250,6 @@ public class CcdCaseUpdater {
 
     /**
      * Submits event to update the case.
-     *
-     * @return either error message in case of error or empty if no error detected
      */
     private void updateCaseInCcd(
         String service,
@@ -291,8 +295,7 @@ public class CcdCaseUpdater {
                 exceptionRecord.id,
                 exception.contentUTF8()
             );
-            log.error(msg, exception);
-            throw new RuntimeException(msg);
+            throw new CcdCallException(msg, exception);
         } catch (FeignException.Conflict exception) {
             throw exception;
         } catch (FeignException exception) {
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -295,8 +295,7 @@ private String getErrorMessage(String service, String existingCaseId, String exc
                 exception.contentUTF8()
             );
             log.error(msg, exception);
-
-            return Optional.of(msg);
+            throw new CcdCallException(msg, exception);
```
doncem, line 298: this is the last one with branch **not** returning a value so single return is from success path which is `optional.empty`. we can get rid of return type and make calling method simpler

lgonczar: The same story says that in case of conflicts errors need to be returned to the caseworker. This means that we need a return value, just not in this place.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -328,4 +326,15 @@ private CaseDataContent buildCaseDataContent(
...
+        final String idamToken;
+        final String s2sToken;
+        final String userId;
+
+        Credentials(String idamToken, String s2sToken, String userId) {
+            this.idamToken = idamToken;
+            this.s2sToken = s2sToken;
+            this.userId = userId;
+        }
+    }
```
doncem, line 339: erm. that falls into another ticket - BPS-1074: refactor CcdCaseUpdater and its surroundings

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
apereverzin: @lgonczar As mentioned in the description for the PR
It appears that FeignException.Conflict is already handled as defined in the task.
It is retrown in `updateCaseInCcd` method and in `updateCase` method it is caught and converted in eror response.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
apereverzin: As I explained in the PR channel I have had to introduce a wrapper class to reduce the number of method parameters (to pass the quality gateway), just using `startEvent.getCaseDetails().getId()` instead of `existingCaseId` caused a failure in functional test

doncem: can use suppression flag for that with the note to be addressed in future prs as it is pending/planned.

the other note is just reverting back the change introduced earlier in pr. which is good

lgonczar: This is a change in the right direction, but I think that when we introduce it, we should wrap credentials in a class as early as possible, i.e. before even calling this class. Worth creating a separate request for that (perhaps as the first one to be merged).

PR #1065, opened 2020-07-22
Extend transformation request by fields required for automated case creation

lgonczar:
**Can't be merged before being verified in demo.**

### Change description ###

Extend the request to transformation endpoint by adding fields required for automated case creation (https://tools.hmcts.net/confluence/pages/viewpage.action?pageId=1425080530#Automatedcasecreationandupdate-transformationEndpointChangesTransformationendpointchanges)
 
```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/callback/ExceptionRecordValidator.java
```diff
@@ -136,6 +142,18 @@ public ExceptionRecordValidator() {
         ).toValidation().mapError(throwable -> "Invalid scannedDocuments format. Error: " + throwable.getMessage());
     }
 
+    private String getEnvelopeId(CaseDetails caseDetails) {
+        return Optional.ofNullable(caseDetails)
+            .map(CaseDetails::getData)
+            .map(data -> (String) data.get(ENVELOPE_ID))
+            .orElseThrow(
+                // we can't apply standard validation, as the caseworker doesn't even know about envelope ID field
+                () -> new UnprocessableCaseDataException(
```
nitinprabhu, line 151: do we have a test to cover this ?

src/integrationTest/resources/ccd/callback/create-case/request/valid-exception-without-form-type.json
```diff
@@ -13,6 +13,7 @@
       "journeyClassification": "EXCEPTION",
       "deliveryDate": "2018-01-02T12:34:56.123",
       "openingDate": "2018-01-03T12:34:56.123Z",
+      "envelopeId": "e0c7b48e-624a-4b1f-ba3e-7a0d50d09e80",
```
nitinprabhu, line 16: `is_automated_process` and `exception_record_case_type_id` are not included in any of the integration test json files should be add them ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/callback/ExceptionRecordValidator.java
lgonczar: We do, we have tests for verifying that the callback endpoint will return a 422 response with the right body when envelope ID is missing. Earlier today I also created a ticket for `ExceptionRecordValidator` unit tests (https://tools.hmcts.net/jira/browse/BPS-1336), as they're all missing.

src/integrationTest/resources/ccd/callback/create-case/request/valid-exception-without-form-type.json
lgonczar: These envelopeIds that I've included are just added to callback data, as they're required there. When it comes to calling transformation endpoint, we have no test verifying what data we've sent to it. In https://tools.hmcts.net/jira/browse/BPS-1336 I've mentioned the need for tests to cover the mapping of callback data to the model representing transformation request. We can also think of integration tests for that (using wiremock).

PR #1051, opened 2020-07-15
Finalize exception record after its OCR is used to update a case

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-1326

PR #1070, opened 2020-07-23
Use more appropriate return model

rkondratowicz:
The `exceptionRecordData` property of the previous model (`ProcessResult`) was set but never used.
Now that it is no longer being set (fixed in https://github.com/hmcts/bulk-scan-orchestrator/pull/1051) the type can be changed.


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -151,33 +148,33 @@ public ProcessResult updateCase(
...
             }
         } catch (UnprocessableEntity exception) {
             ClientServiceErrorResponse errorResponse = serviceResponseParser.parseResponseBody(exception);
-            return new ProcessResult(errorResponse.warnings, errorResponse.errors);
+            return Optional.of(ErrorsAndWarnings.withErrorsAndWarnings(errorResponse.errors, errorResponse.warnings));
         } catch (FeignException.Conflict exception) {
             String msg = getErrorMessage(configItem.getService(), existingCaseId, exceptionRecord.id)
                 + " because it has been updated in the meantime";
             log.error(msg);
             ClientServiceErrorResponse errorResponse = new ClientServiceErrorResponse(singletonList(msg), emptyList());
```
rkondratowicz, line 160: I'll remove those unnecessary objects on a separate PR

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdCaseUpdater.java
```diff
@@ -131,7 +128,7 @@ public ProcessResult updateCase(
                     existingCase.getId(),
                     exceptionRecord.id
                 );
-                return new ProcessResult(updateResponse.warnings, emptyList());
+                return Optional.of(ErrorsAndWarnings.withErrorsAndWarnings(emptyList(), updateResponse.warnings));
```
nitinprabhu, line 131: `ErrorsAndWarnings.withErrorsAndWarnings` can we do static imports here and below ?

PR #1071, opened 2020-07-24
Add configuration properties for automatic envelope processing

lgonczar:

### Change description ###

Add configuration properties required for automatic envelope processing - case creation, case update and case search.

```
[ ] Yes
[x] No
```


src/main/resources/application.yaml
```diff
@@ -91,6 +94,9 @@ service-config:
         - formType: PERSONAL
           ocrFields:
             - last_name
+      auto-case-creation-enabled: ${AUTO_CASE_CREATION_ENABLED_BULKSCANAUTO:false}
+      auto-case-update-enabled: ${AUTO_CASE_UPDATE_ENABLED_BULKSCANAUTO:false}
+      case-definition-has-envelope-ids: ${CASE_DEFINITION_HAS_ENVELOPE_IDS_BULKSCANAUTO:false}
```
nitinprabhu, line 99: normally first PR can be `AUTO_CASE_CREATION_ENABLED_BULKSCANAUTO` `AUTO_CASE_UPDATE_ENABLED_BULKSCANAUTO` and `CASE_DEFINITION_HAS_ENVELOPE_IDS_BULKSCANAUTO` can go in values yaml with false value and then override wherever needed. 
This way you get the flexibility of changing it per environment and also your chart config goes before your code usage.

lgonczar: I'm not sure I understand what you mean.

lgonczar: Done

nitinprabhu: thanks @lgonczar

PR #1072, opened 2020-07-24
Update transformation and update case URLs

aliveni:
https://tools.hmcts.net/jira/browse/BPS-1333

### Change description ###
Updated transformation URLs and update-case URLs to have complete URLs.
~~The URLs will be temporarily overridden in flux config so that the build will not break.
After the code changes, Flux config changes will be overridden.~~

Fluc config PR - https://github.com/hmcts/cnp-flux-config/pull/4956

Update: Included the code changes to use the complete URLs in transformation and update endpoints.

```
[ ] Yes
[x] No
```


aliveni:
> Why not also change transformation client at the same time to not use `/transform-exception-record`
> This way you won't have any flux changes to be done

I thought the config changes should go first. I'll update the PR to include code changes

doncem:
config changes alone should be fine and safe(r). the way url is parsed it sets path as well if present and code would just override. once pods are updated with latest config - the source code `.path("transformation")` would become redundant. up for you to decide but i would vote for config only providing assurance that urls are parsed correctly

PR #1079, opened 2020-07-27
Extend transformation request model

lgonczar:

### Change description ###

This is almost a copy of https://github.com/hmcts/bulk-scan-orchestrator/pull/1065, but with `JsonIgnore` annotation added on the new fields.

Extend transformation request model by the fields required by the automated case creation functionality. These fields are in the models, but not in requests. They will be included when it's confirmed in demo that they're not breaking existing transformation.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/AttachExceptionRecordWithOcrTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/AttachExceptionRecordWithOcrTest.java
@@ -11,6 +11,7 @@ import org.assertj.core.util.Maps;
 import org.json.JSONObject;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
+import org.skyscreamer.jsonassert.JSONAssert;
 import org.springframework.boot.web.server.LocalServerPort;
 import org.springframework.cloud.contract.wiremock.AutoConfigureWireMock;
 import org.springframework.http.HttpHeaders;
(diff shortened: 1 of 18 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/bulkscan/orchestrator/controllers/AttachExceptionRecordWithOcrTest.java
```diff
@@ -241,6 +259,7 @@ void should_fail_correctly_if_ccd_is_down() throws Exception {
         caseData.put(JOURNEY_CLASSIFICATION, SUPPLEMENTARY_EVIDENCE_WITH_OCR);
         caseData.put("formType", "B123");
         caseData.put(SEARCH_CASE_REFERENCE_FIELD_NAME, CASE_ID);
+        caseData.put(ENVELOPE_ID_FIELD_NAME, "2777a98c-cbb4-11ea-87d0-0242ac130003");
```
doncem, line 262: would be good to have a comment where this value is coming from or smth similar for future linking

doncem: by the looks after scrolling all this there are different envelope ids here and this is more generic method: `getCaseData`

lgonczar: This change only extends the existing code by the required field which was absent in the request so far.

The method looks generic, but it isn't - it returns case data expected in the response to callback request from `valid-supplementary-evidence-with-ocr.json`. Method and several variables need renaming, comments need to be added, but that's outside the scope of this PR. I'll create another one for just that.

PR #1083, opened 2020-07-28
Add search case service config property

rkondratowicz:
it will be used to determine whether existing cases should be searched by envelope id or not.
https://tools.hmcts.net/jira/browse/BPS-1339

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -80,6 +80,7 @@ service-config:
       auto-case-creation-enabled: ${AUTO_CASE_CREATION_ENABLED_BULKSCAN}
       auto-case-update-enabled: ${AUTO_CASE_UPDATE_ENABLED_BULKSCAN}
       case-definition-has-envelope-ids: ${CASE_DEFINITION_HAS_ENVELOPE_IDS_BULKSCAN}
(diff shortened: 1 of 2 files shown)


src/main/resources/application.yaml
```diff
@@ -80,6 +80,7 @@ service-config:
       auto-case-creation-enabled: ${AUTO_CASE_CREATION_ENABLED_BULKSCAN}
       auto-case-update-enabled: ${AUTO_CASE_UPDATE_ENABLED_BULKSCAN}
       case-definition-has-envelope-ids: ${CASE_DEFINITION_HAS_ENVELOPE_IDS_BULKSCAN}
+      search-cases-by-envelope-id: false
```
lgonczar, line 83: Maybe it would be worth already relying on an environment variable, so that when we need to turn this on, we just change flux config?

rkondratowicz: I'd prefer to add an env var once it's actually needed

lgonczar: I expect we'll turn it on multiple times - service by service, demo first. Each time we'll need to update chart. Eventually, we'll turn it on everywhere, so it will be less work to make those env changes in one go. And consistent with other related config properties.

PR #1091, opened 2020-07-29
Use CaseFinder in CreateCaseCallbackService

rkondratowicz:
follow up to adding a new method to `CaseFinder` that searches by either the old or new field based on service config (https://github.com/hmcts/bulk-scan-orchestrator/pull/1089)

https://tools.hmcts.net/jira/browse/BPS-1339


src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CreateCaseCallbackServiceTest.java
```diff
@@ -307,7 +309,7 @@ void should_return_existing_case_if_it_exists_in_ccd_for_a_given_exception_recor
         // given
         setUpServiceConfig();
 
-        when(ccdApi.getCaseRefsByBulkScanCaseReference(CASE_ID, SERVICE))
+        when(caseFinder.findCases(any(), any()))
```
rkondratowicz, line 312: unfortunately, this _unit_ tests class injects a real `ExceptionRecordValidator` instead of a mock and I'm unable to use a mocked exception record.

The `ExceptionRecordValidator` class itself has no unit tests either.
There's a story in the backlog to add them.

PR #1101, opened 2020-08-03
Remove CaseData interface

lgonczar:
### Change description ###

Remove CaseData interface, as it doesn't serve any purpose.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/AttachDocsToSupplementaryEvidence.java
```diff
@@ -75,7 +75,7 @@ public boolean attach(Envelope envelope, CaseDetails existingCase) {
     }
 
     private CaseDataContent buildCaseDataContent(Envelope envelope, StartEventResponse startEventResponse) {
-        CaseData caseData = mapper.map(
+        SupplementaryEvidence caseData = mapper.map(
```
rkondratowicz, line 78: how about we rename the var as well?

PR #1103, opened 2020-08-04
Remove `@JsonIgnore` annotatations

doncem:

[Update case update requests in orchestrator to conform to the new format](https://tools.hmcts.net/jira/browse/BPS-1337)

### Change description ###

Models have been introduced and revised a bit to fit the needs. Enabling PR build to test out on demo (hence the prefix)

```
[ ] Yes
[x] No
```


doncem:
Verified on demo - case was updated. Hence client call to update case with extra request fields are OK. Worries resolved

PR #1106, opened 2020-08-04
Create a class for appending envelope references to case data

lgonczar:

### Change description ###

Create a class for appending envelope references to case data. In the next PRs this class will be used for appending envelope reference when:
- creating a case from exception record (and later from the envelope, as well)
- attaching supplementary evidence with OCR from exception record (and later from the envelope, as well)
- attaching supplementary evidence from envelope

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/EnvelopeReference.java
```diff
@@ -0,0 +1,23 @@
...
+
+    @JsonProperty
+    public String id;
+
+    @JsonProperty
+    public String action;
+
+    public EnvelopeReference(
+        @JsonProperty("id") String id,
+        @JsonProperty("action") String action
```
doncem, line 18: are those required? should we even start marking as required since we rarely used that feature before?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EnvelopeReferenceAppender.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+@Service
+public class EnvelopeReferenceAppender {
+
+    private final ObjectMapper objectMapper;
+    private final Set<String> enabledServices;
+
+    public EnvelopeReferenceAppender(
+        ObjectMapper objectMapper,
+        ServiceConfiguration serviceConfiguration
```
doncem, line 31: `@Service` annotation should have `@EnableConfigurationProperties`, no matter if it was already present somewhere but never trust a lazy autoload of the spring boot :wink:

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EnvelopeReferenceAppender.java
```diff
@@ -0,0 +1,91 @@
...
+     * <p>This collection represents bulkScanEnvelopes field in service case and the update will only take place
+     * if the service supports this field (i.e. has it in all its CCD case definitions).</p>
+     *
+     * @return Updated collection, if envelope info was appended. Otherwise, if the CCD case definition
+     *         for the service doesn't support envelope reference collection (bulkScanEnvelopes field),
+     *         empty Optional is returned.
+     */
+    public Optional<List<CcdCollectionElement<EnvelopeReference>>> tryAppendEnvelopeReference(
+        String service,
+        List<Map<String, Object>> existingEnvelopeReferences,
```
doncem, line 55: maybe worth including all params as well, specifically explaining what is this list of maps and/or where from

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EnvelopeReferenceAppender.java
```diff
@@ -0,0 +1,91 @@
...
+        String envelopeId,
+        CaseAction action
+    ) {
+        if (enabledServices.contains(service)) {
+            return Optional.of(
+                appendReference(existingEnvelopeReferences, envelopeId, action)
+            );
+        } else {
+            return Optional.empty();
+        }
```
doncem, line 65: perhaps it triggers too often but if/else + optional just.. triggers.

how about approach like this?:

```java
enabledServices
            .stream()
            .filter()
            .findFirst()
            .map();
```

not against any approach - i guess my functional mind does not accept if/else + optional

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EnvelopeReferenceAppender.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    private List<CcdCollectionElement<EnvelopeReference>> appendReference(
+        List<Map<String, Object>> existingEnvelopeReferences,
+        String envelopeId,
+        CaseAction action
+    ) {
+        if (existingEnvelopeReferences == null || existingEnvelopeReferences.isEmpty()) {
```
doncem, line 73: ideally should never ever be null - let's make it as a rule - avoid null on lists whenever possible

doncem: and in general, this statement can be checked by some library (even spring boot) - `CollectionUtils` or similar

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EnvelopeReferenceAppender.java
```diff
@@ -0,0 +1,91 @@
...
+                .map(ref -> objectMapper.convertValue(ref.get("value"), EnvelopeReference.class))
+                .collect(toList());
+
+            updatedReferences.add(new EnvelopeReference(envelopeId, action.value));
+
+            return updatedReferences
+                .stream()
+                .map(ref -> new CcdCollectionElement<>(ref))
+                .collect(toList());
+        }
```
doncem, line 89: looks like an overuse of stream api. how about:

```java
var updatedReferences = existingEnvelopeReferences
    .stream()
    .map(ref -> objectMapper...)
    .map(CcdCollectionElement<>::new)
    .collect(toList());

updatedReferences.add(new CcdCollectionElement<>(new EnvelopeReference(envelopeId, action.value)));

return updatedReferences;
```

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/definition/CommonCaseFields.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.definition;
+
+/**
+ * Names of the fields used by Bulk Scan in service CCD case definitions.
+ */
+public class CommonCaseFields {
```
doncem, line 6: so is it just common case fields (which we have somewhere) or common fields specific for bulk scan?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/definition/CommonCaseFields.java
```diff
@@ -0,0 +1,17 @@
...
+
+    private CommonCaseFields() {
+        // hiding the constructor - this class is for constants only
```
doncem, line 9: just utility class construct :wink:. and class is `final` then

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EnvelopeReferenceAppender.java
lgonczar: Good point, now I realise I should have used ServiceConfigProvider to retrieve those.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EnvelopeReferenceAppender.java
lgonczar: Will do

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EnvelopeReferenceAppender.java
lgonczar: I briefly considered this, but decided against it for the sake of readability.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EnvelopeReferenceAppender.java
lgonczar: Those references come with case data where they can be null, so best to cover both.

When writing this, I quickly searched for some kind of isNullOrEmpty thing for collection within the available libraries, but couldn't find any. Happy to use it if anyone knows one.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EnvelopeReferenceAppender.java
lgonczar: My intention was to convert to CCD-compatible version at the very end, after the job is done. But as the job is tiny, both are fine. I'll make the change.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/definition/CommonCaseFields.java
lgonczar: The fields that we have somewhere are exception record fields - we know them all. In service cases we're only aware of those fields that we need to set or read - those are common to all case definitions.

I'll rename to `CommonServiceCaseFields` if you're happy.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/definition/CommonCaseFields.java
lgonczar: Will change

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/EnvelopeReference.java
lgonczar: Both are required, actually. I'll add that.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/definition/CommonCaseFields.java
lgonczar: Actually, this is one class to many for this PR - I'll remove it.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/EnvelopeReference.java
lgonczar: done

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EnvelopeReferenceAppender.java
lgonczar: done - replaced with ServiceConfigProvider

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EnvelopeReferenceAppender.java
lgonczar: done

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EnvelopeReferenceAppender.java
lgonczar: used CollectionUtils

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EnvelopeReferenceAppender.java
lgonczar: done

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/definition/CommonCaseFields.java
lgonczar: done

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/definition/CommonCaseFields.java
lgonczar: done - class has been removed

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/definition/CommonCaseFields.java
lgonczar: done - removed the class

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/model/ccd/EnvelopeReference.java
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Represents BulkScanEnvelope complex type in CCD case definitions.
+ */
+public class EnvelopeReference {
+
+    @JsonProperty
+    public String id;
+
+    @JsonProperty
+    public String action;
```
rkondratowicz, line 14: perhaps we could either have constant strings in `CaseAction` class above or use the enum here?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EnvelopeReferenceAppender.java
```diff
@@ -0,0 +1,86 @@
...
+     *
+     * <p>This collection represents bulkScanEnvelopes field in service case and the update will only take place
+     * if the service supports this field (i.e. has it in all its CCD case definitions).</p>
+     *
+     * @param service Name of the service that owns the case
+     * @param existingEnvelopeReferences Current value of the field that holds envelope references in case,
+     *                                   in raw format (after deserialising case data as Object)
+     * @param envelopeId Id of the envelope to be appended
+     * @param action Action that the envelope caused on the case - create/update
+     * @return Updated collection, if envelope info was appended. Otherwise, if the CCD case definition
```
rkondratowicz, line 46: Really weird that this method could take an empty list and return an empty optional.
If nothing got appended I'd expect the original list to be returned.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EnvelopeReferenceAppender.java
```diff
@@ -0,0 +1,86 @@
...
+    public Optional<List<CcdCollectionElement<EnvelopeReference>>> tryAppendEnvelopeReference(
+        String service,
+        List<Map<String, Object>> existingEnvelopeReferences,
+        String envelopeId,
+        CaseAction action
+    ) {
+        if (serviceConfigProvider.getConfig(service).getCaseDefinitionHasEnvelopeIds()) {
+            return Optional.of(
+                appendReference(existingEnvelopeReferences, envelopeId, action)
+            );
```
rkondratowicz, line 59: Wouldn't it be clearer and more straightforward if this check was performed by the class that wants to append something to envelope references?

Imho this logic does not belong here.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EnvelopeReferenceCollectionHelper.java
```diff
@@ -0,0 +1,74 @@
...
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.config.ServiceConfigProvider;
+
+import java.util.List;
+import java.util.Map;
+
+import static java.util.Arrays.asList;
+import static java.util.stream.Collectors.toList;
+
+@Service
+public class EnvelopeReferenceCollectionHelper {
```
lgonczar, line 18: Any ideas for a better name?

nitinprabhu: can we just say `EnvelopeHelper` ?

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/EnvelopeReferenceCollectionHelper.java
```diff
@@ -0,0 +1,74 @@
...
+     *
+     * <p>This collection represents bulkScanEnvelopes field in service case</p>
+     *
+     * @param existingEnvelopeReferences Current value of the field that holds envelope references in case,
+     *                                   in raw format (after deserialising case data as Object)
+     * @param envelopeId                 Id of the envelope to be appended
+     * @param action                     Action that the envelope caused on the case - create/update
+     * @return Updated collection, serialisable to CCD-compatible format
+     */
+    public List<CcdCollectionElement<EnvelopeReference>> appendEnvelopeReference(
```
nitinprabhu, line 46: existingEnvelopeReferences and updatedReferences can they not be of same type ?
Also can we define a wrapper class say `EnvelopeReference` and which can consist of both fields existingReference and updatedReference ?

nitinprabhu: just a thought let me know what you think not that too strong about it.

lgonczar: Maybe I should reduce this method just to converting the collection and the appending would happen in the calling code?

lgonczar: Done

PR #1110, opened 2020-08-05
Fix scanned documents format after attaching exception record reference

aliveni:
### Change description ###
When attaching `exceptionRecordReference` to the scanned documents in `SuccessfulUpdateResponse`, the format of the scanned document is not same as CCD documents format.
- Fix the scanned documents format in the case data
~~Add tests for the `getScannedDocuments()` method in `ScannedDocumentsHelper` class~~
- Update functional tests

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/helper/ScannedDocumentsHelper.java
```diff
@@ -73,25 +74,25 @@ public static void setExceptionRecordIdToScannedDocuments(
...
+                    return ImmutableMap.of("value", scannedDocument);
                 }
             })
             .collect(toList());
         caseData.put(SCANNED_DOCUMENTS, updatedScannedDocuments);
     }
 
     @SuppressWarnings("unchecked")
-    private static List<ScannedDocument> getScannedDocuments(Map<String, Object> caseData) {
+    public static List<ScannedDocument> getScannedDocuments(Map<String, Object> caseData) {
```
apereverzin, line 88: It looks like this method has been made public just for the sake of calling it from tests. If this is the case - this should not be done.

aliveni:
> It looks like `getScannedDocuments` method has been made public just for the sake of calling it from tests. If this is the case - this should not be done.

@apereverzin reverted the change to the `getScannedDocuments` method

PR #1117, opened 2020-08-11
Extend TransformationRequestCreator to create requests based on envelopes

lgonczar:

### Change description ###

Extend TransformationRequestCreator to create requests based on envelopes. So far transformation requests were only created based on ExceptionRecords, but this will change in automated creation mode.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/TransformationRequestCreator.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/TransformationRequestCreator.java
@@ -1,11 +1,38 @@
 package uk.gov.hmcts.reform.bulkscan.orchestrator.client;
 
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.client.model.request.DocumentType;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.client.model.request.DocumentUrl;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.client.model.request.OcrDataField;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.client.model.request.ScannedDocument;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/TransformationRequestCreator.java
```diff
@@ -23,4 +50,75 @@ public TransformationRequest create(ExceptionRecord exceptionRecord) {
             exceptionRecord.ocrDataFields
         );
     }
+
+    public TransformationRequest create(Envelope envelope) {
+        return new TransformationRequest(
+            null,
+            null,
```
lgonczar, line 57: exception record ID and exception record case type ID

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/TransformationRequestCreator.java
```diff
@@ -1,12 +1,39 @@
...
+
+import static java.util.Collections.emptyList;
+import static java.util.stream.Collectors.toList;
 
 @Service
 public class TransformationRequestCreator {
 
+    private final String documentManagementUrl;
+    private final String documentManagementContextPath;
+
```
nitinprabhu, line 28: 🤔 why document url and path are part of TransformationRequestCreator ?

lgonczar: Because it needs to build document URLs and for that it needs these two pieces of data.

nitinprabhu: Is it not processor has already uploaded documents to document management so whole url should be available to orchestrator ? why does orchestrator needs to create it again ?

lgonczar: I think initially the processor used to pass the full URL to the orchestrator, but we were requested to change it and now the processor passes document IDs to the orchestrator instead.

PR #1122, opened 2020-08-12
Add an override of CcdApi's createCase which doesn't accept authentication details

lgonczar:

### Change description ###

Add an override of CcdApi's createCase which doesn't accept authentication details. Instead, it retrieves those details itself, as many other CcdApi's methods do. This override will be used for automatic case creation.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdApi.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdApi.java
@@ -303,6 +303,27 @@ public class CcdApi {
     }
 
     public long createCase(
+        String jurisdiction,
+        String caseTypeId,
+        String eventId,
+        Function<StartEventResponse, CaseDataContent> caseDataContentBuilder,
+        String logContext
+    ) {
+        CcdAuthenticator ccdAuthenticator = authenticatorFactory.createForJurisdiction(jurisdiction);
+
+        return createCase(
+            ccdAuthenticator.getUserToken(),
+            ccdAuthenticator.getServiceToken(),
+            ccdAuthenticator.getUserDetails().getId(),
+            jurisdiction,
+            caseTypeId,
+            eventId,
+            caseDataContentBuilder,
+            logContext
+        );
+    }
+
+    public long createCase(
         String idamToken,
         String s2sToken,
         String userId,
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdApi.java
```diff
@@ -302,6 +302,27 @@ public void attachScannedDocs(
...
     }
 
+    public long createCase(
+        String jurisdiction,
+        String caseTypeId,
+        String eventId,
+        Function<StartEventResponse, CaseDataContent> caseDataContentBuilder,
+        String logContext
+    ) {
+        CcdAuthenticator ccdAuthenticator = authenticatorFactory.createForJurisdiction(jurisdiction);
```
nitinprabhu, line 312: nice for auto case create it will be always backend user

src/test/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/CcdApiCreateCaseTest.java
```diff
@@ -0,0 +1,166 @@
...
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyBoolean;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.BDDMockito.willThrow;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+
+@ExtendWith(MockitoExtension.class)
+public class CcdApiCreateCaseTest {
```
nitinprabhu, line 31: may be worth adding a test case when authentication fails ?

lgonczar: true, will do

lgonczar: I've added the test covering the situation when authentication details retrieval fails.

As for 401/403 errors that could come from CCD, the client doesn't distinguish between them and any other error responses - it just rethrows the exception from the client. This is covered by tests.

PR #1125, opened 2020-08-12
Split TransformationClient into two classes

lgonczar:

### Change description ###

Split `TransformationClient` into:
* higher-level `CaseDataTransformer`, responsible for transforming exception records (and envelopes soon)
* lower-level `TransformationClient`, focused on calling transformation API and parsing the response. 

The reason behind this split is the addition (in the next PR) of calling transformation with envelope data. If the new method was added to `TransformationClient` without this split of responsibilities, that would result in adding a lot of tests (this is what it would look like: https://github.com/hmcts/bulk-scan-orchestrator/pull/1124/files), which can be avoided.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/transformation/CaseDataTransformer.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/transformation/CaseDataTransformer.java
@@ -1,0 +1,33 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.client.transformation;
+
+import org.springframework.stereotype.Component;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.client.TransformationRequestCreator;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/client/transformation/CaseDataTransformer.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public CaseDataTransformer(
+        TransformationRequestCreator requestCreator,
+        TransformationClient transformationClient
+    ) {
+        this.requestCreator = requestCreator;
+        this.transformationClient = transformationClient;
+    }
+
+    public SuccessfulTransformationResponse transformExceptionRecord(
```
lgonczar, line 22: There will be another method in this class: `transformEnvelope`.

PR #1139, opened 2020-08-14
Make EnvelopeHandler create cases by default for NEW_APPLICATION envelopes

lgonczar:

### Change description ###

Make EnvelopeHandler create cases by default for NEW_APPLICATION envelopes. This change turns on the automatic case creation functionality. The PR won't be merged before the corresponding tests are merged.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EnvelopeHandler.java
```diff
@@ -60,15 +68,45 @@ public EnvelopeProcessingResult handleEnvelope(Envelope envelope) {
...
+            case CASE_ALREADY_EXISTS:
+                paymentsProcessor.createPayments(envelope, caseCreationResult.caseCcdId, false);
+                return new EnvelopeProcessingResult(caseCreationResult.caseCcdId, CASE_CREATED);
+            case ABORTED_WITHOUT_FAILURE:
+            case UNRECOVERABLE_FAILURE:
+                return new EnvelopeProcessingResult(createExceptionRecord(envelope), EXCEPTION_RECORD);
+            case POTENTIALLY_RECOVERABLE_FAILURE:
+                if (deliveryCount < MAX_RETRIES_FOR_POTENTIALLY_RECOVERABLE_FAILURES) {
+                    // let the application retry later
+                    throw new CaseCreationException("Case creation failed due to a potentially recoverable error");
```
lgonczar, line 95: Currently, the only way of preventing `MessageProcessor` from completing the queue message is by throwing an exception.

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/EnvelopeHandler.java
```diff
@@ -60,15 +68,45 @@ public EnvelopeProcessingResult handleEnvelope(Envelope envelope) {
...
+                } else {
+                    // too many attempts - fall back to exception record
+                    return new EnvelopeProcessingResult(createExceptionRecord(envelope), EXCEPTION_RECORD);
+                }
+            default:
+                throw new CaseCreationException(
```
nitinprabhu, line 101: should the default case be not exception record otherwise we would keep retrying till deliver count is exhausted and then fallback to exception record ?

lgonczar: Not sure if it matters. This scenario will only happen when a developer adds a new result type for case creation and doesn't include the result type in this switch. The outcome will be an exception record, unless the problem is fixed in the meantime.

I'm happy with either way.

nitinprabhu: ok yeah end result will be exception record but I was thinking it will unnecessarily retry and we will see exceptions/alerts  .
But yes this can only happen if dev forgets to add the result type.

lgonczar:
Closing - build won't pass

PR #1141, opened 2020-08-14
Remove servicebus connection strings from fun tests

doncem:

[Improve the way Service Bus clients are created](https://tools.hmcts.net/jira/browse/BPS-771)

### Change description ###

Noticed payments queue is not used but i'm not removing as part of this ticket - just replace conn strings.

Flux PR shortly coming - it'll need new env vars. Will not remove old to not break anything

```
[ ] Yes
[x] No
```


doncem:
you think correctly @nitinprabhu :slightly_smiling_face: . hence already merged a non-breaking pr there

PR #1158, opened 2020-08-19
Removed core infra compute references as platform cleaned up core-infra

nitinprabhu:


mdayican (approved):

doncem (approved):
at last some terraform cleanu happening. should we delete whole state file?

nitinprabhu:
> at last some terraform cleanu happening. should we delete whole state file?

no we can't needs the bit which is azurerm

lgonczar (approved):

PR #1201, opened 2020-09-08
Use new /userinfo instead of old /user_details in idam

rkondratowicz:


doncem (approved):
as long as all integrated services are fine with it - go for it! :shipit:

PR #1207, opened 2020-09-09
Extract new application handler

rkondratowicz:
Extract steps for creating a case from general `EnvelopeHandler` to a new class.
(before it gets to big, auto case update is coming...)

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/NewApplicationHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/NewApplicationHandler.java
@@ -1,0 +1,72 @@
+package uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.events;
+
+import org.springframework.stereotype.Service;
+import uk.gov.hmcts.reform.bulkscan.orchestrator.services.ccd.PaymentsProcessor;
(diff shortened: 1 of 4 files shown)


apereverzin (approved):

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/NewApplicationHandler.java
```diff
@@ -0,0 +1,72 @@
...
+        AutoCaseCreator caseCreator,
+        PaymentsProcessor paymentsProcessor,
+        CreateExceptionRecord exceptionRecordCreator
+    ) {
+        this.caseCreator = caseCreator;
+        this.paymentsProcessor = paymentsProcessor;
+        this.exceptionRecordCreator = exceptionRecordCreator;
+    }
+
+    public EnvelopeProcessingResult processNewApplication(Envelope envelope, long deliveryCount) {
```
doncem, line 36: is it worth or makes sense to assert classification is definitely NEW_APPLICATION? just in case this handler is misused :shrug:

doncem (approved):

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/events/NewApplicationHandler.java
rkondratowicz: Yeah, it won't hurt adding it

PR #1239, opened 2020-09-24
Fix finding case for auto update

rkondratowicz:
search for a case the same way when we attach supplementary evidence

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/autocaseupdate/AutoCaseUpdater.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/autocaseupdate/AutoCaseUpdater.java
@@ -73,8 +73,7 @@ public class AutoCaseUpdater {
                     log.warn("Auto case update abandoned - case not found for envelope. {}", getLoggingInfo(envelope));
                     return ABANDONED;
                 });
-        } catch (
-            Exception exc) {
+        } catch (Exception exc) {
             log.error(
                 "Error while trying to automatically update a case. {}",
                 getLoggingInfo(envelope),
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/autocaseupdate/AutoCaseUpdater.java
```diff
@@ -43,18 +45,36 @@ public AutoCaseUpdater(
...
-                    String caseId = String.valueOf(matchingCases.get(0));
-                    return handleSingleMatchingCase(caseId, envelope);
-                case 0:
-                    return handleNoMatchingCases(envelope);
-                default:
-                    return handleMultipleMatchingCases(matchingCases, envelope);
-            }
-        } catch (Exception exc) {
+            return caseFinder
+                .findCase(envelope)
```
rkondratowicz, line 49: note: if multiple cases are found (when searching by legacy case ref) this method returns an empty optional

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/autocaseupdate/AutoCaseUpdater.java
```diff
@@ -43,18 +45,36 @@ public AutoCaseUpdater(
...
+                .map(
+                    existingCase -> {
```
aliveni, line 51: Don't we have to handle the scenario where multiple cases found for the envelope?

aliveni: I just saw your comment above.

aliveni (approved):

apereverzin (approved):

PR #1245, opened 2020-09-29
Call auto case updater

rkondratowicz:
From a class that handles envelopes with classification SUPPLEMENTARY_EVIDENCE_WITH_OCR


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/servicebus/domains/processedenvelopes/EnvelopeCcdAction.java
```diff
@@ -2,6 +2,7 @@
 
 public enum EnvelopeCcdAction {
     AUTO_ATTACHED_TO_CASE,
+    AUTO_UPDATED_CASE,
     CASE_CREATED,
     EXCEPTION_RECORD;
```
rkondratowicz, line 7: Or maybe just `CASE_UPDATED`? I wasn't sure.
There's no AUTO in `CASE_CREATED`...
Let me know what you think.

apereverzin: I suggest to rename `CASE_CREATED` to `AUTO_ CASE_CREATED` - later.

apereverzin (approved):

PR #1247, opened 2020-10-01
Create exception record when updating case fails too many times

rkondratowicz:
(in auto case update process)

https://tools.hmcts.net/jira/browse/BPS-1302

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/envelopehandlers/SupplementaryEvidenceWithOcrHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/envelopehandlers/SupplementaryEvidenceWithOcrHandler.java
@@ -15,6 +15,8 @@ import static uk.gov.hmcts.reform.bulkscan.orchestrator.services.servicebus.doma
 @Service
 public class SupplementaryEvidenceWithOcrHandler {
 
+    public static final int MAX_RETRIES = 2;
+
     private final CreateExceptionRecord exceptionRecordCreator;
     private final PaymentsProcessor paymentsProcessor;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/envelopehandlers/SupplementaryEvidenceWithOcrHandler.java
```diff
@@ -15,6 +15,8 @@
 @Service
 public class SupplementaryEvidenceWithOcrHandler {
 
+    public static final int MAX_RETRIES = 2;
```
aliveni, line 18: can we make this value configurable?

rkondratowicz: do we want to have it different depending on which environment it's running?

aliveni: no, just thought it would be easier if we want to increase the retries. it's up to you

aliveni (approved):

src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/ccd/envelopehandlers/SupplementaryEvidenceWithOcrHandler.java
mdayican: how do we retry ? if it fails do we try just after it?

PR #1280, opened 2020-11-19
Added func test

apereverzin:
https://tools.hmcts.net/jira/browse/DTSBPS-159

### Change description ###
Added functional test for auto case update: existing case has first name and last name OCR fields before update. This test should pass after fixing CCD definitions in aat.

```
[ ] Yes
[x] No
```


apereverzin:
Failing build

PR #1289, opened 2020-12-04
Logging instead of throwing exception in case of null fileName

apereverzin:
https://tools.hmcts.net/jira/browse/DTSBPS-272

### Change description ###
Logging instead of throwing exception in case of null fileName

```
[ ] Yes
[x] No
```


apereverzin:
Not necessary anymore

apereverzin:
Not necessary anymore

PR #2716, opened 2024-11-12
remove call to payment queue

Majid-justice:
https://tools.hmcts.net/jira/browse/FACT-1482

remove call to payment queue

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x ] No
```

PR #2741, opened 2025-04-04
FACT-1482: Use payments API rather than SB queue

joshblackmoor:

https://tools.hmcts.net/jira/browse/FACT-1482

### Change description ###

Functionality to use payments API rather than queue. 

Re-created PR, see comments on original #2716 

```
[ ] Yes
[x] No
```


joshblackmoor:
Not needed.

